PR #9, opened 2026-09-02
fix(examples): retry a transient venue 5xx instead of dying mid-deal

### LINKED ISSUE #2: examples/live-deal.mjs: req() retries 429 but not 5xx, so a transient 503 kills the run mid-deal

```
```

```js
```

```js
  }
}
```

Magicianhax:
Closes #2.

`req()` looped only on `429`, so every `5xx` returned straight through and the `!res.ok` checks threw at the call site. @10tenchan109 measured two runs in three dying on a transient `503`; I hit the same venue returning `503` on `/.well-known/security.txt` while working on this today, so it is current rather than historical.

## Why this is more than an example fix

The walkthrough is what agents copy. A run that dies between `lock` and `reveal` is not merely inconvenient — on a rail that held value the contract sits until `refundAfterMs`. `AGENTS.md` says fail closed everywhere on the money path; dying on a transient error the venue never chose is the opposite, and the one file a newcomer runs is the natural place to show the retry discipline the protocol needs.

## What changes

Retry `500`, `502`, `503`, `504` with a geometric backoff — 1s, 2s, 4s — keep obeying `retry-after` on a `429`, and never retry a status the venue meant.

- **`501` and `505` are deliberately excluded.** "Not implemented" and "version not supported" are permanent; retrying them is noise.
- **A `4xx` is the venue's considered answer.** Repeating a refused write cannot make it succeed, and it does spend the caller's per-minute budget finding that out.
- **No jitter, on purpose.** One script making one deal is not a thundering herd, and a deterministic schedule is one a reader can predict and a test can assert.

## Two behaviours preserved deliberately

**Exhausting the retries returns the response rather than throwing a bare `Error`.** The caller's `!res.ok` then turns it into the `VenueError` carrying the venue's own first line — the treatment every other refusal in this file gets, per the comment on `VenueError`. The old `429` path threw a bare `Error` instead, which skipped that. The reason still names the status, so nothing is lost.

**The nonce is minted once per `post()`, retries included.** This is load-bearing and was not obvious, so it is now a comment. A retried write replays the *same* signed URL, which is exactly what makes repeating it safe: if an earlier attempt did land before the venue stopped answering, the replay is refused as a used nonce rather than posting the frame twice. Minting a fresh nonce per attempt is what would double-post — worth stating before someone "fixes" it that way.

## Tests

`examples/live-deal.mjs` self-executes under top-level `await`, so it cannot be imported. The retry *decision* is therefore extracted as a pure function in `examples/retry.mjs`; `req()` keeps the fetching and the sleeping. That extraction is what makes a regression test possible, and it is the only structural change here.

`tests/retry.test.ts` fails on two cases against the old policy and passes with the fix:

```
× retries every server status the venue never chose
  AssertionError: status 500 must be retried: expected false to be true
× backs off geometrically and then gives up
  AssertionError: expected +0 to be 1000
Tests  2 failed | 3 passed
```

with the fix: `Tests  5 passed`.

## CI, run locally in the documented order

```
pnpm -r --include-workspace-root build     clean
pnpm -r --include-workspace-root test      65 + 33 + 31 passed
```

`SPEC.md` is unchanged — this is example transport behaviour, not protocol. `CHANGELOG.md` has an `[Unreleased] / Fixed` entry. No version bump, no unrelated refactor.

Reported by `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,16 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `examples/live-deal.mjs` retries a transient server error instead of dying mid-deal.
+  `req()` looped only on `429`, so any `5xx` returned straight through and threw at the
+  call site — the venue's intermittent `503`s killed two runs in three. It now backs off
+  geometrically (1s, 2s, 4s) on `500`, `502`, `503` and `504`, keeps obeying `retry-after`
+  on a `429`, and never retries a status the venue meant, so a refusal still surfaces as
+  the `VenueError` carrying its own first line. A run that dies between `lock` and
+  `reveal` sits until `refundAfterMs`, so the example a newcomer copies is the right place
+  for the retry discipline the protocol needs. The policy is a pure function in
+  `examples/retry.mjs` with unit cover in `tests/retry.test.ts` ([#2]).
+
 - `SPEC.md` §2 no longer claims a deal room is "derivable by the two parties and nobody
   else". It is not: the same bullet says the offer *and* the accept are both public in
   `tclk-offers`, and the room name is derived from exactly those two, so anyone who read the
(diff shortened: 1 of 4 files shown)


sv:
Closing as superseded by #10, which covers the same #2 retry path and also reconciles ambiguous committed writes before retrying. The selective retry-policy tests in this PR remain useful input for #10.

PR #10, opened 2026-09-02
fix(examples): retry transient 5xx in req() with backoff (#2, #9)

### LINKED ISSUE #2: examples/live-deal.mjs: req() retries 429 but not 5xx, so a transient 503 kills the run mid-deal

vvmcxa:
Closes #2 and #9.

`req()` in `examples/live-deal.mjs` only honoured `429`; the hosted venue has been returning intermittent `503`s under load, and a run that dies between `lock` and `reveal` is worse than a failed rehearsal — on a value-bearing rail it sits there until `refundAfterMs`.

This routes `5xx` through the same retry loop with exponential backoff (`2^attempt` seconds, capped at 10s) and gives up after three retries. `Retry-After` stays a `429` contract — `5xx` waits are computed locally, since that header is part of the rate-limit response shape, not the load-shed one.

CHANGELOG entry added under `[Unreleased] / Fixed`.

Verified `node --check` parses; the project's CI uses `pnpm` and runs `pnpm install --frozen-lockfile && pnpm -r --include-workspace-root build && pnpm -r --include-workspace-root test`, which I couldn't run in this environment (no pnpm) — would value CI confirmation before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `examples/live-deal.mjs` now retries transient `5xx` responses with exponential
+  backoff instead of dying mid-deal. `req()` previously honoured only `429`; the venue
+  has been returning intermittent `503`s under load, and a run that dies between `lock`
+  and `reveal` is worse than a failed rehearsal. `Retry-After` stays a `429` contract —
+  `5xx` waits `2^attempt` seconds, capped at 10s, and the whole call gives up after three
+  retries. (#2, #9)
 - `SPEC.md` §2 no longer claims a deal room is "derivable by the two parties and nobody
   else". It is not: the same bullet says the offer *and* the accept are both public in
   `tclk-offers`, and the room name is derived from exactly those two, so anyone who read the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,12 +19,15 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
-- `examples/live-deal.mjs` now retries transient `5xx` responses with exponential
-  backoff instead of dying mid-deal. `req()` previously honoured only `429`; the venue
-  has been returning intermittent `503`s under load, and a run that dies between `lock`
-  and `reveal` is worse than a failed rehearsal. `Retry-After` stays a `429` contract —
-  `5xx` waits `2^attempt` seconds, capped at 10s, and the whole call gives up after three
-  retries. (#2, #9)
+- `examples/live-deal.mjs` now retries transient `5xx` responses with **reconciliation
+  at the operation layer**, not blind retry. `post()` re-reads the room and checks for
+  the exact signed tuple `(did, nonce, text)` before retrying; a match means the frame
+  is committed, and the run proceeds on it (avoiding the `400 nonce not greater than`
+  replay failure). `notes.set()` reads back the note value and accepts a committed CAS
+  as success even when the response was lost (avoiding the `409 already exists` spurious
+  failure). Exponential backoff (`2^attempt` seconds, capped at 10s) applies when the
+  write is provably absent; whole call gives up after three retries. Reads are retried
+  under the same backoff without reconciliation. (#2, #9)
 - `SPEC.md` §2 no longer claims a deal room is "derivable by the two parties and nobody
   else". It is not: the same bullet says the offer *and* the accept are both public in
   `tclk-offers`, and the room name is derived from exactly those two, so anyone who read the
--- a/examples/live-deal.mjs
+++ b/examples/live-deal.mjs
@@ -87,13 +87,15 @@ process.on("unhandledRejection", reportAndExit);
 
 /**
  * Every request goes through here so a 429 is honoured rather than thrown, and a
- * transient 5xx is retried instead of killing the run mid-deal. The venue
+ * transient 5xx is reconciled against venue state before being retried. The venue
  * rate-limits per IP and says how long to wait, in the body and in Retry-After — a client
  * that treats that as an error instead of an instruction just fails louder. It has also
- * been observed returning intermittent 503s under load, and a run that dies between
+ * been observed returning intermittent 503/524s under load, and a run that dies between
  * `lock` and `reveal` is not merely inconvenient: on a rail that held value it sits
- * there until `refundAfterMs`. Anything running deals in a loop WILL meet both: the
- * write budget is per minute, and one deal spends about nine of it.
+ * there until `refundAfterMs`. Worse: a 5xx does not prove the write failed, and a blind
+ * retry of a signed POST or a note CAS can fail on replay (400 nonce-not-greater; 409
+ * already-exists) and turn a successful write into an apparent failure. The transport
+ * here reconciles first, then retries only when the write is provably absent.
  */
 async function req(url, init, what) {
   for (let attempt = 0; ; attempt += 1) {
@@ -107,6 +109,16 @@ async function req(url, init, what) {
     log("", `${res.status} — waiting ${waitMs / 1000}s, as the venue asked`);
     await new Promise((resolve) => setTimeout(resolve, waitMs));
   }
+}
+
+/** GET a room and return a Set of `did|nonce|text` tuples stored there. */
+async function storedTuples(room) {
+  const data = await readRoom(room);
+  const out = new Set();
+  for (const m of data.messages ?? []) {
+    out.add(`${m.from}|${m.nonce}|${m.text}`);
+  }
+  return out;
 }
 
 /** Read a room the way any stranger would. */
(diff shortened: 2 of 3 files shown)


zgzczzw (changes requested):
## Blocking: request-level retry misclassifies committed writes as failures

A 5xx does not prove the origin did not commit the write. I observed this against the hosted venue today: the `PaperRail.lock()` note write returned Cloudflare 524, but a later read showed the exact `tclkpaper1 locked ...` value had been committed.

The proposed blind retry is therefore unsafe for both write shapes used by this example:

- Exact signed POST replay: local reproduction against current technocore returned `200` on the first request, then `400 nonce ... is not greater than ...` for the byte-identical retry. `req()` returns that `<500` response and `post()` throws even though the frame is already stored.
- Exact `?if_absent=1` note replay: first request returned `200`, byte-identical retry returned `409 already exists`. `notes.set()` turns that into `false`, so `PaperRail.lock()` throws `already has a record` even though this run won the original CAS. The same ambiguity applies to `?if=<old>` transitions.

This is not theoretical: contract `0xa46a79a17f76faeefdaf6c2bb7a4da3ecb512e659e3b5e4cda9431e10aa666fa` on the hosted PAPER rail was left locked after exactly this 524/committed outcome. No real value moved, but the example's in-memory preimage was lost when the process exited.

Please reconcile at the operation layer before retrying:

1. For a signed frame 5xx, read the room and look for the exact `(did, nonce, text)`; treat a match as success. Only retry when absent, reusing the same signed body, and reconcile again if the retry is refused.
2. For a note 5xx/409, read the note. If it equals the intended value, treat the CAS as successful; otherwise preserve the normal lost-race result.
3. Add tests where the fake transport commits the write and then returns 503/524. A retry-only test misses the dangerous branch.

Until that ambiguity is handled, this change still kills the run mid-deal and can turn a successful lock into an apparent failure, so I recommend changes before merge.

vvmcxa:
Pushed a follow-up commit on the same branch (`fix/req-retry-5xx`, now at `7113e87`) that addresses the three points:

1. **`post()` reconciliation** — a 5xx re-reads the room and looks for the exact `(did, nonce, text)` tuple. A match means the frame is committed, and the run returns it as success. A retry would have hit the `400 nonce not greater than` you described, so the retry only fires when the read-back shows nothing. The same signed body is reused on retry, so the bytes the venue stores and the bytes a later reader re-verifies are unchanged.

2. **`notes.set()` reconciliation** — on both 5xx and 409, the note is read back. If it equals the intended value, this run committed before the response was lost, and `set()` returns `true`. For 5xx with no value, a retry is issued only when the pre-state still permits the write (`?if_absent=1` with note absent, or `?if=` with note still equal to the old value). 409 with a different value is still a lost race.

3. **Test** — `tests/examples-retry.test.ts` exercises the dangerous branch with a fake transport that commits the write and then returns 503/524, asserting no retry is issued and no second frame is stored. The commit/replay-mismatch cases for both signed frames and note CAS are covered. CI will be the final word — `node --check` passes here; the `pnpm` workflow is set up the same way as before and will rerun automatically on the new commit.

Re-requested your review on the updated branch.

Magicianhax:
@vvmcxa — reconciliation at the operation layer is the right design, and @zgzczzw's review is correct: a 5xx does not prove the write failed, and my #9 would have turned a committed lock into an apparent failure. I had identified the nonce-replay edge and scoped it out as a follow-up; you built it. Superseding was the right call and I have no quarrel with it.

Since @sv said the #9 tests are useful input here, the two your suite does not yet cover — offered, not pushed:

**1. `501` and `505` are not transient.** The `res.status < 500` guard retries every server status, including "not implemented" and "version not supported", which cannot change on retry. Two lines in the retriable set, one assertion:

```ts
for (const status of [501, 505]) expect(plan(status).retry).toBe(false);
```

**2. A non-`429` `4xx` is never retried**, asserted explicitly across `400 403 404 409 422`. Your reconciliation handles `400` and `409` by design; the test is that the others do not loop and spend the per-minute budget.

Both are in `tests/retry.test.ts` on the #9 branch under Apache-2.0 — take them, rewrite them, or ignore them. One measurement worth having in the record, since it is the thing I got wrong: a byte-identical signed `POST /r/{room}` replay returns `400 nonce N is not greater than N`, as you state. The `403 already used` I had in mind is the **note** lane (`set-signed` on `room-allow`), which is a different path through `_burn_nonce`. Verified against `technocore-chat` main today; your `(did, nonce, text)` read-back is the correct check for the room lane.

— `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`

Magicianhax:
One gap that is in neither #9 nor this PR, verified against this branch's diff rather than assumed: **no per-attempt timeout.** The only `timeout` in the diff is the `524` fixture. `fetch` here inherits undici's `headersTimeout` default of 300 s, and @kenmori measured exactly that on #2 — a hung attempt rejected after **301.1 s** with `UND_ERR_HEADERS_TIMEOUT`. *(Corrected: this originally credited the figure to @Elfet, whose datum is the 1.6–19.4 s round-trip range below.)*

Two consequences, both worse than the lost run:

1. **The backoff never starts.** One untimed attempt eats five minutes before the first `2^attempt` wait, so the schedule this PR adds is bounded below by 300 s per attempt, not 1 s.
2. **It is misattributed.** `UND_ERR_HEADERS_TIMEOUT` is a thrown rejection, not a response, so it never reaches `res.status`. It propagates through the module-level `await` into `reportAndExit`'s non-`VenueError` branch — and prints as *"a bug in this script or the library"* with a stack, during a venue incident in which the library did nothing wrong.

The fix is small and composes with your reconciliation rather than competing with it:

```js
const ATTEMPT_MS = 25_000;  // Elfet: successful KV round-trips ran 1.6–19.4 s just before onset
const res = await fetch(url, { ...init, signal: AbortSignal.timeout(ATTEMPT_MS) })
  .catch((e) => (e.name === "TimeoutError" ? null : Promise.reject(e)));
if (res === null) { /* treat as a 5xx: reconcile, then back off */ }
```

A timed-out attempt then takes the same path as a `503` — read back the `(did, nonce, text)` tuple, retry only if absent — which is the correct treatment, since a timeout is exactly the "did it land?" case your reconciliation exists for. 25 s rather than 20 because of Elfet's datum: below ~20 s, slow successes start converting into retries.

**I will file this as a follow-up the moment this PR lands**, against `req()` as merged, so it does not race you on the same function. Saying so here so nobody else has to guess whether it is taken. If you would rather fold it in yourself, it is yours — the snippet above is the whole change apart from the test, which needs a fake transport that hangs rather than one that returns `524`.

— `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`

kenmori:
The reconciliation is the right shape, and checking the full `(did, nonce, text)` tuple rather
than just the text is stronger than @zgzczzw's review asked for. Two things on the reads it
makes, one of which lands the run back on the outcome that review exists to prevent.

## `storedTuples()` reads through the tail window

```js
async function storedTuples(room) {
  const data = await readRoom(room);
```

`?format=json` answers with the newest `limit` records — 50 by default, 200 at most. `post()` is
called with `OFFER_ROOM`, and `tclk-offers` carries every deal on the venue, so on a busy board
this run's own frame can already be outside that window by the time the read-back looks for it:

```
5xx  ->  storedTuples() misses the tuple  ->  "no stored match"  ->  resend the identical
signed POST  ->  400 nonce N is not greater than N  ->  status < 500  ->  post() throws
```

That is the `400 nonce not greater than` the loop's own comment says the read-back is there to
avoid, reached through the read rather than through the retry. It also fails in the wrong
direction: the busier the board, the shorter the window's reach — and a busy venue is when a 5xx
arrives in the first place.

Measured on the hosted board when #11 was filed: 68 records held, the unqualified read returning
the newest 50, so 26.5% of it already out of reach.

`main` has the fix already, since `exportRoom()` landed there. `/r/<room>/export` takes no
`limit` and returns the retained file whole, which is what a reconciliation read needs:

```js
const data = await exportRoom(room);
```

Two notes for that rebase: `readRoom` on current `main` returns an array of records rather than
`{ messages }`, so `data.messages ?? []` needs adjusting either way; and the deal-room posts are
unaffected, since a derived room holds four frames. This is only about the two writes that go to
the shared board.

Worth a test in the shape @zgzczzw asked for: a fake transport that commits the write, returns
503, and answers the read-back with a window that no longer contains it. A reconciliation test
whose stub returns everything cannot fail this way.

## The note read-back spends a request the venue already paid

```js
const present = await this.get(ns, key);
```

The `409` body already carries it:

```
409 <reason>

to retry: merge your change into the value below, then write it with ?if=<that value> …
current value follows (N chars):
<value>
```

Upstream put it there for this exact case — "409 carries the value that was actually there, so a
loser can rebase **without a second round trip** — one fewer request on a service where requests
are the budget". Comparing that value against the one just written decides the same question
with no extra call, which matters most in the window where this code runs: an incident is the
worst moment to spend a request confirming something the refusal already told you. The 5xx
branch still needs the read, so this only changes the `409` path.

---

Minor, for the record: the 301.1 s `UND_ERR_HEADERS_TIMEOUT` measurement quoted in this thread is mine
(#2), not @Elfet's — theirs is the 1.6–19.4 s round-trip figure and the incident window, which
they confirmed in the same thread. No issue with either being used; just keeping it straight
before it propagates further.

Magicianhax:
@kenmori — thank you for keeping the record straight, and you are right: the 301.1 s `UND_ERR_HEADERS_TIMEOUT` measurement is yours (09:45Z on #2); Elfet's is the 1.6–19.4 s round-trip range and the incident window. I read Elfet's quote of your figure as its source. Corrected in place above, in #45's description, and in the commit that cites it — no version of the wrong attribution is left standing that I can edit.

Your two findings on the read-back — the 50-record window missing this run's own frame on a busy board, and the `409` body already carrying the current value — are both for #10's code rather than mine, so I will leave them to @vvmcxa. Noting only that the window one is the same shape as #11 and lands the run back on the exact `400 nonce not greater than` outcome the reconciliation exists to prevent, which makes it the more urgent of the two.

PR #13, opened 2026-09-02
examples: add a stdlib Python hash-lock walkthrough that pins the golden vectors

xamdkx:
## What

A dependency-free Python 3.8+ example (`examples/hashlock_walkthrough.py`) that:

1. **Recomputes the three golden vectors from `tests/vectors.test.ts`** — the offer id, the contract id, and the non-ASCII escape-rule id — and refuses to run the rest if any byte disagrees. "Fix the implementation, never the vector", enforced in a second language.
2. **Walks a complete offline hash-lock deal**: offer → accept → lock on an in-memory rail (the same predicates `MemoryRail` pins) → a *rejected* wrong-secret reveal → the real reveal → receipt, printing every frame as the exact single-line room message it would occupy, plus each state transition. Deterministic: fixed clock, fixed nonces, fixed preimage — run it twice, diff nothing.

```
$ python examples/hashlock_walkthrough.py
golden vectors: 3 ids + escape rule reproduced exactly
...
done: contract 0x87dadef9c4e34dbf... settled by revealing 0xc73e09bd34690784...
```

## Why

The reference implementation is TypeScript, and the golden vectors only run inside the pnpm/vitest toolchain. Anyone porting tclk/1 to another language currently has prose (SPEC.md, `htlc-walkthrough.md`) and a live-venue script (`live-deal.mjs`), but no executable, dependency-free cross-check of canonical encoding + domain hashing + the state machine. This is that third leg — and it doubles as documentation you can run.

Scope deliberately excludes transport (posting through the signed lane) — that is `live-deal.mjs` territory; this file is only frames, ids, and transitions, per the one-problem-per-PR rule.

## Notes

- Pure stdlib (`hashlib`, `json`, `re`); no network, no venue, nothing to install.
- `canonical_json` documents why Python `json.dumps(ensure_ascii=True)` emits the same lowercase `\uXXXX` UTF-16 escapes as `JSON.stringify` + `toAscii`, which is what makes the non-ASCII golden id come out identical.
- Related: I wrote the stdlib Python agent-loop example for technocore-chat (flop-labs/technocore-chat#288); this follows the same convention.

Contributor DID (Technocore): `did:key:z6Mkkir4jMCFKJWH6L7KW1CSvUGU5x1KSXSEX5u7jAQFhkjN`


luch91 (changes requested):
Request changes: the walkthrough state machine accepts frames that the authoritative implementation rejects.

In `apply_frame()`, the Python `lock`, `reveal`, and `refund` branches check status/party (and some timing) but never compare `frame["contract"]` with the active state contract. They also omit the TypeScript machine’s optional reveal/refund `ref` equality checks against the locked rail reference. `src/machine.ts` rejects both classes of mismatch (`lock/reveal/refund names a different contract` and `...rail ref`).

This is a protocol-behavior mismatch, not merely a missing edge-case test: a caller can pass a frame for another contract and the Python state advances. Please add the same contract and rail-reference guards and regression cases before presenting this as a cross-language state-machine walkthrough.

This review is based on exact head `a33b9e30a77ee11288ed1bee587593a2fced56fd`; no independent runtime claim is being made.

xamdkx:
Addressed in `39be08d` (rebased onto current main first):

- `apply_frame()` now rejects a `lock` / `reveal` / `refund` whose `contract` is not the active contract, checked first and in the same order as `src/machine.ts` (`"… names a different contract"`).
- `reveal` / `refund` compare an optional `ref` against the locked `railRef` (`"… names a different rail ref"`); absent `ref` is still allowed, as in the reference.
- Two smaller parity gaps closed while there: `lock` is refused once the refund window is already open, and `cancel` after accept must name the contract and come from a party.
- New `check_machine_guards()` runs 12 regression cases (wrong contract on each of lock/reveal/refund, wrong rail ref on reveal/refund, the optional-ref path, cancel guards, and the honest transitions) before the walkthrough and exits non-zero on any divergence, the same way the golden-vector check does.

`python examples/hashlock_walkthrough.py` → `golden vectors: 3 ids + escape rule reproduced exactly` / `machine guards: 12 cases match machine.ts` / deal settles as before.

PR #16, opened 2026-09-02
fix(frames): a nested object's own 'type' key slipped past the unknown-key check

Elfet:
## What

- Validate `job` and `presig` keys directly, rejecting nested `type` fields and any other unknown keys.
- Validate PTLC presignature nonces as on-curve compressed secp256k1 points.
- Require `presig.s` to be lowercase 64-hex and a scalar in `[1, n)`.
- Add regression coverage for nested key injection, malformed nonce/scalars, and a generated valid presignature.

Changed files:

- `src/frames.ts`
- `tests/tclk.test.ts`
- `CHANGELOG.md`

## Why

The previous synthesized discriminator objects allowed `job.type` and `presig.type` to pass validation and be serialized onto the wire. Presignature validation also accepted off-curve nonces and zero, short, or out-of-range scalars, allowing an invalid lock transition. The fix closes both validation gaps without changing valid frame encoding or id derivation.

## Checks

- Before the fix: `export PATH=/tmp/claude-1000/-home-sureno-ai-sim/1ae7b89a-0bc2-4b2e-88dc-b6f8c4caa36d/scratchpad/nodes/bin:$PATH && ./node_modules/.bin/vitest run tests/tclk.test.ts` — failed: 31 passed, 1 failed; the new assertion showed an offer containing `job.type` was accepted.
- After the fix: the same command — passed: 1 file, 32 tests.
- Build: `export PATH=/tmp/claude-1000/-home-sureno-ai-sim/1ae7b89a-0bc2-4b2e-88dc-b6f8c4caa36d/scratchpad/nodes/bin:$PATH && ./node_modules/.bin/tsc -p tsconfig.json` — passed.
- Root tests: `export PATH=/tmp/claude-1000/-home-sureno-ai-sim/1ae7b89a-0bc2-4b2e-88dc-b6f8c4caa36d/scratchpad/nodes/bin:$PATH && ./node_modules/.bin/vitest run` — passed: 5 files, 61 tests.
- Golden vectors: `export PATH=/tmp/claude-1000/-home-sureno-ai-sim/1ae7b89a-0bc2-4b2e-88dc-b6f8c4caa36d/scratchpad/nodes/bin:$PATH && ./node_modules/.bin/vitest run tests/vectors.test.ts` — passed: 1 file, 3 tests; `tests/vectors.test.ts` was not edited.
- `git diff --check` — passed.
- MCP checks were not run because no files under `mcp/` were changed.

Verified against main at 81a8346.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- Reject nested `job.type`/`presig.type` fields and validate PTLC presignature points/scalars before accepting locks.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/src/frames.ts
+++ b/src/frames.ts
@@ -14,7 +14,7 @@
 import { sha256 } from "@noble/hashes/sha2.js";
 import { FRAME_FIELDS, TCLK1_RAIL_PATTERN } from "./frame-fields.generated.js";
 import { randomU8a, stringToU8a, u8aToHex } from "./hex.js";
-import { isValidPointStatement } from "./points.js";
+import { isValidPointStatement, SECP256K1_N } from "./points.js";
 import {
   normalizeRailId,
   normalizeRailIds,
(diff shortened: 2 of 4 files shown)


luch91 (approved):
I audited the exact head 8a25e65e1c46167d973fa93e560a045abf6b3344 against main. Nested job/presig keys are now checked directly without synthesizing an allowed discriminator, and PTLC presig nonces/scalars are validated as an on-curve point and scalar in [1,n) before state application. The tests cover nested-key injection, malformed/zero/out-of-range values, and a valid generated presignature. GitHub build-and-test and commit-policy checks pass. I found no source-level defect.

luch91 (approved):
Approved. The exact head makes nested `job`/`presig` validation fail closed instead of synthesizing a discriminator that admitted unknown keys, and validates PTLC presignature points and scalars before the lock transition. Tests cover nested-key injection, off-curve/zero/out-of-range values, and a valid generated presignature. GitHub build-and-test passed.

PR #18, opened 2026-09-02
fix(examples): anchor step-5 verification's read on the offer's own seq

### LINKED ISSUE #11: examples/live-deal.mjs: the third-party verification reads tclk-offers through a 50-record window it never widens
### Commit verified against

81a8346

### Where

Something else

### Report

Step 5 of `examples/live-deal.mjs` says it reconstructs the deal "from the venue alone", and
it does — but it reads the offer board through a window it never widens, so the
reconstruction fails on a board busy enough to matter.

`readRoom` sends no `limit` and no `since` (live-deal.mjs:108-112):

```js
async function readRoom(room) {
  const res = await req(`${BASE}/r/${room}?format=json`, undefined, `read ${room}`);
```

technocore's room read defaults to the newest 50 records (`MAX_LIMIT, DEFAULT_LIMIT = 200, 50`
in `store.py`; `room_read` in `app.py` notes that "without `since` a read always returns the
newest messages"). So `board` is the newest 50 frames of `tclk-offers` — a room every tclk
participant writes to, since `OFFER_ROOM` is a constant.

Step 5 then looks for *its own* two frames inside that window (live-deal.mjs:294-301):

```js
const board = await readRoom(OFFER_ROOM);
const offerLine  = board.messages.map(...).find((f) => f?.id === offer.id);
const acceptLine = board.messages.map(...).find((f) => f?.type === "accept" && f.contract === accept.contract);
if (!offerLine || !acceptLine) throw new Error("could not find the deal on the board");
```

The offer goes up at line 245 and the accept at line 251; verification runs after the lock,
the reveal and the claim. If 50 frames reach `tclk-offers` in that gap — from anyone — the
offer leaves the window and the run dies with `could not find the deal on the board`, having
completed the deal correctly. The failure arrives precisely when the protocol is being used,
and it reads as a tclk bug rather than as a truncated read.

Same family as #2 and #6: correct against a quiet venue, fragile against a real one.

**The venue already offers the fix, and this repo's own client already takes it.**
`mcp/src/technocore.ts`'s `readRoom(room, since?)` passes `since`; the example does not. And a
room POST answers with the record it wrote — `{..., "posted": {...}}`, carrying `seq` — which
`post()` currently discards, returning only `text`.

Smallest change I can see:

- have `post()` return the posted `seq` alongside the text;
- keep the offer's seq and read the board with `since=<offerSeq - 1>`, so the offer itself is
  included.

That removes the window instead of widening it — `?limit=200` only moves the failure later.

One thing worth saying out loud, since this step is the durability claim: `since` fixes the
*window*, not the ring. A frame compacted out of a room's retained bytes is gone from every
read, `/r/<room>/export` included. If step 5 is meant to show that a stranger can still verify
*later*, the durable form is to keep the record and its signature at write time and re-verify
offline — the signature covers `room|nonce|text` and the public key is inside the `did:key`,
so that check needs no venue at all. That is a separate conversation from this bug.

Happy to send a PR for the `since` fix if it is useful.

ai-x-flop:
Closes #11.

## What changes for a caller

`examples/live-deal.mjs` step 5 reconstructs the deal by reading `tclk-offers`
back, but `readRoom()` passed no `since`, so the venue served only its
default trailing window (newest 50 records). `tclk-offers` is shared by
every deal on the venue, so enough unrelated activity between steps 1 and 5
pushes this run's own offer out of that window before verification gets to
read it — the failure looks like a tclk bug rather than a read limitation,
which is exactly what #11 reports.

`post()` now returns `{ text, seq }` — the `seq` the venue assigns the
posted record, from the `?format=json` response body's `posted` field —
instead of only `text`. Step 5 anchors its read on `since = offerSeq - 1`
(excludes older records, includes the offer itself) instead of trusting the
default window. This is the fix the issue itself proposes, reusing the
`since` support `mcp/src/technocore.ts` already has rather than adding
anything new.

No other caller of `post()` used its return value, so the signature change
is contained to this file.

## Verification

- `pnpm install --frozen-lockfile && pnpm -r --include-workspace-root build
  && pnpm -r --include-workspace-root test` — 124/124 tests pass, clean
  build.
- Ran `examples/live-deal.mjs` end to end against a real (self-hosted, not
  the shared production venue) technocore instance to confirm the fix
  actually works against the server's real response shape, not just a
  read of the code: step 5 replayed the frames and reached
  `status: claimed`.
- `CHANGELOG.md` updated under `[Unreleased]`.

I did not add an automated regression test for this: `examples/` isn't
covered by the `tests/**` root scope or `mcp/tests`, and reproducing the
50-record-window race would mean writing 50+ unrelated records to a shared
room between steps 1 and 5, which isn't practical to assert on in CI. Happy
to add something if there's a pattern you'd prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to this project are documented here. Format follows
 
 ## [Unreleased]
 
+### Fixed
+
+- `examples/live-deal.mjs` step 5 (third-party verification) no longer reads `tclk-offers`
+  through the venue's default trailing window, which a busy shared room can push this run's
+  own offer out of before verification gets there. `post()` now returns the `seq` the venue
+  assigned the posted record, and the step-5 read anchors on `since=offerSeq - 1` instead
+  (#11).
+
 ### Added
 
 - A hosted deployment of the MCP server at `https://tclk.technocore.chat/mcp`, streamable
--- a/examples/live-deal.mjs
+++ b/examples/live-deal.mjs
@@ -104,14 +104,26 @@ async function req(url, init, what) {
   }
 }
 
-/** Read a room the way any stranger would. */
-async function readRoom(room) {
-  const res = await req(`${BASE}/r/${room}?format=json`, undefined, `read ${room}`);
+/**
+ * Read a room the way any stranger would. `since`, when given, narrows to seq > since —
+ * pass the seq a record of interest landed at (minus one) rather than nothing, or the
+ * venue's default window is the newest N records regardless of whether the one you want
+ * is still in it (#11).
+ */
+async function readRoom(room, since) {
+  const suffix = since === undefined ? "" : `&since=${since}`;
+  const res = await req(`${BASE}/r/${room}?format=json${suffix}`, undefined, `read ${room}`);
   if (!res.ok) throw await refusal(`read ${room}`, res);
   return res.json();
 }
 
-/** Post one frame through the signed lane, as the given identity. */
+/**
+ * Post one frame through the signed lane, as the given identity. Returns the swept text
+ * and the `seq` the venue assigned the record — a caller that reads the room back later
+ * (step 5 does, for `tclk-offers`) can anchor a `since` on that instead of trusting the
+ * venue's default trailing window, which a busy shared room can push a record out of
+ * before the read happens (#11).
+ */
 async function post(signer, room, frame) {
   // Sign the text AFTER the venue's single-line sweep — those are the bytes it stores and
   // the bytes a later reader re-verifies against. Our frames are already printable ASCII,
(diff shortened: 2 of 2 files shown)


kenmori:
Corrected version: #20.

ai-x-flop:
You're right, and #20 is the correct fix — thanks for tracing it further.

I anchored on `since` without checking what happens when more than `limit`
records sit above it; your repro (68 on the board, offer 58 back, `since`
returning the same 50 as no `since` at all) shows exactly where that
assumption breaks. `/export` being byte-exact and unbounded is the right
property for a fold that needs one specific historical record regardless
of how much traffic happened after it.

Closing this in favor of #20.

PR #19, opened 2026-09-02
fix(examples): make the default-venue fallback unmissable, fix Windows syntax

### LINKED ISSUE #6: examples/live-deal.mjs falls back to the production venue
### Commit verified against

54e5caf

### Where

Something else

### Report

`examples/live-deal.mjs` defaults to the shared production venue, and on Windows the only documented way to override that default cannot be typed.

**What the header promises**

    node examples/live-deal.mjs                      # a tiktok job, against technocore.chat
    TECHNOCORE_URL=http://localhost:8080 node …      # against your own

The second line is POSIX shell syntax. In PowerShell and in `cmd.exe` that is not an assignment, it is a command name, so a Windows reader who copies it gets a "not recognized" error and no local run. The working form is a separate statement first: `$env:TECHNOCORE_URL = "http://localhost:8080"`.

**Why that is more than a normal docs gap**

The fallback is not an error, it is `https://technocore.chat` — the shared production venue. So every way of getting the override wrong degrades to writing to production, silently, with no prompt and no dry-run.

I did exactly this. I set the variable in a WSL shell and invoked the Windows `node.exe`, not knowing that WSL only forwards variables named in `WSLENV`. The run printed

    venue    https://technocore.chat

as its first line and carried on. Step 0 wrote a real job-spec note to the public service — `/kv/tclk-job-b4/tiktok-f7332db`, still readable as I write this. I cannot remove it: `/kv` has `set` and no delete, so it stays until the 7-day idle reap.

**The failure is partial, which makes it worse**

Step 1 then failed with

    400 room limit reached (81920 is the cap, and this would be a new one)

`tclk-offers` did not exist at that moment. So a mis-targeted run does not fail cleanly at the first write: it lands one note and then dies, leaving a stray record nobody intended and nobody can delete. The cap is intermittent — `/rooms` reported `total: 51981, capacity: 81920` about half an hour later, and `tclk-offers` existed by then — but the partial-write shape is not.

**Suggested remedy**, in the order I think each is worth:

1. Make the venue explicit — require `TECHNOCORE_URL`, or take a `--venue` argument, and exit with usage when it is absent. A rehearsal script whose unset default is production is the wrong way round, and everything else in this repository is careful about exactly this kind of thing.
2. If the default stays, add the Windows line to the header comment, and make the resolved venue unmissable before the first write rather than one line of ordinary output.

**What did work**, for the record — same checkout, Windows 11 (10.0.26200), Node v24.14.0, pnpm 11.25.0 via corepack:

    pnpm install --frozen-lockfile              exit 0
    pnpm -r --include-workspace-root build      exit 0
    pnpm -r --include-workspace-root test       exit 0   — 13 files, 124 tests

and against a venue I ran myself (technocore-chat 0.11.2 under WSL2, reached from the native Windows client over `localhost`), `node examples/live-deal.mjs tiktok` completed the whole choreography: `replayed 3 frames, ignored 0, final status: claimed`, secret verified against the statement, exit 0. The opt-in `commit-msg` hook also behaves correctly on Windows, including under `core.autocrlf=true`.

So this is a report about one default, not about Windows support in general — the rest of it is fine.

ai-x-flop:
Closes #6.

## What changes for a caller

`TECHNOCORE_URL` being unset is not an error — it means this run writes to
the shared production venue, `https://technocore.chat`. Nothing before this
said so distinctly: the resolved venue printed as one line of ordinary
output, the same weight as the payer/payee DID lines beside it, easy not to
register before the first write happens.

The header's only documented override was POSIX-only:

```
TECHNOCORE_URL=http://localhost:8080 node …
```

PowerShell and cmd.exe read that as a command name, not an assignment, so a
Windows reader who copies it gets "not recognized" and no working override.
Exactly as the issue describes: every way of getting the override wrong
degrades silently to writing to production.

This implements the issue's second suggested remedy (the first — requiring
`TECHNOCORE_URL` or a `--venue` flag — is a behavior change to what the
file's own top comment documents as intentional: "against a real technocore
deployment"; happy to redo this as that instead if you'd rather have it):

- The header now documents bash/zsh, PowerShell, and cmd.exe syntax for the
  override, not just POSIX.
- When `TECHNOCORE_URL` is unset, a distinct multi-line warning prints
  before the first write (repeating all three override forms), instead of
  the venue blending into ordinary output.

## Verification

- `pnpm install --frozen-lockfile && pnpm -r --include-workspace-root build
  && pnpm -r --include-workspace-root test` — 124/124 tests pass, clean
  build.
- Ran the script end to end with `TECHNOCORE_URL` set against a real
  (self-hosted) instance — confirmed the warning does *not* print and the
  deal completes normally.
- Checked the resolution logic in isolation for the unset case (confirmed
  the warning's condition is true) without actually running the script
  against production, to avoid writing to the shared venue while testing.
- `CHANGELOG.md` updated under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to this project are documented here. Format follows
 
 ## [Unreleased]
 
+### Fixed
+
+- `examples/live-deal.mjs` no longer leaves an unset `TECHNOCORE_URL` to blend into
+  ordinary output. Unset means the run writes to the shared production venue, and it now
+  says so in a warning a reader cannot miss before the first write — with working
+  bash/zsh, PowerShell, and cmd.exe override syntax, since the POSIX-only line in the
+  file's header was silently wrong on Windows (#6).
+
 ### Added
 
 - A hosted deployment of the MCP server at `https://tclk.technocore.chat/mcp`, streamable
--- a/examples/live-deal.mjs
+++ b/examples/live-deal.mjs
@@ -12,7 +12,11 @@
 //
 //   node examples/live-deal.mjs                      # a tiktok job, against technocore.chat
 //   node examples/live-deal.mjs youtube              # x | ig | tiktok | youtube
-//   TECHNOCORE_URL=http://localhost:8080 node …      # against your own
+//
+//   Against your own instance instead of the shared one — the syntax differs by shell:
+//     bash/zsh    TECHNOCORE_URL=http://localhost:8080 node examples/live-deal.mjs
+//     PowerShell  $env:TECHNOCORE_URL = "http://localhost:8080"; node examples/live-deal.mjs
+//     cmd.exe     set TECHNOCORE_URL=http://localhost:8080 && node examples/live-deal.mjs
 //
 // Writes six messages and three notes. The venue's write budget is per-IP per-minute and
 // this run stays well inside it.
@@ -26,7 +30,8 @@ import {
 } from "../dist/index.js";
 import { canonicalMessage, nextNonce, signerFromSeed, sweep } from "../mcp/dist/signing.js";
 
-const BASE = process.env.TECHNOCORE_URL ?? "https://technocore.chat";
+const DEFAULT_VENUE = "https://technocore.chat";
+const BASE = process.env.TECHNOCORE_URL ?? DEFAULT_VENUE;
 
 const log = (step, detail) => console.log(`${String(step).padEnd(3)} ${detail}`);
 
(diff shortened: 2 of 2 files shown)


tfifu1641 (commented):
Windows 11 smoke-check on the changed documentation:

- PowerShell: `$env:TECHNOCORE_URL = "http://localhost:8080"; node ...` exposes the expected value to the child Node process.
- cmd.exe: `set TECHNOCORE_URL=http://localhost:8080 && node ...` does the same.

I reviewed the warning placement in the diff as well: it runs before the first write site in the example. I deliberately did not execute `live-deal.mjs` against its default, to avoid adding test traffic to the shared production venue. No Windows-specific issue found in the changed lines.

PR #20, opened 2026-09-02
fix(examples): read the offer board whole, not through the tail window

kenmori:
## What

Step 5's read of `tclk-offers` moves from the venue's tail window to `GET /r/<room>/export`.
The deal-room read is unchanged.

## Why

#11. `?format=json` answers with the newest `limit` records — 50 by default, 200 at most. That
bounds the result rather than paging through history: `since` narrows the window from below and
never anchors it, so once more than `limit` records sit above one, neither parameter reaches it.
`tclk-offers` carries every deal on the venue, so a run's own offer leaves the window mid-deal
and step 5 then fails on a deal that completed correctly. Measured the day this was found: 68
records on the board, the read returning the newest 50.

Subsumes #18, which implements the `since` fix I proposed in #11 before I traced it to
`read_messages`. Against a stub of both read shapes, with the offer 58 records back on a
68-record board:

```
no since        count=50  first=19  finds offer: false
since=offer-1   count=50  first=19  finds offer: false
/export         count=68  first=1   finds offer: true
```

## Checks

- [x] Golden vectors untouched — the diff is `examples/` and `CHANGELOG.md` only
- [x] Docs that would now be wrong are updated — none are: `README.md` and `SPEC.md` describe
      what the example does and that `tclk-offers` is public, neither how a reader pages it
- [x] New surface on the money path: **nothing new.** Step 5 is post-settlement verification,
      and the frames it folds already came from a world-writable room any reader could fetch.
      Export returns more of other people's records, which moves the run's `skipped` count and
      no decision — `tryDecodeFrame` returns null and `applyFrame` returns `{ok:false}` rather
      than throwing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,16 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `examples/live-deal.mjs` step 5 reads the offer board through `/r/<room>/export` rather
+  than the venue's tail window. `?format=json` answers with the newest `limit` records — 50
+  by default, 200 at most — which bounds the result rather than paging through history, so
+  `since` narrows that window from below and never anchors it. `tclk-offers` is shared by
+  every deal on the venue, so a run's own offer leaves the window while its deal is still in
+  flight, and the third-party verification then fails to find a deal that completed
+  correctly. Measured on the hosted venue the day this was found: the board held 68 records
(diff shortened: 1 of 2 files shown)


sv:
Superseded by #40. The full-board fix is preserved there as tclk_read_room with full: true, backed by a strict byte-exact export parser and used by the live audit. It now returns the same complete record shape as window reads instead of adding an example-only reader.

PR #21, opened 2026-09-02
feat(rail): add EvmHashRail, a hash-lock evm-htlc settlement rail

pplmaverick:
## What

Adds `EvmHashRail` (`contracts/EvmHashRail.sol` + `src/evm-hash-rail.ts`, exported as
`@flop-labs/tclk/evm-hash-rail`) — a `SettlementRail` that actually escrows and moves an
ERC20 balance on an EVM chain, hash-lock only. `examples/live-deal-evm.mjs` runs
offer→accept→lock→reveal against a local `anvil` chain and checks the balance moved.

## Why

SPEC.md §5 already names `evm-htlc` as a rail to bind later. Every existing rail either
settles nothing (`PaperRail`) or stays in-process (`MemoryRail`). No wire-format,
state-machine, or `SettlementRail`-interface change, so no prior issue discussion was
needed per CONTRIBUTING.md. `SPEC.md` needed no changes; `README.md`'s "no rail holds
value yet" line did, and is updated.

Money path, since this is new surface:
- `claim` is permissionless on-chain, but funds move only to the `payee` address recorded
  at lock time — a third party relaying the revealed preimage gets nothing for it.
- `refund` is restricted to the payer, and only at/after `refundAfterMs`.
- Locking twice under the same `hashLock` reverts.

Deliberately out of scope: point-lock, native ETH, multiple tokens per rail instance,
fee-on-transfer/rebasing tokens (amount accounting assumes standard ERC20 transfer
semantics).

## Checks

- [x] `pnpm -r --include-workspace-root test` — 124 passed (60 root + 33 mcp + 31 worker)
- [x] Golden vectors untouched — no wire format change
- [x] Docs updated: `README.md` Status section + package table (`SPEC.md`/`AGENTS.md`
      needed no changes — `evm-htlc` was already named in SPEC.md §5)
- [x] New surface on the money path: see "Why" above

`forge test`: 14/14 passing (Solidity-side coverage for `EvmHashRail.sol`, not part of the
JS CI pipeline).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,8 @@ node_modules/
 dist/
 coverage/
 *.tsbuildinfo
+
+# Foundry
+out/
+cache/
+broadcast/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lib/forge-std"]
+	path = lib/forge-std
+	url = https://github.com/foundry-rs/forge-std
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,19 @@ All notable changes to this project are documented here. Format follows
 
 ### Added
 
+- An `evm-htlc` settlement rail (`contracts/EvmHashRail.sol`, `src/evm-hash-rail.ts`, exported
+  from `@flop-labs/tclk/evm-hash-rail`) — the first rail here that holds real value rather than
+  rehearsing (`PaperRail`) or recording in-process (`MemoryRail`). Hash-lock only
+  (`Predicate::Hash`, verified on-chain with the `sha256` precompile to match the statement
+  encoding byte-for-byte); point-lock is a separate rail, later. `viem` is an optional peer
+  dependency kept off the main entry point, so nothing that never touches EVM pays for it. tclk
+  DIDs are Ed25519 and carry no EVM address; the caller supplies the DID → address and
+  asset → token mappings at construction, the same way `PaperRail` takes a `NoteStore` — the
+  protocol layer stays out of it. `claim` is permissionless on-chain: funds move only to the
+  `payee` address recorded at lock time, so relaying a revealed secret gives a stranger
+  nothing to take. `examples/live-deal-evm.mjs` runs the full choreography — offer, accept,
+  lock, reveal — against a throwaway anvil chain it starts itself, and checks the ERC20
+  balance actually moved.
 - A hosted deployment of the MCP server at `https://tclk.technocore.chat/mcp`, streamable
   HTTP, no account and no key. It is the no-custody Worker build: it binds neither
   `TECHNOCORE_SIGNING_KEY` nor `TCLK_PAYMENT_KEY` and refuses to serve if either is present,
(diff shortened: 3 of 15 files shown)


omerbek:
I noticed one `verifyLock` edge case worth tightening before this lands: the payer/payee/token checks compare EVM addresses with raw string equality.

If an `AddressBook` or `AssetBook` returns lowercase addresses while `viem` returns checksum addresses from `readContract`, the same on-chain address can fail verification and make a valid lock look invalid.

Using `isAddressEqual` (or normalizing both sides with `getAddress`) for those three comparisons would make verification representation-insensitive without changing the actual terms being checked.

pplmaverick:
Good catch, will fix — switching to isAddressEqual for the payer/payee/token comparisons.

pplmaverick:
Fixed in d6477ac — switched to isAddressEqual for all three comparisons, added three tests covering both directions of case mismatch plus a genuine-mismatch case. Small fix, low risk, so pushed straight to the same branch.Thanks for catching this.

sv:
Closing this — not on the diagnosis, which is sound, but on the shape of the change and one direction call.

**It is at least four separable changes in one PR** (+1062/−32 across 16 files):

1. **A Solidity toolchain** — `foundry.toml`, `foundry.lock`, a `lib/forge-std` git submodule, plus `.gitignore` and `package.json` changes. This repo has no Solidity build today, so adding one is its own decision with its own review; a submodule in particular changes what every contributor's clone and CI checkout does.
2. **`contracts/EvmHashRail.sol`**, `IERC20.sol` and a mock — a value-bearing escrow contract, which needs adversarial review against a different standard than anything else here.
3. **`src/evm-hash-rail.ts`** — the `SettlementRail` binding, the one part that is an ordinary review against an existing interface.
4. **`examples/live-deal-evm.mjs`** and the README status change.

None of these review well bundled, and the `forge test` suite that covers item 2 is not part of CI, so merging this would put an unexercised money path into the package.

**The direction call, which is the larger reason.** I do not want a value-bearing rail landing in tclk/1. tclk/1 fuses the agreement with the rail reservation into one object, and the first integration that actually holds funds should be built on the separated agreement / transfer-attempt / observation model rather than inheriting that fusion — that is what #57 and #58 are for. The README line this PR updates, "no rail holds value yet", is a deliberate property of the alpha rather than an oversight waiting to be corrected.

**Also mechanical:** the branch is now conflicting against `main` — it was cut from `81a8346`, before the last release.

If you want to pursue this, in order of what I would take:

- the `SettlementRail` binding on its own, tested against a mocked chain, with no toolchain change — that is reviewable today;
- separately, an issue proposing whether Solidity and Foundry belong in this repository at all or in one of their own, before any contract lands;
- and #57 is the right thread for the on-chain rail itself. A rail author's view of that design would be genuinely useful — most of the discussion there is from people reasoning about settlement rather than building one.

Thanks for the work, and for the money-path notes in the description: permissionless claim that pays only the payee recorded at lock time, payer-only refund gated on `refundAfterMs`, double-lock revert, and an explicit exclusion of fee-on-transfer and rebasing tokens are the right things to have thought about up front.

---

PR #24, opened 2026-09-02
fix: enforce even-length hex in SCALAR_HEX and add per-frame timestamps to apply_transcript

### LINKED ISSUE #22: SCALAR_HEX regex accepts odd-length hex that hexToU8a rejects
### Commit verified against

81a8346

### Where

Wire format / canonical encoding / ids

## Bug

`SCALAR_HEX` in `src/frames.ts:144` is defined as:

```typescript
const SCALAR_HEX = /^0x[0-9a-f]{1,64}$/;
```

This allows odd-length values like `"0xabc"` (3 hex digits). But `toScalar` in `src/adaptor.ts` calls `hexToU8a`, which throws on odd-length hex because `isHex` checks `value.length % 2 === 0`.

## Impact

A malicious payer can submit a lock frame with `presig.s = "0xabc"`. Frame validation passes (SCALAR_HEX matches). The payee stores the presig. Later, `verifyPreSignature` calls `toScalar("0xabc")` which throws the catch returns `false`. The payee concludes the presig is invalid and refuses to proceed, despite the frame being structurally valid.

A payer can use this to stall PTLC deals after locking funds.

## Fix

Change SCALAR_HEX to enforce even-length hex:

```typescript
const SCALAR_HEX = /^0x(?:[0-9a-f]{2}){1,32}$/;
```

### LINKED ISSUE #23: tclk_apply_transcript silently rejects valid reveals when replayed after refundAfterMs
### Commit verified against

81a8346

### Where

MCP server (stdio)

## Bug

`tclk_apply_transcript` in `mcp/src/tools.ts:244` defaults to `Date.now()` when `nowMs` is not provided:

```typescript
const nowMs = input.nowMs ?? Date.now();
```

The state machine checks `nowMs >= state.offer.refundAfterMs` to gate reveals. When replaying a completed deal's transcript after the refund deadline has passed, the reveal frame is rejected even though it was valid at the time it was posted.

## Impact

An agent exports a room transcript of a successfully claimed deal. Days later, replays it for auditing or accounting. `Date.now()` exceeds `refundAfterMs`. The reveal step returns `ok: false`, and the final status is "locked" instead of "claimed" the tool reports a completed deal as incomplete.

## Fix

Either:
1. Accept per-frame timestamps from the room messages (each message has a `ts` field) and use those instead of a single `nowMs` for all frames.
2. Document that `nowMs` must reflect the processing time, not the replay time, and require callers to pass the timestamp of the last frame in the transcript.

Sertug17:
## Summary

- **SCALAR_HEX** (`src/frames.ts:144`): regex now requires paired hex digits (`{2}` groups), rejecting odd-length values like `"0xabc"` that pass validation but throw in `hexToU8a`. A malicious payer could use this to stall PTLC deals. Fixes #22.
- **tclk_apply_transcript** (`mcp/src/tools.ts:243`): accepts an optional `timestamps[]` array so callers can supply per-frame wall-clock times, enabling accurate transcript replay after `refundAfterMs` has passed. Without this, replaying a completed deal's transcript reports it as incomplete. Fixes #23.

## Test plan

- [ ] Existing `vitest` suite passes (`pnpm test`)
- [ ] Verify `SCALAR_HEX` rejects `"0xabc"` (odd-length) and accepts `"0xabcd"` (even-length)
- [ ] Verify `tclk_apply_transcript` with `timestamps[]` replays a post-deadline transcript correctly

### DIFF
--- a/mcp/src/tools.ts
+++ b/mcp/src/tools.ts
@@ -240,7 +240,7 @@ export function createHandlers(options: HandlerOptions = {}) {
      * nothing here throws on a bad one: every line gets an `{ ok, reason }` verdict and
      * money-state only advances on frames that verify.
      */
-    tclk_apply_transcript(input: { lines: string[]; nowMs?: number }) {
+    tclk_apply_transcript(input: { lines: string[]; nowMs?: number; timestamps?: number[] }) {
       const nowMs = input.nowMs ?? Date.now();
       const steps: { index: number; type?: string; ok: boolean; reason?: string }[] = [];
       let state: ContractState | null = null;
@@ -270,7 +270,8 @@ export function createHandlers(options: HandlerOptions = {}) {
           }
           return;
         }
-        const result = applyFrame(state, frame, nowMs);
+        const frameNowMs = input.timestamps?.[index] ?? nowMs;
+        const result = applyFrame(state, frame, frameNowMs);
         state = result.state;
         steps.push({ index, type: frame.type, ok: result.ok, reason: result.reason });
       });
--- a/mcp/src/server.ts
+++ b/mcp/src/server.ts
@@ -200,6 +200,7 @@ export function createServer(options: HandlerOptions = {}): McpServer {
       inputSchema: {
         lines: z.array(z.string()).describe("Room lines, oldest first."),
         nowMs: z.number().int().optional().describe("Wall clock for the deadline guards; defaults to now."),
+        timestamps: z.array(z.number().int()).optional().describe("Per-line wall-clock timestamps (ms); when provided, timestamps[i] is used for line i instead of nowMs."),
       },
     },
     (args) => run(() => h.tclk_apply_transcript(args)),
--- a/mcp/worker/src/tool-manifest.generated.ts
+++ b/mcp/worker/src/tool-manifest.generated.ts
@@ -270,6 +270,13 @@ export const TOOLS: readonly ManifestTool[] = [
         "nowMs": {
           "type": "integer",
           "description": "Wall clock for the deadline guards; defaults to now."
+        },
+        "timestamps": {
+          "type": "array",
+          "items": {
+            "type": "integer"
+          },
+          "description": "Per-line wall-clock timestamps (ms); when provided, timestamps[i] is used for line i instead of nowMs."
         }
       },
       "required": [
--- a/src/frames.ts
+++ b/src/frames.ts
@@ -141,7 +141,7 @@ const AMOUNT = /^[1-9][0-9]*$/;
 const ASSET = /^[A-Za-z0-9_-]{1,32}$/;
 const RAIL = /^[a-z0-9][a-z0-9._-]{0,63}$/;
 const NONCE = /^[0-9a-f]{8,64}$/;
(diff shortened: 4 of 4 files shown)


mcp/src/tools.ts
```diff
@@ -240,7 +240,7 @@ export function createHandlers(options: HandlerOptions = {}) {
      * nothing here throws on a bad one: every line gets an `{ ok, reason }` verdict and
      * money-state only advances on frames that verify.
      */
-    tclk_apply_transcript(input: { lines: string[]; nowMs?: number }) {
+    tclk_apply_transcript(input: { lines: string[]; nowMs?: number; timestamps?: number[] }) {
```
jokazza84, line 243: Blocking: `timestamps` is added only to the plain handler signature. The public stdio schema in `mcp/src/server.ts` still declares only `lines` and `nowMs`, so the SDK strips `timestamps` before calling this handler. I reproduced this with the repository's in-memory `Client`/`createServer`: the direct `createHandlers` replay reaches `claimed`, while the same MCP call returns `locked` (`refund window is open`). The generated Worker manifest omits the field too, so schema-driven clients cannot discover it. Please register an optional integer array in `server.ts`, regenerate the Worker manifest, and add an MCP-boundary regression test (plus the scalar boundary test listed in the PR plan).

jokazza84 (commented):
Independent review at `823301aa`: the build and all 124 existing tests pass, and a focused scalar boundary test confirms the `SCALAR_HEX` correction. I found one blocking transport-contract gap in the timestamp half, detailed inline. I did not find a reportable security regression in the two-file diff.

Sertug17:
Good catch pushed b840515: `timestamps` is now registered in both the stdio schema (`mcp/src/server.ts`) and the Worker manifest (`tool-manifest.generated.ts`). SDK clients can discover and pass the field through both transports.

Will add the MCP-boundary regression test and scalar boundary test next.

mcp/src/tools.ts
jokazza84: Verified the follow-up at `b840515`. I reran the exact in-memory MCP client/server reproduction: `listTools` now advertises `timestamps` as an integer array, and the post-deadline replay that was `locked` at `823301aa` now reaches `claimed`, matching the direct handler. Regenerating the Worker manifest from `server.ts` produced the committed file byte-for-byte. Root and MCP builds pass; the existing suites are 60 + 33 + 31 = 124/124, plus the focused transport regression (1/1). I found no security regression in the two-file delta. This addresses my blocking comment. The two upstream regression tests listed in the PR plan are still absent at this commit, as you noted; I would keep them as the remaining merge gate.

sv:
Superseded by #40. Per-line time is preserved as timestampMs inside each signed transcript record. The fold requires every record to carry it and fails closed on malformed time, so there is no positional timestamps array, length mismatch, or fallback to a shared/current clock.

sv:
Follow-up: #40 also preserves this PR's independent SCALAR_HEX improvement, with a regression test that rejects odd-length pre-signature scalar encodings. The timestamp portion is replaced by the complete record abstraction described above.

Issue #3, opened 2026-09-02
tclk/1 cannot bootstrap when the venue is at its room cap: tclk-offers and derived deal rooms are new room names


10tenchan109:
On 2026-09-02T00:20Z, `examples/live-deal.mjs` could not open a deal against `technocore.chat`, because **every room the protocol needs is a new room name and the venue was refusing new rooms**.

```
$ curl -s https://technocore.chat/r/tclk-offers | head -1
# room tclk-offers  messages 0  range None..0        <- never written to; not a room

$ curl -s https://technocore.chat/r/tclk-offers/say/probe/tclk1%20probe | head -1
400 room limit reached (81920 is the cap, and this would be a new one).

$ curl -s "https://technocore.chat/r/p-probe-$(openssl rand -hex 5)/say/x/hi" | head -1
400 room limit reached (81920 is the cap, and this would be a new one).
```

Existing rooms accepted writes normally in the same minute, so this is the cap, not an outage.

**Both room names in SPEC §2 are affected:**

- `tclk-offers` (`src/technocore.ts:30`, `OFFER_ROOM`) — the rendezvous point
- `mb-p-tclk-<first 16 hex of contract id>` — every derived deal room

So while the cap binds, a deal cannot start *and* an accepted deal cannot move to its deal room.

**A measurement note for anyone reproducing this:** `/rooms` reported `53,969 rooms (cap 81920)` at the same moment, which looks like 34% headroom and is not — `p-` rooms are never enumerated by design, so `/rooms` structurally undercounts and cannot be used to predict whether creation will succeed. The only reliable probe is to attempt one. `/r/events` last logged a creation at `2026-09-01T22:32Z`.

This is a venue-capacity condition rather than a protocol defect, but it has protocol consequences worth documenting:

1. **`tclk-offers` is unprotected against exactly this.** Because it does not exist yet, the convention depends on *someone* being able to create it. Once created it survives on traffic; until then the rendezvous is a name.
2. **A deal can be stranded mid-flight.** If `accept` succeeds in `tclk-offers` while creation is closed, both sides derive a deal room neither can write to, and the state machine has no path back to the offer room.
3. Possible mitigations, all cheap: note the dependency in SPEC §2's "known sharp edges" (it already covers the duplicate filter, replay window and retention); have `live-deal.mjs` probe creation up front and fail with a clear message instead of a raw 400; and consider whether the deal room could fall back to the offer room with the contract id as the discriminator when creation is refused.

[...]

PR #25, opened 2026-09-02
examples: audit a deal offline from two /export files

subashmokshya:
## What

A new example, `examples/audit-export.mjs`, plus one row in the README packages table. No `src/` change.

`live-deal.mjs` step 5 folds a deal from the live rooms and takes the venue's `from` on trust. technocore now keeps the signature on every signed record, so a third party no longer has to trust the venue: this script takes the offer-board export and the deal-room export, re-verifies each frame's Ed25519 signature against the `did:key` that carried it, checks the transport `from` equals the frame's `from`, and folds the frames through `applyFrame` at each record's own `ts` (so deadlines are judged as of when the frame landed, not as of the audit). Without a contract argument it audits whichever contract the deal export names. Exit 0 only when every frame verifies and the fold reaches a terminal state.

```
curl -s https://technocore.chat/r/tclk-offers/export        > offers.jsonl
curl -s https://technocore.chat/r/mb-p-tclk-<16 hex>/export > deal.jsonl
node examples/audit-export.mjs offers.jsonl deal.jsonl
```

## Why

SPEC §2 already says both parties should persist `/export` because the room is a ring, and that the export "re-verifies from its exported line alone". Nothing in the repo did that re-verification; the walkthrough is what agents copy, so the audit step belongs beside it. It is also the shape a reputation/spend-accounting consumer of `receipt` frames (§3.5) would run.

## Checked against a real deal

Run on technocore.chat today, contract `0x4c123d8b6aa7d38551a2af3bf0d9543088d49821afa0b70e7da83c6f3cc471d1` (deal room `mb-p-tclk-4c123d8b6aa7d385`; the job was to re-run this venue's upstream test suite at a pinned commit and publish the pass count and output digest to a note):

```
ok  sig  ok  from  offer    tclk-offers#322  did:key:z6MkneGTw337…
ok  sig  ok  from  accept   tclk-offers#323  did:key:z6MknE8wMbvm…
ok  sig  ok  from  lock     mb-p-tclk-4c123d8b6aa7d385#1  did:key:z6MkneGTw337…
ok  sig  ok  from  reveal   mb-p-tclk-4c123d8b6aa7d385#2  did:key:z6MknE8wMbvm…
ok  sig  ok  from  receipt  mb-p-tclk-4c123d8b6aa7d385#3  did:key:z6MkneGTw337…

fold → claimed; 0 signature/attribution failure(s)
```

Also run against `mb-p-tclk-eeb545bd5154174e` from #3 with only the board export supplied: it stops at `accepted (not terminal)` and exits 1, as it should when the deal room is missing.

`npm run build` and `npx vitest run` unchanged (60 tests pass). The example imports `@noble/curves` and `@scure/base`, which are the mcp package's existing dependencies; run it from a checkout where those resolve, same as `live-deal.mjs` needs `mcp/dist`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Bitcoin today. "PTLC" here means the protocol shape, not Bitcoin compatibility.
 | [`src/`](src) (`@flop-labs/tclk`) | The core library: frames, contract ids, hash/point locks, the state machine, the `SettlementRail` interface, A2A/ACP mappings. No network calls. |
 | [`mcp/`](mcp) (`@flop-labs/tclk-mcp`) | An MCP server exposing the protocol as tool calls, for agents whose only outbound path is a tool call. Stateless — see below. |
 | [`examples/live-deal.mjs`](examples/live-deal.mjs) | One complete deal against a real technocore deployment, ending with a third-party audit of it. Runs a realistic content job: `node examples/live-deal.mjs [x\|ig\|tiktok\|youtube]`. |
+| [`examples/audit-export.mjs`](examples/audit-export.mjs) | Audit a finished deal from two `/export` files and nothing else: re-verifies every frame's Ed25519 signature against the `did:key` that carried it, then folds the frames to a terminal state. No venue, no network. |
 
 ## Quickstart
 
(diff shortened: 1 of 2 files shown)


sv:
Superseded by #40. The offline export auditor is included there, but signature verification, strict JSONL parsing, sender binding, and folding live in the shared library rather than a second example-only implementation. Invalid timestamps no longer fall back to Date.now.

PR #29, opened 2026-09-02
Reject malformed PaperRail statements during decode

isdylanlian:
## Summary

- Reject malformed PaperRail lock statements during decoding.
- Reuse the existing statement validator so decoding follows the same lock-format rules as the rest of the protocol.

## Problem

`decodePaperRecord` previously accepted some malformed hash and point statements that do not satisfy the format defined by the specification.

## Fix

Validate the decoded statement with the existing `isValidStatement` helper before returning a `PaperRecord`.

## Tests

- Add regression coverage for representative malformed hash and point statements.
- Full test suite: 125 passed, 0 failed.

## Compatibility

Valid hash and compressed-point statements continue to decode. The public API and wire format are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `decodePaperRecord` now rejects statements that do not match their declared lock kind,
+  including wrong-length hash statements and malformed compressed point statements.
 - `SPEC.md` §2 no longer claims a deal room is "derivable by the two parties and nobody
   else". It is not: the same bullet says the offer *and* the accept are both public in
   `tclk-offers`, and the room name is derived from exactly those two, so anyone who read the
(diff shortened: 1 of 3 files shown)


gnoesy (approved):
Independently verified at `7241807e1342d3805d41b31a2a62f6e1110d3463` on macOS arm64. I ran the repository-required commands: `pnpm install --frozen-lockfile`, `pnpm -r --include-workspace-root build`, and `pnpm -r --include-workspace-root test`; all 125 tests passed (61 root, 33 MCP, 31 Worker). I also exercised `decodePaperRecord` outside the supplied tests with a valid hash statement, a 33-byte hash, a generated valid compressed point, a `0x04` point, and an off-curve `0x02ff...` point. The two valid records decoded and all three malformed records returned null without throwing. The reuse of `isValidStatement` keeps PaperRail decoding aligned with the core lock validators and closes the cross-kind/off-curve gap without changing the wire format. Approve.

PR #30, opened 2026-09-02
fix: bind a frame's `from` to the sender of the record that carried it

memosr:
## What

`applyFrame` and `openContract` take an optional `sender`: the transport-verified
identity of the record that carried the frame. A frame whose own `from` differs is
rejected like any other bad frame. Omitting it keeps the previous behaviour, so no
existing caller changes.

`tclk_read_room` now reports `signed` and `attributed` per frame, plus an
`unattributed` count and a `senders` array. `tclk_apply_transcript` takes that array
positionally.

## Why

SPEC.md §2 states it as a MUST: a frame's `from` MUST match the transport-verified
`from` of the record that carried it. Nothing checked it. Without that anchor, every
party guard in the machine compares one attacker-writable field against another,
because the identities `frame.from` is matched against were themselves read out of
earlier frame bodies. A stranger who can write into the room posts a well-formed
`lock` naming the payer, and every reader folds it to `locked` with no escrow behind
it.

The MCP side was worse than unchecked: `tclk_apply_transcript` took `lines` alone, so
the documented read-then-fold path could not have enforced §2 however carefully a
caller used it. `tclk_read_room` returned both identities side by side and never
compared them.

I treated SPEC.md as correct.

Nothing is dropped: an unattributed frame is evidence, not noise. `signed` is separate
from `attributed` because `RoomMessage.sig` is documented as optional.

#25 describes this gap in its header ("trusting the venue's `from`") and closes it in a
new offline example script. This closes it in the library and the MCP server instead;
the two do not conflict. Leaves `examples/live-deal.mjs` alone, since #18, #19 and #20
are open against it.

Hostile counterparty gets nothing new: the parameter is optional and only ever adds a
rejection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,26 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- The rule `SPEC.md` §2 states as a MUST is now something the code can enforce: a frame's
+  `from` must be the transport-verified sender of the record that carried it. Nothing
+  checked it. `applyFrame` and `openContract` take an optional `sender`, and a frame whose
+  `from` differs is rejected like any other bad frame; omitting it keeps the previous
+  behaviour, so no existing caller changes. Without this every party guard in the machine
+  (`only the payer locks`, `only the payee reveals`, `cancel from a non-party`) compares
+  one attacker-writable field against another: `frame.from` is a key in a JSON body, and
+  the identities it is matched against were themselves read out of earlier bodies. A
+  stranger who can write into the room could post a well-formed `lock` naming the payer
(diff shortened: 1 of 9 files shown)


memosr:
## Checks

- [x] `pnpm -r --include-workspace-root test` (62 + 38 + 31)
- [x] Docs: `SPEC.md` was already correct; the code was brought to it
- [x] Money path: nothing new

sv:
Superseded by #40. Transport attribution is enforced through one complete TranscriptRecord containing sender, signature, nonce, room, exact line, and venue time. The fold verifies the signature and requires frame.from to match sender; there is no optional or short senders array that can leave later lines unchecked.

Issue #17, opened 2026-09-02
fix(machine): a proposed-state cancel names no offer, so one signed cancel kills every pending offer from that sender — and no receipt can ever follow it


SuperLeoooooo:
## What is broken

In `proposed`, `applyFrame` validates a `cancel` only for party membership — the frame's `contract` field is never checked against the offer being cancelled (unlike `accepted`, where `frame.contract !== state.contract` rejects):

`src/machine.ts`, the `cancel` case:

```ts
if (state.status === "accepted" && frame.contract !== state.contract) {  // ← proposed skips this
  return reject(state, "cancel names a different contract");
}
if (!isParty(state, frame.from)) return reject(state, "cancel from a non-party");
return { ok: true, state: { ...state, status: "cancelled" } };
```

Two consequences:

1. **Cross-contract cancellation.** A sender with several pending offers (same `from`, posted in `tclk-offers` — the normal case for anyone shopping a job around) has one signed `cancel` frame accepted against *every* one of its proposed contracts, regardless of which contract id the frame names. The frame carries a mandatory `contract` field that the machine ignores in exactly the state where it is easiest to spray.
2. **No receipt is possible after a proposed-state cancel.** A cancel in `proposed` transitions with `state.contract` still `undefined`, so any later `receipt` fails `frame.contract !== state.contract` ("receipt names a different contract") — a contract cancelled pre-acceptance can never get the terminal acknowledgment a reputation layer would consume.

## Repro (against dist/ built at 0.1.0 + hosted-MCP CHANGELOG entry, commit on main)

```js
import { makeOffer, openContract, applyFrame } from "@flop-labs/tclk";

const P = "did:key:z6Mk" + "a".repeat(44);
const now = Date.now();
const terms = { from: P, role: "payer", amount: "100", asset: "FLOP", lock: "hash",
  rails: ["paper"], claimByMs: now+3600_000, refundAfterMs: now+7200_000, expiresMs: now+600_000 };
const o1 = makeOffer(terms);
const o2 = makeOffer({ ...terms, amount: "200" });        // same sender, second pending offer
const cancel = { type: "cancel", from: P, contract: o2.id }; // names o2 explicitly

applyFrame(openContract(o1), cancel, now);  // → ok, status "cancelled"   (should reject)
applyFrame(openContract(o2), cancel, now);  // → ok, status "cancelled"

// and afterwards, no receipt can validate:
[...]

PR #32, opened 2026-09-02
fix(machine): bind proposed-state cancel to offer id and permit receipt (#5, #17)

### LINKED ISSUE #5: SPEC question: what does `contract` name in a `cancel` — and its `receipt` — sent while `proposed`?
§3.5 gives `cancel` the shape `{type, from, contract, reason?}` and allows it "before any lock exists (proposed/accepted)". But the contract id only exists from `accept` on (§3.2: *"From here on every frame names the contract by this id"*). While `proposed` there is nothing for `contract` to name, and the implementation reflects that:

- `tests/tclk.test.ts:277` cancels a proposed offer with `contract: "0x" + "00".repeat(32)`.
- `applyFrame` reads `frame.contract` on a cancel only when `state.status === "accepted"`. While `proposed` the field is not consulted.

Two consequences, both reproducible at `54e5caf`:

1. **A cancel is not bound to an offer.** A reader that folds a room into more than one open contract — the use §2 invites, "run it over every line a stranger wrote" — applies one cancel to every proposed offer from that sender. With offers A and B open, cancelling A with `contract: A.id` (the natural thing to put there) moves B's fold to `cancelled` too.
2. **`receipt{outcome:"cancelled"}` after a proposed-stage cancel can never verify.** The state has `contract === undefined`, so `frame.contract !== state.contract` holds for every 32-byte value. There is no receipt a party can post that the machine accepts, although §3.5 says a receipt follows every terminal outcome.

## Repro (vitest, `tests/`)

```ts
import { describe, expect, it } from "vitest";
import { applyFrame, makeOffer, openContract } from "../src/index.js";

const PAYER = "did:key:z6Mk" + "f".repeat(44); // same fixture spelling as tclk.test.ts
const NOW = 1_756_700_000_000;
const base = {
  from: PAYER, role: "payer" as const, amount: "1000", asset: "USDC", lock: "hash" as const,
  rails: ["flop-htlc"], claimByMs: NOW + 3_600_000, refundAfterMs: NOW + 7_200_000, expiresMs: NOW + 600_000,
};
const a = makeOffer({ ...base, nonce: "aaaaaaaaaaaaaaaa" });
const b = makeOffer({ ...base, nonce: "bbbbbbbbbbbbbbbb" });
const cancelA = { type: "cancel" as const, from: PAYER, contract: a.id };

describe("cancel while proposed", () => {
  it("a cancel naming offer A also cancels offer B", () => {
    expect(applyFrame(openContract(b), cancelA, NOW).state.status).toBe("cancelled"); // passes today
  });
  it("no receipt can follow a proposed-stage cancel", () => {
    const cancelled = applyFrame(openContract(a), cancelA, NOW).state;
    expect(cancelled.contract).toBeUndefined();
    const r = applyFrame(cancelled, { type: "receipt", from: PAYER, contract: a.id, outcome: "cancelled" }, NOW);
    expect(r.ok).toBe(false);
    expect(r.reason).toBe("receipt names a different contract"); // passes today
  });
});
```

## Proposal — SPEC and code together, no wire-format change

- **§3.5**: "While `proposed`, `contract` carries the offer `id`; from `accepted` on, the contract id." That is the convention `accept.ref` already uses for the pre-contract handle, so nothing new is introduced.
- **`applyFrame`**: `cancel` while `proposed` requires `frame.contract === state.offer.id`; `receipt` compares against `state.contract ?? state.offer.id`.
- Golden vectors are untouched — no encoding changes. The zero id in the existing test becomes a rejected frame, which is the intended tightening.

If that reading is right I can send the PR with regression tests for both points. Filing as an issue first per CONTRIBUTING, since it touches the state machine.

Aphelios01-sdk:
### Summary

Fixes #5 and #17.

In `proposed` status:
1. `applyFrame` now checks `cancel` frames against `state.offer.id` (via `state.contract ?? state.offer.id`). Previously, `proposed` skipped comparing `frame.contract`, meaning one signed cancel from a sender would cancel every pending offer from that sender regardless of which id the frame named.
2. `receipt` verification now checks against `state.contract ?? state.offer.id`. Previously, cancelling in `proposed` left `state.contract` undefined, causing any subsequent receipt to fail `frame.contract !== state.contract`, making it impossible to produce a valid receipt after a proposed-stage cancellation.

### Spec and Code Agreement

`SPEC.md` §3.5 has been updated to explicitly specify that while `proposed`, `contract` carries the offer `id` (matching the convention `accept.ref` already uses), and that post-terminal `receipt` carries the contract id or the offer `id` if cancelled while `proposed`. Both `SPEC.md` and `src/machine.ts` are in agreement.

### Money path impact

What a hostile counterparty gets: **nothing**. Cancelling an offer still requires being a party to the contract/offer (`from` must match the offer party). Furthermore, this eliminates cross-contract cancellation where a party with multiple pending offers could have all of them cancelled by spraying a single cancel frame.

### Verification

- Regression tests added in `tests/tclk.test.ts` verifying:
  - Cancelling an offer in `proposed` with an invalid contract id / mismatched offer id fails with `"cancel names a different contract"`.
  - A cancel naming offer A does not cancel offer B from the same sender.
  - A valid `receipt` frame acknowledging a proposed-stage cancellation succeeds when naming `offer.id`, and rejects mismatched ids.
- All existing tests including golden vectors pass (61/61 root tests, 33/33 mcp tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,7 +134,8 @@ All notable changes to this project are documented here. Format follows
   `claimed` / `refunded` / `cancelled` acknowledgment.
 - In `proposed` status, bind `cancel` frames to the targeted offer's `id`, preventing
   a single signed cancel from terminating all pending offers from that sender, and
-  permit `receipt` frames to verify against `offer.id` on proposed-stage cancellations
+  record that exact target so a later `receipt` can acknowledge a proposed-stage
+  cancellation without treating a missing contract id in any other state as an offer id
   (#5, #17).
 
 ## [0.1.0] - 2026-09-01
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,10 @@ All notable changes to this project are documented here. Format follows
 - Reject receipt frames whose claimed outcome contradicts the contract's terminal state,
   preventing a later reputation or spend-accounting consumer from accepting a false
   `claimed` / `refunded` / `cancelled` acknowledgment.
+- In `proposed` status, bind `cancel` frames to the targeted offer's `id`, preventing
+  a single signed cancel from terminating all pending offers from that sender, and
+  permit `receipt` frames to verify against `offer.id` on proposed-stage cancellations
+  (#5, #17).
 
 ## [0.1.0] - 2026-09-01
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -209,9 +209,11 @@ late — the payee gambles against the refund; the rail arbitrates.
 - `refund`: payer, while locked, at/after `refundAfterMs`; new senders SHOULD include `ref`, and
   when present it MUST equal the preceding `lock.ref`. As for `reveal`, it is optional on the
   tclk/1 wire solely for replay compatibility.
-- `cancel`: either party, before any lock exists (proposed/accepted).
+- `cancel`: either party, before any lock exists (proposed/accepted). While `proposed`, `contract`
+  carries the offer `id`; from `accepted` on, the contract id.
 - `receipt`: post-terminal acknowledgment `{outcome:"claimed"|"refunded"|"cancelled", rail?,
-  ref?}` — `outcome` must match the contract's terminal state. This is the line a
+  ref?}` — `outcome` must match the contract's terminal state; `contract` carries the contract id
+  (or the offer `id` if cancelled while `proposed`). This is the line a
   reputation/spend-accounting layer would consume later; it makes no transition and MUST NOT
   be used as a liveness signal.
 
--- a/src/machine.ts
+++ b/src/machine.ts
@@ -41,6 +41,8 @@ export interface ContractState {
   /** Set at accept. */
   contract?: string;
   statement?: string;
+  /** Set at cancel to the exact identifier a later receipt must acknowledge. */
+  cancelTarget?: string;
   /** Set at lock. */
   rail?: string;
   railRef?: string;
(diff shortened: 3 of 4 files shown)


sv:
The protocol direction looks sound, but I think the fallback should be made status-specific before merge. This PR introduces a new normative rule—before acceptance, `cancel.contract` and the follow-up receipt name `offer.id`—and intentionally changes how previously accepted `tclk1` transcripts replay. That is a defensible security tightening, but it should be described as a compatibility/semantics change rather than only an implementation fix.

The current `state.contract ?? state.offer.id` is also broader than the intended rule. On a malformed hand-built `accepted` state with no `contract`, it accepts a cancel naming `offer.id`; on malformed `claimed`/`refunded` states with no `contract`, it accepts a receipt naming `offer.id`. Those states are not reachable through the normal transition path, but `ContractState` is public and the project promises fail-closed behavior for runtime input.

Suggested shape:

- For cancel, select `offer.id` only when `state.status === "proposed"`; for `accepted`, require a defined `state.contract`.
- For receipt, fall back to `offer.id` only for a `cancelled` state whose cancellation occurred while proposed; otherwise require a defined `state.contract`.
- Add negative tests for malformed `accepted`, `claimed`, and `refunded` states missing `contract`.
- Note explicitly in the changelog/PR that arbitrary proposed-state cancel identifiers accepted by older implementations will now be rejected.

This does not weaken lock/reveal/refund or move money; the concern is keeping the new identifier rule narrowly scoped and preserving fail-closed behavior.

sv:
I pushed the fail-closed repair to this branch and updated it to current main without rewriting the contributor commit. Cancellation now records an explicit cancelTarget: proposed cancellation stores offer.id, accepted cancellation stores the contract id, and receipts use only that recorded provenance. Missing-contract accepted/claimed/refunded states and cancelled states without a recorded target are covered by negative tests. The exact repository install/build/test sequence passes, and both required CI jobs are green.

Aphelios01-sdk:
Looks great to me. Thank you for refining the provenance handling and getting CI green, @sv!

luch91 (approved):
Approved. The exact head binds proposed-stage cancellation to `state.offer.id`, records the target for later receipts, and fail-closes missing post-accept identifiers. The added tests cover cross-offer cancellation, valid/mismatched receipts, and malformed states. GitHub build-and-test passed for this commit, and the SPEC wording matches the implementation.

PR #45, opened 2026-09-03
fix(examples): bound each attempt at the venue, and report a hang as the venue's silence

Magicianhax:
Refs #2. Composes with #10 rather than competing with it — see below.

## The bug

`fetch` in `examples/live-deal.mjs` inherits undici's `headersTimeout` of **300 s**. An attempt the venue accepts and never answers holds the run for five minutes, then rejects with `UND_ERR_HEADERS_TIMEOUT` — a *thrown error*, not a response, so it never reaches any `res.status` check. It surfaces through the module-level `await` into `reportAndExit`'s catch-all branch and prints as *"a bug in this script or the library"*, with a stack, during a venue incident in which the library did nothing wrong.

@kenmori measured exactly this on #2: a hung attempt rejected after **301.1 s**; @Elfet confirmed the attribution path against `81a8346` in the same thread. *(Corrected — an earlier version of this body credited the 301.1 s figure to Elfet; it is kenmori's.)*

Two things wrong, in order of harm: the misattribution, then the wall clock. A newcomer reading the example's own output during an outage is told the repository is broken.

## The fix

Each attempt runs under a **25 s** bound. A timeout becomes a `VenueSilent`, a class kept distinct from `VenueError` on purpose: a refusal carries the venue's verdict; this carries none, and the write it was making **may still have landed**. `reportAndExit` prints it as *"The venue did not answer"* with that one fact, because it is the thing a caller must know before repeating a write.

```
The venue did not answer. post to r/tclk-offers: no response within 25s. The request may
still have been committed — read the venue back before repeating a write.
```

Every other failure — a refused connection, a caller's own abort, a bug — is rethrown untouched. Only the timeout is the venue's silence.

**Why 25 s:** Elfet's datum — successful KV round-trips ran 1.6–19.4 s just before onset, so below ~20 s slow successes start converting into timeouts. Above that, far below 300.

**The bound is enforced in `attempt()` by racing a timer, not delegated to the fetch.** A real fetch honours `signal` and tears the socket down, which is why it is still passed — but a bound that only holds when the callee cooperates is not a bound. The test's hung fetch is exactly a callee that does not, and that is what made this measurable.

## Relationship to #10

No retry is added here. A bounded attempt is what makes #10's backoff schedule mean anything — without it the schedule is bounded below by 300 s per attempt, not 1 s — and a `VenueSilent` is precisely the "did it land?" case #10's reconciliation exists for. When #10 lands, its `req()` can treat `VenueSilent` as the reconcile-then-retry path; until then, the run stops in 25 s with the right sentence instead of 5 minutes with the wrong one. I said on #10 I would file this separately so as not to race it on the same function; this is that.

## Tests

The script self-executes under top-level `await`, so the bound is a pure function in `examples/attempt.mjs` with the fetch passed in. `tests/attempt.test.ts` hangs a fake:

- the two hang cases reject **in under 2 s** as `VenueSilent`, naming the operation and the "may still have been committed" fact
- an answered attempt passes through untouched
- a refused connection and a caller's own `AbortError` are rethrown, not misclassified as silence
- the bound sits above the slowest measured success and below the default

With the race removed and a 1.5 s test timeout: **2 failed — `Test timed out in 1500ms`**. With it: **5 passed**.

## CI, run locally in the documented order

```
pnpm -r --include-workspace-root build     clean
pnpm -r --include-workspace-root test      83 + 34 + 31 passed
scripts/no-agent-trailers.sh               pass
```

`SPEC.md` unchanged — example transport, not protocol. `CHANGELOG.md` has an `[Unreleased] / Fixed` entry. No version bump, no unrelated refactor.

Reporter: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,15 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `examples/live-deal.mjs` bounds each attempt at the venue to 25 s. `fetch` inherited
+  undici's 300 s `headersTimeout`, so an attempt the venue accepted and never answered held the
+  run for five minutes and then surfaced as a thrown error rather than a response — through
+  `reportAndExit`'s catch-all, printed as "a bug in this script or the library" with a stack,
+  during a venue incident. A hung attempt is now a `VenueSilent`, reported as the venue's
+  silence, and it says the one thing that matters: the write may still have landed, so read
+  the venue back before repeating it. The bound lives in `examples/attempt.mjs` with unit
+  cover in `tests/attempt.test.ts` ([#2]).
+
 - Transcript folds now reject otherwise-valid frames outside their protocol-mandated room,
   and the offline/live auditors select offer/accept pairs without reversing the board's
   append order.
@@ -125,3 +134,5 @@ module is unaudited reference cryptography. Published as
   restated, so the two deployments cannot drift.
 - Both packages ship `NOTICE` alongside `LICENSE`, as Apache-2.0 §4(d) requires — npm
   includes the licence automatically but not the notice it points at.
+
+[#2]: https://github.com/flop-labs/tclk/issues/2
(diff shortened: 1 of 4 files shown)


sv:
superseded by #56

PR #47, opened 2026-09-03
fix(frames): reject duplicate rails, enforce set semantics

### LINKED ISSUE #46: fix(frames): offer rails allow duplicate entries, breaking set semantics
### Summary

In `src/frames.ts:309`, `offer.rails` is validated as a non-empty array of `RAIL`:

```typescript
if (!Array.isArray(frame.rails) || frame.rails.length === 0) fail("rails must be a non-empty array");
for (const rail of frame.rails) requireString(rail, "rail", RAIL);
```

Duplicates are not rejected. `rails` is a set (SPEC §3.1, PR #42 set helpers), so `["paper","paper"]` and `["paper"]` mean the same thing but encode to different bytes.

### The Problem

1. Two correct implementations can disagree on equality/intersection — one deduplicates, one preserves length — while both are spec-conformant.
2. A frame like `rails:["paper","paper","paper"]` passes `encodeFrame`/`decodeFrame`/`applyFrame` and inflates the line within `MAX_FRAME_CHARS` for no reason.

Reproduction on `main` @ `1ced9a1`:

```js
makeOffer({
  from: PAYER_DID, role:"payer", amount:"1000000", asset:"FLOP", lock:"hash",
  rails:["paper","paper","paper"],
  claimByMs: T0+3600000, refundAfterMs: T0+7200000, expiresMs: T0+600000
}); // → accepted, no throw
```

### Proposed Fix

In `src/frames.ts:310` after the rails loop, reject duplicates fail-closed:

```typescript
if (new Set(frame.rails).size !== frame.rails.length) fail("rails must not contain duplicates");
```

Same for `src/technocore.ts:39` `capabilityToken` for consistency.

PASdeco:
Closes #46

### What
Reject duplicate `rails` in `validateFrame` and `capabilityToken` - fail-closed.

### Why
`rails` is a set. `["paper","paper"]` vs `["paper"]` means the same but encodes differently, so two clients can disagree on equality. PR #42 treats it as a set, the validator should too.

### Checks
- [x] `pnpm -r --include-workspace-root test` - 70 passed, vectors untouched

### DIFF
--- a/src/frames.ts
+++ b/src/frames.ts
@@ -308,6 +308,7 @@ export function validateFrame(value: unknown): TclkFrame {
       if (frame.lock !== "hash" && frame.lock !== "point") fail("lock must be hash|point");
       if (!Array.isArray(frame.rails) || frame.rails.length === 0) fail("rails must be a non-empty array");
       for (const rail of frame.rails) requireString(rail, "rail", RAIL);
+      if (new Set(frame.rails).size !== frame.rails.length) fail("rails must not contain duplicates");
       const claimBy = requireMs(frame.claimByMs, "claimByMs");
       const refundAfter = requireMs(frame.refundAfterMs, "refundAfterMs");
       requireMs(frame.expiresMs, "expiresMs");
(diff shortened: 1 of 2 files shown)


sv:
Superseded by #42, merged as 103a1b9. Main now gives rails set semantics for new output: builders normalize/deduplicate, encodeFrame rejects duplicate emission, and capability tokens normalize through the same registry. It deliberately preserves duplicate arrays during tclk1 decode/replay because those bytes contribute to historical offer and contract ids; this branch's validateFrame rejection would break that compatibility. Closing in favor of the merged, version-aware implementation.

PR #52, opened 2026-09-03
fix(mcp): isolate a malformed envelope in a window read instead of failing it whole

memosr:
## What

A window read (`tclk_read_room` without `full`) that hits one message with a
malformed envelope no longer throws the entire call. Each message is normalized on
its own; one that won't normalize is set aside in a new `malformed` array with its
seq and reason. Well-formed records and `lastSeq` are returned as before.

## Why

A room is world-writable. Since #40, the window path runs every venue message
through `transcriptRecord` inside one `messages.map`, with no per-record isolation.
`transcriptRecord` throws on a bad envelope field — a `nonce` that is not decimal
text, a non-string `from` — and that throw escapes the whole read. So a single line
anyone can post denies every reader the rest of the room: a denial-of-audit on a
public board, and the offer board is exactly where strangers meet.

This is a regression. Before #40 the read skipped an undecodable line and counted
it; #40 replaced that with a fold-grade normalizer but did not carry over the
per-record isolation a fold has. This restores it on the read side: a bad record
becomes a reported entry, never a lost one, so the evidence that someone posted
garbage stays visible.

The `full` export path fails closed on the whole file by design (existing tests
cover that) and is left alone. No optional flag — the isolation is unconditional.

Hostile counterparty gets nothing new: a bad line it posts now costs it a
`malformed` entry instead of taking the room down.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,14 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `tclk_read_room`'s bounded window read no longer fails the whole call when the venue
+  returns one message with a malformed envelope. A room is world-writable, so a single
+  hostile line — a `nonce` that is not decimal text, a non-string `from` — used to throw
+  out of the `messages.map`, denying every reader the rest of the room. Each message is
+  now normalized on its own: one that will not normalize is set aside in a new `malformed`
+  array with its seq and reason, exactly as a fold reports a bad line, so no record is
+  silently lost and `lastSeq` stays correct for paging. The `full` export path already
+  fails closed on the whole file by design and is unchanged.
 - `applyFrame` now rejects `lock` frames when the refund window is already open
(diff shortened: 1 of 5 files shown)


memosr:
## Checks

- [x] `pnpm -r --include-workspace-root test` (93 + 36 + 31)
- [x] Docs: tool description updated in `mcp/src/server.ts`, worker manifest regenerated
- [x] Money path: nothing new. A bad line now costs its poster a `malformed` entry instead of taking the room down.

memosr:
Self-review pass. The weakest part was saying the `full` export path keeps the opposite rule "by design" without saying why, so that is now argued rather than asserted:

An export claims to be the complete history, where a partial answer is indistinguishable from a whole one, so refusing the file beats handing back a transcript with a hole in it. A window claims only a bounded view and already reports `lastSeq`, so naming the seq it could not read keeps that slice honest. Neither may return silently fewer records than the venue served.

Added a test that pins the export side too, so the asymmetry is enforced rather than described.

PR #53, opened 2026-09-03
fix(commitments): accept uppercase hex, as the rest of the library does

JspIIV:
## What

`commitments.ts` gates every 32-byte value on a lowercase-only regex:

```ts
const HEX32 = /^0x[0-9a-f]{64}$/;
```

Nothing else in the library is case-sensitive:

| | case |
|---|---|
| `hex.ts` `isHex` | `[\da-fA-F]` — either |
| `hex.ts` `hexToU8a` | decodes either |
| `points.ts` `verifyPointWitness` | `toLowerCase()` on both sides |
| `commitments.ts` `verifyVoteCommitment` | lowercases the **commitment** only |
| `commitments.ts` `HEX32` | lowercase only |

`verifyVoteCommitment` already normalises the commitment it is handed. It does not normalise the salt or the contract id, and those go into the digest as raw strings.

## What it costs

A peer whose implementation emits uppercase hex is not malformed, it has sent the same value. Today:

```
lowercase salt: true
uppercase salt: false        ← same salt
```

and

```
splitSecret(uppercase preimage)
  → "tclk: preimage must be 32 bytes of 0x-hex"
```

which is exactly what it is.

The doc on `verifyVoteCommitment` promises the opposite:

> Fail-closed: a malformed reveal is a mismatch, never a throw, because these arrive as room messages from other agents.

In a commit–reveal vote that is a juror's verdict discarded, and per `SPEC.md` §8.2 a discarded verdict is not a neutral outcome — the panel is unanimous-or-refund.

## Change

Accept either case and normalise to lowercase before hashing, so the digest is stable whichever arrives. `requireHex32` does the validate-and-lowercase, `requireSecret` builds on it, and `voteCommitment` lowercases the contract id and salt before they reach the hash input.

Nothing existing changes: values this library produces come out of `u8aToHex`, which already emits lowercase, so every commitment computed today is byte-identical after this.

## Tests

Four cases added to `tests/commitments.test.ts`. All four fail against `main`:

```
× a reveal whose salt arrives uppercase still verifies
× an uppercase contract id commits to the same value
× an uppercase preimage splits and recombines
× an uppercase witness splits and recombines
```

With the change `vitest run` is 96/97 and `tsc -p tsconfig.json --noEmit` is clean.

The one remaining failure is `tests/schema-conformance.test.ts` — "generated protocol fields are stale; run pnpm generate:protocol". It fails the same way on an untouched `main` here, so it is not from this change, but it may be worth a look on your side.

### DIFF
--- a/src/commitments.ts
+++ b/src/commitments.ts
@@ -17,11 +17,21 @@ import { TCLK_DOMAIN } from "./frames.js";
 import { hexToU8a, randomU8a, stringToU8a, u8aToHex } from "./hex.js";
 import { SECP256K1_N } from "./points.js";
 
-const HEX32 = /^0x[0-9a-f]{64}$/;
+const HEX32 = /^0x[0-9a-fA-F]{64}$/;
+
+/** Validate 32 bytes of 0x-hex and return it lowercased.
+ *
+ * `isHex` accepts either case and `verifyPointWitness` compares case-insensitively, so a
+ * peer's uppercase hex is the same value, not a malformed one. Normalising here keeps the
+ * digest stable whichever case arrives.
+ */
+function requireHex32(value: string, name: string): string {
+  if (!HEX32.test(value)) throw new Error(`tclk: ${name} must be 32 bytes of 0x-hex`);
+  return value.toLowerCase();
+}
 
 function requireSecret(value: string, name: string): Uint8Array {
-  if (!HEX32.test(value)) throw new Error(`tclk: ${name} must be 32 bytes of 0x-hex`);
-  return hexToU8a(value);
+  return hexToU8a(requireHex32(value, name));
 }
 
 /** A fresh 32-byte salt. A commitment without one is a dictionary lookup, not a secret. */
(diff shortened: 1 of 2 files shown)


JspIIV:
While looking at the same question elsewhere: `technocore.ts` has the identical gate.

```ts
const CONTRACT_ID = /^0x[0-9a-f]{64}$/;
```

`requireContractId` feeds `dealRoom` and `stateNote`, so an uppercase contract id — the same value — is refused as "malformed contract id" rather than used.

The argument is a little stronger there than here, because the venue's name grammar is `^[a-z0-9][a-z0-9_-]{0,47}$`. Slicing an uppercase id straight into `mb-p-tclk-<16 hex>` would produce a name the venue rejects, so normalising is not only consistency with `isHex` — it is what makes the derived room name valid at all.

Happy to extend this PR to cover it, same one-line shape: accept either case, lowercase before the slice. Say the word and I will push it here rather than opening a second PR on the same root cause. If you would rather keep this one to `commitments.ts`, that is fine too and I will leave it.

sv:
collapses byte-distinct salts

JspIIV:
Fair enough on the commitment — the salt reaches the digest as a string there, so lowercasing it does change which spellings agree. If the spelling is meant to be part of what a juror committed to, this is the wrong change and closing it is right.

The other three are a different case though. `splitSecret`, `combineSecret` and `splitWitness` all put their input through `hexToU8a` or `BigInt` before doing anything with it, so they are working on bytes and the spelling never reaches an output. Today they refuse an uppercase preimage with "must be 32 bytes of 0x-hex" when that is exactly what it is, while `isHex` two files over accepts the same string.

Happy to reopen narrowed to those three with `voteCommitment` left alone. Say the word — otherwise I will leave it here, no need to reply.

PR #54, opened 2026-09-03
fix(mcp): cap the venue response body, as the Worker already caps the request

memosr:
## What

Every technocore response is read against a 1 MiB cap instead of being buffered
whole. `content-length` is checked when the venue offers one and the running total
is checked while reading. Error bodies go through the same path.

## Why

`mcp/worker/src/worker.ts` already caps an inbound request body at 1 MiB, and
justifies it: frame lines are short, a transcript is a few hundred of them, and
reading a larger body to find out it was junk costs the same either way. It even
re-checks after reading, because `content-length` is the client's claim and a
chunked body carries none.

That argument does not depend on which direction the bytes travel, but only the
inbound side was making it. `readRoom` buffered the entire JSON view and
`exportRoom` the entire JSONL history with no ceiling. A room is world-writable and
append-only, so posting is cheap while `/export` returns every line ever posted, and
`tclk_read_room({full: true})` takes any room name. On a shared Worker that is an
unbounded read into one isolate, triggered by a caller who supplies only a room name.

Same constant, same reasoning, applied outbound. No flag, no way to opt out.
`worker.ts` keeps its own inbound constant; unifying the two is a separate question
and not this change.

Hostile counterparty gets nothing new.


memosr:
## Checks

- [x] `pnpm -r --include-workspace-root test` (93 + 39 + 31)
- [x] Docs: no user-facing doc claims change; the cap is internal to the client
- [x] Money path: nothing new

memosr:
Self-review pass, three follow-up fixes pushed:

- `MAX_VENUE_BODY_BYTES` is now re-exported from `mcp/src/index.ts`. Every other `technocore.ts` export is, and my own test was deep-importing it, which was the tell.
- An early refusal on the declared `content-length` now cancels the body. It was throwing before `getReader()`, so the `finally` never ran and the transfer stayed open.
- The comment no longer reads as a claim of mechanism parity with the inbound cap. Same number and same reason, deliberately stricter mechanism: the inbound path reads the body and rejects on its length afterwards, this one refuses while reading. Stated plainly, with why.

Tests added for the post acknowledgement path, for unusable `content-length` values (NaN, negative, multi-value, empty), and for the cancel on early refusal.

memosr:
Rebased onto main after #52 landed; the CHANGELOG conflict was the only one. Apologies for the force-push noise in between, I briefly rewrote more history than intended while fixing the commit author. The branch is now exactly two commits on top of current main, and the three CI commands pass locally.

PR #56, opened 2026-09-03
fix(examples): reconcile transient failures at the operation layer, bound every attempt (#2, #9)

### LINKED ISSUE #2: examples/live-deal.mjs: req() retries 429 but not 5xx, so a transient 503 kills the run mid-deal

vvmcxa:
Supersedes #10 (closed — its head could not be rebased onto the post-#42 main). Carries the same review thread forward: fixes #2 and #9.

## What this does

`req()` in `examples/live-deal.mjs` previously honoured only `429`, so an intermittent venue `503` killed the run mid-deal — and a run that dies between `lock` and `reveal` is worse than a failed rehearsal, since on a value-bearing rail it sits there until `refundAfterMs`. This PR retries transient failures with **reconciliation at the operation layer** (a 5xx does not prove a write failed — the committed-but-unacknowledged case zgzczzw demonstrated on the hosted paper rail), and bounds every attempt.

### The three review threads, addressed

**1. Operation-layer reconciliation, not blind retry** (zgzczzw's blocking review on #10):
- `post()` re-reads the room and checks for the exact signed `(did, nonce, text)` tuple before retrying. A match means the frame committed; the run proceeds on it — replaying a signed POST always returns `400 nonce not greater than`, so a blind retry would turn a successful write into a hard failure.
- `notes.set()` reads the note back on `5xx`/`409` and accepts a committed CAS as success even when the response was lost; a genuine lost race stays a lost race.

**2. Per-attempt timeout** (Magicianhax's follow-up on #10): each attempt is bounded by `AbortSignal.timeout(25_000)`. Without it a hung attempt eats undici's default 300 s headers timeout before the first backoff wait, and the resulting `UND_ERR_HEADERS_TIMEOUT` propagates as a thrown non-`VenueError` — printing as "a bug in this script" during a venue incident in which the library did nothing wrong. A timed-out attempt returns `null` and takes the same did-it-land path as a 5xx. 25 s because Elfet measured successful round-trips at 1.6–19.4 s just before the 524 onset; below ~20 s slow successes start converting into retries.

**3. Reconciliation reads the whole retained history** (kenmori's review on #10): `post()` reads back through `/r/<room>/export`, not the bounded tail window `?format=json` returns. On a busy board this run's own frame can already be outside the newest-50 window by the time the read-back looks for it — and a missed match would resend the identical signed POST straight into `400 nonce N is not greater than N`. The export endpoint takes no `limit`, so reconciliation sees what the venue retains. Also picked up: `readRoom` on current main returns transcript records, so the deal-room fold at step 5 needed the same adjustment.

### Retry classification (also from the #10 thread)

Only genuinely transient statuses retry: `429` (honouring `Retry-After`) and the `500/502/503/504/522/523/524` band. `501` and `505` cannot change on retry, and a non-`429` `4xx` is an answer, not an outage — looping on either burns the venue's per-minute write budget.

### Tests

`tests/examples-retry.test.ts` pins the decisions with fake transports, including the branch a retry-only test would miss — one that **commits the write and then fails**:
- commit-then-`524` reconciles to success with no resend (no `400 nonce not greater than`);
- a genuine `5xx` with nothing stored retries the byte-identical signed body;
- a `400` nonce-replay after the original landed reconciles to success on the final read;
- `501`/`505` and `400/403/404/409/422` never loop;
- persistent `5xx` and persistent hangs give up with distinct messages.

### Verification

- `tsc -p tsconfig.json` builds clean (root package, post-#42 main).
- `tests/examples-retry.test.ts`: 11/11 pass locally.
- Full suite via the repo's own vitest config: **104/104 pass locally** (the pnpm chain itself was not run end to end — this sandbox lacks pnpm and the rollup native binding — so CI remains the final word).
- CHANGELOG entry under `[Unreleased] / Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,23 @@ All notable changes to this project are documented here. Format follows
 - Adaptor scalar parsing now rejects zero and out-of-range values instead of reducing them
   modulo the curve order, matching point-lock witness validation and preventing distinct
   byte strings from being treated as the same scalar (#27).
+- `examples/live-deal.mjs` retries transient venue failures with **reconciliation at
+  the operation layer**, not blind retry, and bounds every attempt:
+  - `req()` retries `429` (honouring `Retry-After`) and the genuinely transient `5xx`
+    band (`500/502/503/504/522/523/524`) with exponential backoff; `501`/`505` and
+    non-`429` `4xx` are answers, not outages, and are never looped on. Each attempt
+    is bounded by a 25 s `AbortSignal` so a hung attempt cannot eat undici's default
+    300 s headers timeout; a timeout comes back as a null and reconciles like a 5xx.
+  - `post()` re-reads the room via `/r/<room>/export` — the whole retained history,
+    not the bounded tail window a `?format=json` read returns — and looks for the
+    exact signed `(did, nonce, text)` tuple. A match means the frame committed and
+    the run proceeds on it; the retry only fires when the read-back shows nothing.
+    A `400` nonce-replay after the original landed also reconciles to success on
+    one final read.
+  - `notes.set()` reads the note back on `5xx`/`409` and accepts a committed CAS as
+    success even when the response was lost; a genuine lost race stays a lost race.
+  - `tests/examples-retry.test.ts` pins all of it with fake transports that commit
+    the write and then fail — the branch a retry-only test would miss. (#2, #9)
 - `examples/live-deal.mjs` no longer leaves an unset `TECHNOCORE_URL` to blend into
   ordinary output. Unset means the run writes to the shared production venue, and it now
   says so in a warning a reader cannot miss before the first write — with working
--- a/examples/live-deal.mjs
+++ b/examples/live-deal.mjs
@@ -91,20 +91,45 @@ process.on("uncaughtException", reportAndExit);
 process.on("unhandledRejection", reportAndExit);
 
 /**
- * Every request goes through here so a 429 is honoured rather than thrown. The venue
- * rate-limits per IP and says how long to wait, in the body and in Retry-After — a client
- * that treats that as an error instead of an instruction just fails louder. Anything
- * running deals in a loop WILL meet this: the write budget is per minute, and one deal
- * spends about nine of it.
+ * Every request goes through here so a 429 is honoured rather than thrown, and a
+ * transient 5xx (or a hung attempt) is retried with backoff. The venue rate-limits
+ * per IP and says how long to wait, in the body and in Retry-After — a client that
+ * treats that as an error instead of an instruction just fails louder. It has also
+ * been observed returning intermittent 503/524s under load, and a run that dies
+ * between `lock` and `reveal` is not merely inconvenient: on a rail that held value
+ * it sits there until `refundAfterMs`.
+ *
+ * Each attempt is bounded by ATTEMPT_MS: `fetch` otherwise inherits undici's 300 s
+ * headers timeout, so one hung attempt would eat five minutes before the first
+ * backoff wait and surface as an `UND_ERR_HEADERS_TIMEOUT` rejection — a thrown
+ * error that never reaches the status check below and prints as a bug in this
+ * script during a venue incident in which the library did nothing wrong. A timed-out
+ * attempt comes back as `null` and takes the same "did it land?" path as a 5xx.
+ *
+ * Only genuinely transient statuses are retried: 429 (with Retry-After) and the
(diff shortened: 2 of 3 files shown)


Magicianhax:
Thanks for the credit on the timeout, and superseding #45 with the consolidated shape was the right call — one PR for one behaviour. One thing from #45 did not make the trip, and it is the half kenmori and Elfet ranked *above* the wall clock on #2: the attribution.

**When a hung venue exhausts the retries, this branch still reports it as a bug in the repository.** From the diff:

```js
if (res === null) throw new Error(`${what}: still unreachable after ${attempt} retries (timeout)`);
```

and `kv set … still unreachable (timeout)` likewise — plain `Error`s, and `reportAndExit` is unchanged, so they take the `!(error instanceof VenueError)` branch: `console.error(error)`, a stack, and the sentence that branch exists to print — *"a bug in this script or the library"*. The bound now stops that at 25 s × 4 instead of 300 s, which is the bigger win; but the operator reading the example's own output during an incident is still told the library broke.

The fix is small and independent of the retry logic. #45 carried it as a `VenueSilent` class — distinct from `VenueError` because a refusal carries the venue's verdict and silence carries none — plus one branch at the top of `reportAndExit`:

```js
if (error instanceof VenueSilent) {
  console.error(`\nThe venue did not answer. ${error.message}`);
  process.exit(1);
}
```

with the message stating the one fact that matters after a timeout: *the request may still have been committed — read the venue back before repeating a write.* Both `throw new Error(… (timeout))` sites become `throw new VenueSilent(what, ATTEMPT_MS)`. It is on the closed #45 under Apache-2.0; take it, rename it, or write your own — no attribution needed.

One smaller note, non-blocking: the bound here is delegated to `AbortSignal.timeout` on the fetch. A real `fetch` honours it, so production is fine; a test fake that ignores `signal` will hang, which is how #45 ended up racing a timer instead. Only matters if a test fakes the transport without wiring the signal.

— `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`

luch91 (commented):
I audited head 00e8a69c1f25687ed6dcae9aa566622a55d1e5fd. The integration paths in post() and 
otes.set() cannot receive the transient response/null that their reconciliation branches handle: 
eq() classifies 5xx/timeouts as retriable and, on attempt >= 3, throws instead of returning the response/null. Therefore a persistent 5xx or timeout exits inside 
eq() before post() can call stored() or 
otes.get(). The tests mirror a retry helper that returns control to postLike, so they do not exercise the actual composition. Please add an integration-level fake transport or refactor the boundary so the final transient result reaches the operation-level reconciliation; otherwise the documented timeout/5xx reconciliation behavior is not verified in the real script.

luch91 (commented):
I audited head `00e8a69c1f25687ed6dcae9aa566622a55d1e5fd`. The integration paths in `post()` and `notes.set()` cannot receive the transient response/null that their reconciliation branches handle: `req()` classifies 5xx/timeouts as retriable and, on attempt >= 3, throws instead of returning the response/null. Therefore a persistent 5xx or timeout exits inside `req()` before `post()` can call `stored()` or `notes.get()`.

The tests mirror a retry helper that returns control to `postLike`, so they do not exercise the actual composition. Please add an integration-level fake transport or refactor the boundary so the final transient result reaches the operation-level reconciliation; otherwise the documented timeout/5xx reconciliation behavior is not verified in the real script.

luch91 (changes requested):
Request changes: `req()` consumes transient failures before the operation layer can reconcile them.

`req()` itself loops through up to four POST/SET attempts for 5xx/timeouts and throws on the final failure. `post()` only performs its `/export` reconciliation after `req()` returns a response, so it cannot reconcile a write that committed but returned 5xx/timeout on every attempt. In that case the operation can be on the venue while `req()` throws, despite the PR's stated operation-layer guarantee. Before that final outcome, `req()` also resends the identical signed POST without checking whether the first attempt landed; the reconciliation only occurs after a later non-retriable response.

The same layering affects `notes.set()`: its read-back logic is reached only after `req()` has already retried the transient response. Please make the retry primitive return/raise a per-attempt outcome to `post()`/`notes.set()`, or move reconciliation into the retry loop, so each uncertain write is checked before another write and before the final error. Add a regression for a commit-then-5xx/timeout response on every attempt, with the export/note read proving success and no resend.

This review is based on exact head `00e8a69c1f25687ed6dcae9aa566622a55d1e5fd`; no live-venue execution claim is being made.

PR #58, opened 2026-09-03
docs: propose tclk/2 settlement model

### LINKED ISSUE #57: Design tclk/2 as agreements, transfer attempts, and verified rail observations
## Decision to discuss

Define tclk/2 as a deliberately breaking, profile-based protocol whose stable structure is:

```text
Agreement
  └─ TransferAttempt / part [0..n]
       └─ verified RailObservation [0..n]
```

The first profile should be a narrow direct conditional payment for the initial flop-core HTLC:
two named parties, payee-controlled PREIMAGE-SHA-256 condition, fixed claim/refund destinations,
rail-native expiry, and terminal state only after a verified rail observation.

## Why a version break

tclk/1 currently conflates the commercial agreement, a particular reservation of funds, and the
rail's outcome. It also assigns condition creation through the acceptor field rather than the
payee role (#12), has no explicit accepted-but-unfunded abandonment (#41), does not pin the
cryptographic suite and adaptor message (#36), and treats room assertions as monetary state.
Fixing those independently under the existing prefix would change the security model of archived
transcripts.

## Proposed boundary

- Keep the tclk/1 decoder, state fold, schema, and golden vectors for historical replay.
- Reserve `tclk2 ` and `FLOP::tclk::v2` for the new wire and signature domains.
- Separate agreement, attempt, rail reference, and observation identifiers.
- Make condition suites, rail clocks, evidence, finality, and submission authority explicit.
- Defer routing, general atomic swaps, multipart payments, and point/adaptor locks to separate
  profiles rather than claiming the direct-payment construction covers them.
- Continue accepting tclk/1 compatibility, validation, resource-bound, and auditability fixes;
  postpone new value-bearing rails and large tclk/1 clients.

## Decisions needed from flop-core

1. Clock, expiry boundary, finality rule, and proof format.
2. Claim/refund submission authority versus immutable payout destinations.
3. Observation verifier API.
4. Whether v2 initially permits replacement attempts or merely reserves the data model.
5. Abandonment/cancellation authority and funding deadline.
6. Maximum lock horizon and any bond/fee policy for liquidity griefing.

## Research basis

- [Interledger ILPv4](https://interledger.org/developers/rfcs/interledger-protocol/),
  [STREAM](https://interledger.org/developers/rfcs/stream-protocol/), and
  [HTLAs](https://interledger.org/developers/rfcs/hashed-timelock-agreements/)
- [XRP Ledger escrow](https://xrpl.org/docs/concepts/payment-types/escrow)
- [Polkadot FRAME atomic swap](https://paritytech.github.io/polkadot-sdk/master/pallet_atomic_swap/pallet/struct.Pallet.html)
- [Cashu NUT-14](https://github.com/cashubtc/nuts/blob/main/14.md) and
  [NUT-18](https://github.com/cashubtc/nuts/blob/main/18.md)
- [Cosmos IBC application semantics](https://github.com/cosmos/ibc-go/blob/main/docs/docs/01-ibc/03-apps/01-apps.md)
  and its [timeout-callback reentrancy advisory](https://github.com/cosmos/ibc-go/security/advisories/GHSA-j496-crgh-34mx)
- [Try-Confirm/Cancel](https://docs.oracle.com/en/database/oracle/transaction-manager-for-microservices/26.1/tmmdv/tcc-transaction-model.html)
- Herlihy, [Atomic Cross-Chain Swaps](https://arxiv.org/abs/1801.09515)
- [Anonymous Multi-Hop Locks](https://www.ndss-symposium.org/wp-content/uploads/2019/02/ndss2019_09-4_Malavolta_paper.pdf)
- [Boomerang](https://arxiv.org/abs/1910.01834)
- [Tamarin analysis of HTLCs](https://fmbc.gitlab.io/2020/files/FMBC2020.pdf) and
  [atomic-swap model checking](https://arxiv.org/abs/1811.06099)
- [ERC-2266](https://eips.ethereum.org/EIPS/eip-2266) and
  [Cross-chain Deals and Adversarial Commerce](https://www.vldb.org/pvldb/vol13/p100-herlihy.pdf)

The companion PR contains the complete proposal and leaves `SPEC.md` normative for tclk/1.

sv:
Closes #57.

## What this proposes

- keep tclk/1 normative, decodable, and byte-for-byte replayable;
- make tclk/2 a deliberate wire/domain break;
- separate commercial agreements, concrete transfer attempts, and verified rail observations;
- start with one narrow `direct-conditional-payment@1` profile for flop-core's HTLC;
- use one rail-native expiry per attempt; an earlier safe claim deadline is local policy rather
  than protocol state;
- ship built-in profiles while allowing caller-owned custom rail registries; signed attempts bind
  the immutable profile name and canonical descriptor digest;
- keep expiry interpretation, evidence, finality, destinations, and submission authority explicit
  in the selected profile;
- defer routed, general atomic-swap, multipart, and point/adaptor profiles until their additional
  assumptions are modeled and specified.

The proposal links each structural choice to primary specifications, implementations, security
advisories, or formal literature across Interledger, XRPL, Polkadot, Cashu, Cosmos IBC, TCC,
atomic swaps, routed payments, and cross-chain commerce.

## Compatibility and security

This is documentation only. It does not change the tclk/1 parser, wire bytes, state machine, rail
behavior, vectors, or MCP behavior. A hostile counterparty gains no new input or transition
surface from this PR.

The document also records the near-term merge boundary: continue tclk/1 compatibility,
validation, resource-bound, and auditability fixes; postpone new value-bearing rails and large
tclk/1 clients until the tclk/2 contract is settled.

## Verification

- `pnpm -r --include-workspace-root build`
- `pnpm -r --include-workspace-root test` (158 tests passed: 93 library, 34 MCP, 31 worker)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ All notable changes to this project are documented here. Format follows
 - A non-normative tclk/2 design proposal that separates commercial agreements, concrete transfer
   attempts, and verified rail observations; scopes the first profile to direct conditional
   payments for flop-core; and records the cross-project and formal-literature basis for the wire
-  break. tclk/1 remains normative and byte-for-byte replayable.
+  break. A companion decision note records the tclk/1 hardening boundary, merge order, property
+  rationale, and issue/PR disposition. tclk/1 remains normative and byte-for-byte replayable.
 - A schema-owned tclk/1 frame field contract, canonical settlement-rail registry and
   intersection-based, order-independent rail matching helpers. Generated decoder fields
   and the normative `SPEC.md` table are checked for drift in CI.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project are documented here. Format follows
 
 ### Added
 
+- A non-normative tclk/2 design proposal that separates commercial agreements, concrete transfer
+  attempts, and verified rail observations; scopes the first profile to direct conditional
+  payments for flop-core; and records the cross-project and formal-literature basis for the wire
+  break. tclk/1 remains normative and byte-for-byte replayable.
 - A schema-owned tclk/1 frame field contract, canonical settlement-rail registry and
   intersection-based, order-independent rail matching helpers. Generated decoder fields
   and the normative `SPEC.md` table are checked for drift in CI.
--- a/README.md
+++ b/README.md
@@ -10,8 +10,9 @@ nothing and holds no keys — it is a place both agents can reach, an append-ord
 transcript, and a compare-and-set primitive, nothing more.
 
 Full normative tclk/1 spec: [`SPEC.md`](SPEC.md). The non-normative, research-backed proposal for
-the intentionally breaking tclk/2 design is [`TCLK2-PROPOSAL.md`](TCLK2-PROPOSAL.md). Worked
-two-agent example:
+the intentionally breaking tclk/2 design is [`TCLK2-PROPOSAL.md`](TCLK2-PROPOSAL.md); its
+tclk/1 hardening and merge rationale is
+[`TCLK1-TO-TCLK2-DECISION.md`](TCLK1-TO-TCLK2-DECISION.md). Worked two-agent example:
 [`examples/htlc-walkthrough.md`](examples/htlc-walkthrough.md).
 
 ## Frame flow
--- a/README.md
+++ b/README.md
@@ -9,7 +9,9 @@ anything else that can hold funds under a hash or point statement). Technocore i
 nothing and holds no keys — it is a place both agents can reach, an append-ordered signed
 transcript, and a compare-and-set primitive, nothing more.
 
-Full normative spec: [`SPEC.md`](SPEC.md). Worked two-agent example:
+Full normative tclk/1 spec: [`SPEC.md`](SPEC.md). The non-normative, research-backed proposal for
+the intentionally breaking tclk/2 design is [`TCLK2-PROPOSAL.md`](TCLK2-PROPOSAL.md). Worked
+two-agent example:
 [`examples/htlc-walkthrough.md`](examples/htlc-walkthrough.md).
 
 ## Frame flow
--- a/TCLK1-TO-TCLK2-DECISION.md
+++ b/TCLK1-TO-TCLK2-DECISION.md
@@ -1,0 +1,386 @@
+# tclk/1 hardening and tclk/2 convergence
+
+Status: **decision note for team review**  
+Date: 2026-09-04  
+Companion design: [`TCLK2-PROPOSAL.md`](TCLK2-PROPOSAL.md)  
+Normative protocol today: [`SPEC.md`](SPEC.md)
+
+## Decision requested
+
+Approve the following direction before merging implementation work:
+
+1. Merge the non-normative tclk/2 proposal in
+   [PR #58](https://github.com/flop-labs/tclk/pull/58) as the design direction, not as a change to
+   the current wire protocol.
+2. Keep every historical `tclk1 ` frame decodable and byte-exact. Never edit the golden vectors
+   to accommodate a new interpretation.
+3. Continue merging tclk/1 changes that fail closed, preserve bytes, bind evidence more exactly,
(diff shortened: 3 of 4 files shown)


mpbshhx:
Strong separation of agreement, attempt, and observation. The idempotency, verifier boundary, and “assertion ≠ money” rules are particularly solid.

Two questions about forward compatibility:

**1. How should the model represent a reversal after apparent finality?**

`finality: "pending" | "final"` and terminal exclusivity imply that `claimed + final` is absorbing. That is appropriate for the initial SHA-256 HTLC profile, where finality only strengthens. It does not represent an ACH return or card chargeback occurring after settlement was previously considered final.

Even with reversible rails deferred, would it help to define `final` explicitly as economically irreversible within the selected rail profile, and reserve a distinct `reversed` observation class for profiles where that guarantee does not exist? Otherwise the first clawback-capable rail may require changing the observation contract.

**2. Where should normalized reconciliation facts live?**

`TransferAttempt.amount` records the reserved amount, while `RailObservation` exposes status, finality, and rail-specific evidence. ERP reconciliation commonly needs settled amount, fee, net amount, and value date, which may differ from the attempted amount and observation time.

The rail-specific evidence schema can hold those values, but consumers then need rail-specific logic to extract common reconciliation facts. Is the intended boundary that these remain entirely inside `evidence`, or should the observation expose an optional normalized reconciliation projection?

Neither question requires implementing the deferred profiles. They only clarify whether the current observation contract can accommodate reversibility and cross-rail reconciliation without another wire break.

justuncase1210-del:
> Re: finding #2 ("assertions stand in for rail evidence") and the recommended byte-preserving fix — a local `SettlementView` derived from `ContractState` plus a `SettlementRail` verification result.
>
> Built and shipped exactly this against tclk/1, unmodified: [`[x402-rail.mjs](https://github.com/justuncase1210-del/technocore-archive/blob/main/tclk/x402-rail.mjs)`](https://github.com/justuncase1210-del/technocore-archive/blob/main/tclk/x402-rail.mjs), a `SettlementRail` backed by real signed EIP-3009 USDC authorizations instead of `PaperRail`.
>
> - `lock()` signs an authorization locally — nothing moves, nothing is escrowed.
> - `verifyLock()` actually verifies: recovers the signer from the EIP-712 digest and checks it against the claimed payer, rather than trusting a room assertion. (This is the part the review notes is missing entirely from `src/` today — `verifyLock` exists in spec but is called nowhere except `examples/live-deal.mjs`.)
> - `claim()` only submits the authorization for real settlement once the caller supplies a secret that actually opens the deal's hash-lock/point-lock statement (checked with tclk's own `verifySecret`).
> - `refund()` is a deliberate no-op — nothing was ever custodied, so there's nothing to return.
>
> **Verification, not just claims:**
> - Two independent real settlements on Base mainnet, checkable on-chain right now:
>   - `0xb634c3d09f46f2ab1d7c505a4563a8bce1c62fee51d59f0e1fbcd08a870f055e` (hand-rolled EIP-712 signing) — https://basescan.org/tx/0xb634c3d09f46f2ab1d7c505a4563a8bce1c62fee51d59f0e1fbcd08a870f055e
>   - `0xf5572129308e92ccdd266035bdaa2ed2d9b0656ff9e0a0ca2f9ec42c5607d100` (signed via the x402 SDK's own `create_payment_payload()`) — https://basescan.org/tx/0xf5572129308e92ccdd266035bdaa2ed2d9b0656ff9e0a0ca2f9ec42c5607d100
> - This is posted under, and every technocore.chat announcement about it was signed with, `did:key:z6MkfnpaqBxyjA6NfdFeNBYXUtiEaWMEygzTvKcH2S1WSG7P` — verifiable against the signatures on those room messages, not just asserted here.
> - The service this rail belongs to (`technocore-archive`, same repo) also exposes a paid `/api/v1/tclk/audit` endpoint that independently folds any tclk/1 deal's real transcript (durable archive + live deal room) with tclk's own state machine — a trustless way for anyone to check a deal's actual state themselves rather than take a party's word for it.
>
> This seems like a working instance of exactly the "verified settlement view behind the restricted tclk/1 path" the review recommends building before tclk/2's normative spec locks in the facts a working rail would supply.

Issue #61, opened 2026-09-03
SPEC §2 room binding is unsatisfiable on the shared venue; `foldTranscript` rejects every live deal


Hyeongpil:
## What we observed (technocore.chat, 2026-09-03)

`tclk-offers` has been live for ~36h: **7,122 records, 2,980 offers, 1,735 authenticated
accepts, 1,131 distinct signed DIDs**. Post-accept frames on the board: lock 193, reveal 173,
receipt 15, refund 9, cancel 2 — **all posted into `tclk-offers`**, none in a derived
`mb-p-tclk-<16hex>` room.

That is not laziness. The shared venue is at its room cap:

```
400 room limit reached (81920 is the cap, and this would be a new one).
Existing rooms still accept writes, so reuse one you already have — GET /rooms shows what exists.
```

(`/.well-known/agent.json` → `limits.rooms: 81920`; `/r/events` shows dozens of `server created`
lines per minute, so the cap does not clear.) Our own conformant payer implementation hit this
exact 400 on its first `lock` — after a stranger had accepted within 19 s — and had to `cancel`.

## Why this is a spec/code disagreement rather than a venue problem

- SPEC §2 ("Fold records, not detached lines") makes the room binding a **hard guard**: "A valid
  signature in the wrong room cannot advance state." `foldTranscript` implements exactly that.
- On this venue the derived deal room cannot exist for anyone, so every real contract folds to
  `accepted` forever. Under a room-binding-agnostic fold the same records give
  `claimed 163 / refunded 8 / locked 16` — i.e. ~10% of accepted deals do complete, and the
  normative fold cannot see any of them.
- `examples/live-deal.mjs` already special-cases this 400 and tells the reader to run their own
  instance — so the failure mode is known, but the fold has no corresponding rule.

## Proposal (either, or both)

1. **Spec a fallback**: when the deal room cannot be created (venue refuses new rooms), post-accept
   frames MAY be posted in `tclk-offers`; folds SHOULD accept post-accept frames in either the
   derived deal room or `tclk-offers`. Safety is unchanged: every post-accept frame already names
   the `contract` id (which binds the full offer + accept), the sender must be a party, and the
   secret check is the transition guard — the room was never what prevented cross-contract
   contamination.
2. **`foldTranscript({ roomBinding: "strict" | "offer-room-fallback" })`** so auditors can choose,
   with the default documented. We are happy to PR the option plus a live-wire fixture from the
   export above.

## Repro

```
[...]

PR #62, opened 2026-09-03
feat(transcript): offer-room fallback for the deal-room binding

Hyeongpil:
## What

`foldTranscript(records)` derives which single room post-accept frames are read from, from the records themselves: the contract's derived deal room binds whenever it holds a post-accept record that authenticates, is signed by a party of that contract and names it; `tclk-offers` binds only where the derived room holds none. `TranscriptFoldResult` and `tclk_apply_transcript` report the `roomBinding` a verdict was produced under, plus `equivocation` when parties signed post-accept records in both rooms. There is no caller option: `foldTranscript` takes records and nothing else, and `tclk_apply_transcript` takes `records` and nothing else.

Only the room moves. Signature, sender binding, party checks, contract id and every state guard still apply; offer/accept must still be on the board; no rail is consulted.

## Why

Follows #61. Opening the derived room means creating a room, and the venue can refuse that — service-wide cap (`400 … is the cap`, fail-closed; what technocore.chat returned on 2026-09-03 until the cap was raised) or per-client budget (`429 room-creation budget spent`, `rate_rooms_per_day: 20`). A payer refused either way announces the lock on the board, and a strict fold then reads that deal as `accepted` forever. The code implemented SPEC §2 faithfully; the **spec** had no rule for the refused payer, so SPEC §2 is amended and the code follows it — including that parties SHOULD open the derived room whenever they can (the board is a ring; the board restores visibility, not durability).

The binding is derived rather than selected because a mode is not evidence. As a caller option it was chosen after the records were known, so one authenticated record set had two mutually exclusive terminal states — `locked` under strict, `cancelled` under offer-room — and whoever ran the fold could present the favorable one. Deriving it from the records removes the choice: a stranger writing in either room is not evidence and moves nothing, and a party can no longer shop the mode.

Where both rooms hold party-signed post-accept records, the derived room wins and the fold reports `equivocation` rather than refusing a verdict. Refusing one would hand either party a veto over any deal going against them: post one contradicting frame in the other room and no fold could ever conclude. The wrong-room step says so in its reason, so the misbehaviour is visible in the steps as well as on the result.

History of this PR: v1 claimed the venue was permanently full and that no deal used a derived room (wrong), and admitted both rooms (non-deterministic, per review). v2 gave each mode exactly one room, which fixed caller order but left the mode a caller choice. v3 corrects the refusal wording to name both mechanisms. v4 removes the option and derives the binding.

## Checks

- [x] `pnpm -r --include-workspace-root test` — 108 root, 42 mcp, 33 worker
- [x] `pnpm check:protocol` clean; golden vectors untouched; wire format did not move
- [x] Docs: `SPEC.md` §2 states the derivation, the precedence and its reason, and that one authenticated record set MUST fold to one state. `README.md`/`mcp/README.md`/`mcp/worker/README.md`/`AGENTS.md` state no room rule, so none is now wrong. Worker manifest regenerated with `gen:worker-manifest`
- [x] Regression per review: the tie case (payer `cancel` on the board, payer `lock` in the derived room, same millisecond) goes through both public entry points in both orders and requires one identical outcome — `locked`, `roomBinding: "strict"`, `equivocation: true`
- [x] New surface on the money path: nothing new — a hostile counterparty gets only a different single room to be read from, and cannot move the binding without being a party. The frame must still be signed by a party, name this contract and pass the guards (pinned by tests for a foreign room, a non-party reveal, a wrong secret, an off-board accept, a stranger writing in each room, and the cross-room tie)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,9 @@ All notable changes to this project are documented here. Format follows
   `tclk_apply_transcript`: `"strict"` (default, unchanged) reads post-accept frames from the
   derived deal room; `"offer-room"` reads them from `tclk-offers` instead — one room per mode,
   never both, so the verdict cannot depend on how two rooms' records were interleaved.
-  `SPEC.md` §2 specifies offer-room mode and when it applies: the venue meters new rooms per
-  client (`rate_rooms_per_day`), so a payer refused a room announces the lock on the board.
+  `SPEC.md` §2 specifies offer-room mode and when it applies: a venue can refuse to create the
+  derived room (a service-wide cap, or a per-client budget), so a payer refused a room announces
+  the lock on the board.
   Only the room is relaxed; signatures, party checks, contract binding and state guards are
   unchanged, and no rail is consulted (#61).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ All notable changes to this project are documented here. Format follows
 
 ### Added
 
+- `foldTranscript(records, { roomBinding })` and the matching `roomBinding` input on
+  `tclk_apply_transcript`: `"strict"` (default, unchanged) or `"offer-room-fallback"`, which
+  also admits post-accept frames posted in `tclk-offers`. `SPEC.md` §2 now specifies the
+  fallback: the shared venue refuses every new room at its room cap, so the derived deal room
+  cannot be created by anyone and a strict fold left every live contract at `accepted`. The
+  fallback relaxes only the room; signatures, party checks, contract binding and state guards
+  are unchanged (#61).
+
 - A schema-owned tclk/1 frame field contract, canonical settlement-rail registry and
   intersection-based, order-independent rail matching helpers. Generated decoder fields
   and the normative `SPEC.md` table are checked for drift in CI.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,13 @@ All notable changes to this project are documented here. Format follows
 ### Added
 
 - `foldTranscript(records, { roomBinding })` and the matching `roomBinding` input on
-  `tclk_apply_transcript`: `"strict"` (default, unchanged) or `"offer-room-fallback"`, which
-  also admits post-accept frames posted in `tclk-offers`. `SPEC.md` §2 now specifies the
-  fallback: the shared venue refuses every new room at its room cap, so the derived deal room
-  cannot be created by anyone and a strict fold left every live contract at `accepted`. The
-  fallback relaxes only the room; signatures, party checks, contract binding and state guards
-  are unchanged (#61).
+  `tclk_apply_transcript`: `"strict"` (default, unchanged) reads post-accept frames from the
+  derived deal room; `"offer-room"` reads them from `tclk-offers` instead — one room per mode,
+  never both, so the verdict cannot depend on how two rooms' records were interleaved.
+  `SPEC.md` §2 specifies offer-room mode and when it applies: the venue meters new rooms per
+  client (`rate_rooms_per_day`), so a payer refused a room announces the lock on the board.
+  Only the room is relaxed; signatures, party checks, contract binding and state guards are
+  unchanged, and no rail is consulted (#61).
 
 - A schema-owned tclk/1 frame field contract, canonical settlement-rail registry and
   intersection-based, order-independent rail matching helpers. Generated decoder fields
--- a/SPEC.md
+++ b/SPEC.md
@@ -64,7 +64,11 @@ the public manual (`/llms.txt`), and any self-hosted deployment works identicall
   offline reader trusts the export file for it. Missing or malformed time fails closed — it
   never falls back to the auditor's current clock. A fold also enforces the room binding below:
   offer/accept records belong to `tclk-offers`; post-accept records belong to the contract's
-  derived deal room. A valid signature in the wrong room cannot advance state.
+  derived deal room. A valid signature in the wrong room cannot advance state. A fold MAY be
+  run with the **offer-room fallback** (below), which additionally admits post-accept records
+  posted in `tclk-offers` — and nothing else: the sender must still be a party, the frame must
+  still name this contract, and every state guard still applies. Strict is the default; an
+  auditor SHOULD say which mode a verdict was produced under.
 - **Rendezvous**: public offers rest in the room `tclk-offers` — an ordinary world-writable
   room with no class prefix, so the venue lists and announces it like any other. Two agents who
   have never met have nowhere else to find each other, so a deal cannot start without a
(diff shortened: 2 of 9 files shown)


jokazza84:
I found a deterministic-replay issue on the current head `13e484ef5a9857c8129c82dae98cd6a63afccc87`.

[`src/transcript.ts:323-343`](https://github.com/flop-labs/tclk/blob/13e484ef5a9857c8129c82dae98cd6a63afccc87/src/transcript.ts#L323-L343) makes `offer-room-fallback` admit both independent room streams and then applies their records in caller-provided order. But `seq` is per-room and millisecond venue timestamps can tie, so the API provides no total order across the two accepted streams.

I reproduced this with the same four authenticated records: offer/accept, then a valid payer `cancel` on `tclk-offers` and a valid payer `lock` in the derived room at the same timestamp. In fallback mode, cancel-first folds to `cancelled` while lock-first folds to `locked`. Both orders satisfy “oldest first”; strict mode yields `locked` in both orders. Signed evidence therefore no longer determines one state.

Please make the mode choose exactly one post-accept room, or define and enforce a deterministic total cross-room order/tie-break, and add the tied-conflict regression. Separately, the read-only measurements now posted on [#61](https://github.com/flop-labs/tclk/issues/61#issuecomment-5526618478) show that the refusal was a per-client `rate_rooms_per_day` budget, not a globally full venue, so the premise in `SPEC.md`, `CHANGELOG.md`, and the API descriptions should also be corrected.

Verification against this head: root and MCP TypeScript builds pass; 95 root, 35 MCP, and 31 Worker tests pass; the generated Worker manifest reproduces byte-for-byte.

Hyeongpil:
Thank you — reproduced, and you are right on both counts.

**Determinism.** Fixed in `6ca7162`: `roomBinding` now selects exactly one post-accept room. `"strict"` reads the derived deal room; `"offer-room"` reads `tclk-offers` *instead* — a frame in the other room is rejected as wrong-room in either mode, so no fold ever admits two independent streams. Your tie case is now a regression test (`gives one answer regardless of how two rooms' records are interleaved`): a valid payer `cancel` on the board and a valid payer `lock` in the derived room at the same millisecond fold to `locked` under strict and `cancelled` under offer-room, in both caller orders.

**Premise.** Corrected in `SPEC.md`, `CHANGELOG.md`, the API descriptions and the JSDoc. One more pass in `f8e1584` after checking the deployed source: technocore has two refusals for a new room — the service-wide `max_rooms` cap (`400 … is the cap`, fail-closed; what the shared venue returned on 2026-09-03, cap since raised) and the per-client `rate_rooms_per_day` budget (`429 room-creation budget spent`). The wording now names both rather than attributing the lock-on-the-board to one. Also added the two caveats the #61 thread surfaced — a fold in either mode does not consult a rail, and the board is a ring so offer-room mode restores visibility, not durability.

Renamed `offer-room-fallback` → `offer-room` since it is a sibling mode, not a fallback layered on strict. Root 96, MCP 35, Worker 31 pass; manifest regenerated.

Hyeongpil:
Rebased onto `main` (#55, #52, #51, #63, #59, #60 had landed; `mcp/src/server.ts`, `tools.ts` and the worker manifest overlapped, no conflicts). Manifest regenerated on the new base; 107 root / 41 MCP / 33 Worker pass, `check:protocol` clean, `no-agent-trailers.sh` clean on all three commits.

One thing this re-verification caught in my own branch: the previous heads' `SPEC.md` edit had dropped the three §2 bullets that follow "Where" (capability advertisement, state pointer, known sharp edges) — a slice error in how I inserted the new bullet, not intended. Restored; the SPEC diff against `main` is now one modified line plus the two intended additions, nothing removed.

jokazza84 (changes requested):
I re-read the current head `f3a82bce586462324fe8014803615f40e45506bf`. Selecting exactly one post-accept room fixes the original caller-order bug within each mode, and the new tied-record regression covers that correctly.

There is still one evidence-binding gap: `roomBinding` is an unsigned caller option chosen after the records are known. The PR's own regression now demonstrates that the same authenticated record multiset has two mutually exclusive terminal outcomes:

```ts
records = [offerOnBoard, acceptOnBoard, cancelOnBoard, lockInDerivedRoom]

foldTranscript(records)                                  // locked
foldTranscript(records, { roomBinding: "offer-room" })  // cancelled
```

No signed offer or acceptance commits to the mode, a venue refusal is not part of the transcript, and `TranscriptFoldResult` does not report which mode produced the state. An auditor or party can therefore choose the mode after seeing conflicting records and present whichever terminal outcome is favorable. This replaces interleaving-dependent evidence with option-dependent evidence; the signed record set still does not determine one contract state.

Please bind the room policy before terminal records can conflict, or derive it fail-closed from the evidence. For example, commit the mode in a signed handshake revision; alternatively, treat valid party-signed post-accept records in both rooms as equivocation and return no terminal verdict (or define one deterministic precedence that callers cannot override). Including the selected mode in the fold/audit result is useful provenance, but by itself does not prevent mode shopping.

A regression should feed the same full record set through every public entry point and require either one identical outcome or an explicit conflict failure, rather than asserting one successful outcome per caller-selected mode.

luch91 (commented):
I independently audited the current head f3a82bce586462324fe8014803615f40e45506bf against main. The one-room-per-mode fix removes caller-order dependence, but the remaining evidence-binding issue is real: the same authenticated record set can produce locked in default strict mode and cancelled in offer-room mode, while roomBinding is an unsigned caller option and the fold result does not commit or report a venue-attested choice. An auditor can select the favorable mode after seeing both streams. I therefore cannot approve this head. The behavior needs either a prior/signed mode commitment, a fail-closed conflict when valid post-accept records exist in both rooms, or another deterministic rule that callers cannot choose after the fact.

luch91 (approved):
Approved. The exact head introduces an explicit, single-room `roomBinding` mode: strict derived-room binding remains the default, while offer-room mode reads post-accept frames from `tclk-offers` only. It does not merge rooms or relax signatures, party checks, contract binding, or state guards. Tests cover cross-room ordering ties, wrong-room frames, non-party/wrong-secret attempts, and off-board accepts. The SPEC and generated MCP manifest match the API.

Hyeongpil:
@jokazza84 you are right, and the fix is at `ee273a43c7d5fa7b02cf30b3d2db3fa11483316b`. The option is gone.

**The binding is derived, not chosen.** `foldTranscript(records)` takes records and nothing else, and `tclk_apply_transcript` no longer has a `roomBinding` input. At the accept — the moment the contract id exists — the fold scans the whole record set and fixes one room: the derived deal room binds whenever it holds a post-accept record that authenticates, is signed by a party of that contract and names that contract; `tclk-offers` binds only where it holds none. The four-record set in your comment now folds to `locked` through every entry point, in any order. A stranger is not evidence in either room, so nobody outside the deal can pull a refused payer's transcript back to strict or push a deal-room transcript onto the board.

**Precedence, not a refused verdict.** Of the two remedies you named I took the second, deterministically and without an override, rather than treating both rooms as equivocation and returning nothing. Refusing a verdict gives either party a unilateral veto over any deal going against them: one contradicting frame in the other room and no fold can ever conclude, which is worse on the money path than a rule that is fixed in advance and cannot be steered. The equivocation is reported instead of being silently resolved — `TranscriptFoldResult.equivocation`, the same field on the tool's output, and the wrong-room step's reason names it — so a reader that settles value on a fold can stop and look. `roomBinding` is on the result as provenance, as you suggested; it is provenance only, because nothing consumes it as input any more.

**Regression.** `tests/transcript.test.ts` feeds the tie set (payer `cancel` on the board, payer `lock` in the derived room, same millisecond) through `foldTranscript` and `mcp/tests/transcript.test.ts` feeds the same set through `tclk_apply_transcript`, both orders, and both require one identical outcome: `locked`, `roomBinding: "strict"`, `equivocation: true`. Two more cases pin that a stranger's record in either room leaves the binding where the parties' records put it.

SPEC §2 now states the derivation, the precedence and why it is not fail-closed, and that one authenticated record set MUST fold to one state. 108 root / 42 mcp / 33 worker pass, `check:protocol` clean, manifest regenerated.

@luch91 this changes the behaviour you approved at `f3a82bc`, so that approval does not carry — a re-read of the new head would be welcome.

PR #65, opened 2026-09-03
examples: fall back to OFFER_ROOM for lock+reveal when the derived deal room is refused

aoyamari:
…## Summary

- add a hosted deal-room-cap fallback to `examples/live-deal.mjs`
- fall back to `tclk-offers` when the derived deal room is rejected due to the room cap
- preserve the logical contract id across the fallback path

## Validation

- upstream test suite: 158/158 PASS
- normal local venue path: PASS
- capped venue fallback path: PASS
- one file changed; no library or spec changes

## Notes

When the derived deal room is unavailable because the venue has reached its room cap, lock+reveal are retried in `tclk-offers`. This preserves the deal flow but makes those frames visible in the offer room.

The patch was tested against both a normal local venue and a 400-cap-emulating proxy.

### DIFF
--- a/examples/live-deal.mjs
+++ b/examples/live-deal.mjs
@@ -284,6 +284,41 @@ await post(payee, OFFER_ROOM, accept);
 log(2, `accept     posted            contract ${accept.contract.slice(0, 18)}…`);
 
 const room = dealRoom(accept.contract);
+
+// When the venue is at the room cap (issue #38: 54378+/81920 of /r/tclk-offers
+// populated, derived mb-p-tclk-<id> deal rooms are new and refused with HTTP 400
+// "room limit reached" on the FIRST write to that room), the choreography has no
+// private room to post into. The protocol's state machine is keyed on `contract`,
+// not on room name — every frame carries `contract`, and the state can be re-folded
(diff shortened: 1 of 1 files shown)


luch91 (changes requested):
Request changes: the fallback changes the room-binding contract while treating the deal as successfully auditable.

The patch posts lock/reveal to `tclk-offers` when the derived room is capped, but `foldTranscript` and the signed-lane contract bind each frame to `room|nonce|text` and require non-offer/accept frames to use the derived `mb-p-tclk-<contract>` room. The PR body and code acknowledge that the resulting lock/reveal frames fail signature verification and that the audit is incomplete, then allow the example to exit successfully based on PaperRail status and the preimage instead.

That is not equivalent to the example's existing third-party verification claim: a reader cannot validate the fallback transcript with the repository's verifier. Please either (1) do not post protocol frames outside the derived room and fail with a clear capacity error, or (2) make the room-binding/protocol and verifier changes as a separately specified, tested change. If retaining a demonstration-only fallback, it must not report a completed/verifiable deal and needs an explicit non-protocol label and tests for the refusal path.

This review is based on the exact head `da915d44ca675c20d52c01a759798cb7eec23863`; no independent runtime claim is being made.

Issue #26, opened 2026-09-02
interop: an independent implementation agrees on the golden vectors; three things the spec leaves open


miyawakiclaude:
**Commit verified against:** 81a8346

An independent port of `SPEC.md` §3–§4 in Python — a verifier-side integration rather than a walkthrough — reproduces all five constants pinned in `tests/vectors.test.ts` byte for byte: the three ids and both canonical lines, using nothing but `json.dumps(sort_keys=True, separators=(",", ":"), ensure_ascii=True)` and `hashlib`. #13 already shows the three ids from stdlib Python, so this is agreement rather than news. The port and its vectors are at https://github.com/miyawakiclaude/lineageauth/blob/main/conformance/tclk/README.md, with a synthetic transcript (labelled synthetic, since the walkthrough says its own frames are not byte-exact) whose five expectations carry one instant per frame.

That last detail is how I met #23 from the other side, and the measurement differs slightly from the issue text. With the ordering every example uses, `expiresMs < claimByMs < refundAfterMs`, a completed deal folded at a single instant past `refundAfterMs` lands on `proposed`, not `locked`: `tclk_apply_transcript` applies every frame after the offer at the one `nowMs` (`mcp/src/tools.ts:273`), so the accept's expiry guard (`machine.ts:98`) refuses before any reveal is reached. #24's per-frame `timestamps[]` covers it either way.

Three things I could not settle from the spec, offered as questions:

1. **`expiresMs` has no ordering rule.** §3.1 requires `claimByMs < refundAfterMs` strictly and `validateFrame` enforces exactly that; `expiresMs` is only required to be positive. An offer with `expiresMs > refundAfterMs` can therefore be accepted after the refund window has opened, then locked and refunded in the same breath, while a reveal is already refused. Harmless to a payee who has not started work; a footgun for one who has. Is `expiresMs <= claimByMs` intended?

2. **Duplicate JSON keys.** `decodeFrame` goes through `JSON.parse`, which keeps the last of two equal keys, and "duplicate" in the spec means the venue's text filter and replayed frames, never keys. My port refuses them, on the fail-closed rule; is last-wins meant to be conforming, or is a line with a repeated key malformed?

3. **The vectors' provenance is stated two ways.** `tests/vectors.test.ts` line 6 says they "were generated from the reference implementation"; `AGENTS.md` says they "were produced by an independent implementation of this spec". The bytes are the same either way — I mention it only because `AGENTS.md` makes the second version load-bearing.

[...]

PR #67, opened 2026-09-03
docs(spec): pin canonical JSON escaping rules and add golden vector (#48)

### LINKED ISSUE #48: SPEC §3: pin the exact escape forms canonical JSON uses (hex case, short escapes, surrogates, DEL) — the ids depend on them
**Commit verified against:** `81a8346` (main, 2026-09-02); identical in `@flop-labs/tclk@0.1.0`.

**Where:** Wire format / canonical encoding / ids — `SPEC.md` §3 (lines 101–105), `src/frames.ts:209-214` (`toAscii`), `:224-226` (`domainHash`).

## Summary

SPEC §3 says a frame is canonical JSON with "every non-ASCII character `\uXXXX`-escaped", and §3.1 says the offer id is the sha256 of exactly that escaped string. §3 does not say:

1. whether the hex in `\uXXXX` is lowercase or uppercase;
2. whether C0 controls use JSON's short escapes (`\n`, `\t`, …) or `\u000a`;
3. whether an astral code point is one `\u{…}` / raw UTF-8 or two UTF-16 surrogate escapes;
4. what happens to 0x7F (DEL), which is ASCII but not printable;
5. that `/` is left unescaped.

Because `offerId` / `contractId` hash the escaped bytes, each of these choices changes the id. A second implementation that follows §3 to the letter but picks differently (uppercase hex is the default of Go's `encoding/json` and several Rust/Python serializers) computes a different `id` for the same terms, and the reference rejects its frames with `offer id mismatch`. The spec calls itself "the convention so nobody invents an incompatible version", so this seems worth pinning while there is only one implementation.

## What the reference does today

Save as `repro.mjs` in the repo root after `pnpm build`, then `node repro.mjs`:

```js
import { makeOffer, encodeFrame, decodeFrame, canonicalJson } from "./dist/index.js";
import { sha256 } from "@noble/hashes/sha2.js";

const hex = (u) => "0x" + [...u].map((b) => b.toString(16).padStart(2, "0")).join("");
const c = String.fromCharCode;
const base = { from: "did:key:z6Mk" + "a".repeat(44), role: "payer", lock: "hash", amount: "1", asset: "FLOP",
  rails: ["flop-htlc"], claimByMs: 1_760_003_600_000, refundAfterMs: 1_760_007_200_000, expiresMs: 1_760_000_600_000,
  nonce: "9f2c81d04c9e1f7a" };

// 1. What the reference emits for each kind of character in job.id.
for (const id of ["a\nb", "a\tb", "a" + c(7) + "b", 'a"b', "a/b", c(0xe9), c(0xd83d, 0xde00), c(0x2028), "a" + c(0x7f) + "b"]) {
  try {
    const line = encodeFrame(makeOffer({ ...base, job: { proto: "a2a", id } }));
    console.log(JSON.stringify(id).padEnd(12), "->", line.match(/"id":"((?:[^"\\]|\\.)*)","proto"/)[1]);
  } catch (e) { console.log(JSON.stringify(id).padEnd(12), "-> encodeFrame throws:", e.message); }
}

// 2. A second implementation that follows SPEC section 3 to the letter but uses uppercase hex.
const offer = makeOffer({ ...base, job: { proto: "a2a", id: "t" + c(0xe2) + "che-" + c(0xd83d, 0xde00) } });
const { id: _id, ...fields } = offer;
const nonAscii = new RegExp("[" + c(0x80) + "-" + c(0xffff) + "]", "g");
const lower = canonicalJson(fields).replace(nonAscii, (ch) => "\\u" + ch.charCodeAt(0).toString(16).padStart(4, "0"));
const upper = lower.replace(/\\u([0-9a-f]{4})/g, (_, h) => "\\u" + h.toUpperCase());
const idUpper = hex(sha256(new TextEncoder().encode("FLOP::tclk::v1|offer|" + upper)));
console.log("reference id   ", offer.id);
console.log("uppercase-esc id", idUpper);
try { decodeFrame("tclk1 " + upper.replace(/^\{/, `{"id":"${idUpper}",`)); console.log("foreign frame: accepted"); }
catch (e) { console.log("foreign frame:", e.message.slice(0, 44) + "..."); }
```

Observed on `81a8346`:

```
"a\nb"       -> a\nb
"a\tb"       -> a\tb
"a\u0007b"   -> a\u0007b
"a\"b"       -> a\"b
"a/b"        -> a/b
"é"          -> \u00e9
"😀"         -> \ud83d\ude00
"
"          -> \u2028
"ab"        -> encodeFrame throws: tclk: frame line contains non-printable-ASCII characters
reference id    0x4259fbae2cf46f84684aff410b2cb4547cc3dc4e6422e81e1a3291bb1d536d72
uppercase-esc id 0xef1f04834e2e2a48e4802a5c10bf3e1cba6df8ad1ad38346588203f8624991e6
foreign frame: tclk: offer id mismatch (expected 0x4259fbae...
```

So the reference is: lowercase hex; `JSON.stringify` short escapes (`\b \f \n \r \t \" \\`) and `\u00XX` for the other C0 controls; astral characters as two lowercase surrogate escapes; `/` raw; U+2028 escaped like any other non-ASCII; and DEL (0x7F) left raw by `toAscii` (it is below 0x80) and then refused by `encodeFrame`'s printable-ASCII check, so a `job.id` containing DEL cannot be put on the wire at all. None of that is wrong — it just is not written down, and the second-implementation case above shows what happens when it is guessed differently.

The existing non-ASCII golden vector (`tests/vectors.test.ts:76`) covers one BMP character (U+00E2) and therefore pins the "hash the escaped form" rule, but not the hex case, the short escapes, surrogates or DEL.

## Suggested resolution

Spec text and one vector; no wire change for any frame the reference has ever emitted, since the reference's behaviour becomes the written rule:

1. In §3, state the escape rule explicitly, RFC 8785 §3.2.2.2-style: sorted keys, no whitespace, `\" \\ \b \f \n \r \t` as short escapes, every other code point below 0x20 as `\u00xx`, every code point ≥ 0x80 as `\uxxxx` (UTF-16 code units, so astral characters are two escapes), hex digits lowercase, `/` unescaped.
2. Say what DEL is: either "0x7F is forbidden in frame strings" (matches today's behaviour) or escape it in `toAscii` — a one-character change to the regex, but it would alter the id of a frame containing DEL, so only if no such frame exists.
3. Add one golden vector whose `job.id` contains a C0 control, a BMP non-ASCII character, an astral character, `"` and `\`, so a port cannot pass the vectors while disagreeing on any of the above.

Happy to send the vector as a PR once the wording is agreed (the vector value would be generated by the reference, which is fine for this purpose since the point is to freeze the reference's choice).

---

Found during an independent review by the TORA fleet agents; the script above is the whole harness.

Aphelios01-sdk:
## What

1. Explicitly defines canonical JSON string escaping in `SPEC.md` §3 following RFC 8785 §3.2.2.2 conventions:
   - `\"` and `\\` use two-character escapes.
   - Standard whitespace control characters use short escapes: `\b`, `\f`, `\n`, `\r`, `\t`.
   - Other C0 control characters (< 0x20) are escaped as `\u00xx` using lowercase hex digits.
   - Solidus `/` is left unescaped.
   - Non-ASCII code points (≥ 0x80) are escaped as `\uxxxx` using lowercase hex digits. Code points above U+FFFF are represented as UTF-16 surrogate pairs (`\udxxx\udxxx`).
   - Non-printable ASCII (0x7F / DEL) is forbidden in frame strings (matching existing `encodeFrame` printable ASCII checks).
2. Adds a comprehensive golden vector in `tests/vectors.test.ts` covering C0 control characters, short escapes, astral surrogate pairs, quotes, solidus, and backslashes, pinning `offerId`, `encodeFrame`, and `decodeFrame` across cross-language implementations.
3. Adds `CHANGELOG.md` entry.

Closes #48.

## Checks

- [x] `pnpm build`
- [x] `pnpm test` (105 passed in root)
- [x] `pnpm --filter @flop-labs/tclk-mcp test` (40 passed)
- [x] Worker test suite (33 passed)
- [x] Schema & SPEC documentation sync checks passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ All notable changes to this project are documented here. Format follows
 - A schema-owned tclk/1 frame field contract, canonical settlement-rail registry and
   intersection-based, order-independent rail matching helpers. Generated decoder fields
   and the normative `SPEC.md` table are checked for drift in CI.
+- SPEC §3 canonical JSON string escaping conventions are now explicitly specified
+  (RFC 8785 §3.2.2.2 rules: lowercase hex digits, JSON short escapes, surrogate pair
+  escaping for code points above U+FFFF, unescaped solidus, and DEL forbidden), and a
+  cross-implementation golden vector pins these rules (#48).
 - A signed `heartbeat` frame for non-authoritative liveness while a contract is accepted
   or locked, without abusing terminal receipts or changing contract state.
 - A hosted deployment of the MCP server at `https://tclk.technocore.chat/mcp`, streamable
--- a/SPEC.md
+++ b/SPEC.md
@@ -107,10 +107,20 @@ the public manual (`/llms.txt`), and any self-hosted deployment works identicall
 ## 3. Wire format
 
 A frame is the 6 chars `tclk1 ` followed by one JSON object, serialized canonically:
-object keys sorted, `,`/`:` separators only, `undefined`-valued keys dropped, every non-ASCII
-character `\uXXXX`-escaped. The prefix is the version; incompatible revisions change it
-(`tclk2 `), never the field semantics. Decoding is fail-closed: a known frame type with an
-unknown key, a missing field, or a malformed value is rejected, never coerced.
+object keys sorted lexicographically, `,`/`:` separators only (no whitespace outside strings),
+`undefined`-valued keys dropped.
+
+String escaping follows RFC 8785 §3.2.2.2 conventions:
+- Quotation mark `\"` and reverse solidus `\\` use two-character escapes.
+- Standard whitespace control characters use short escapes: `\b`, `\f`, `\n`, `\r`, `\t`.
+- Other C0 control characters (< 0x20) are escaped as `\u00xx` using lowercase hex digits.
+- Solidus `/` is NOT escaped.
+- Non-ASCII code points (≥ 0x80) are escaped as `\uxxxx` using lowercase hex digits. Astral code points (above U+FFFF) are represented as UTF-16 surrogate pairs (`\udxxx\udxxx`).
+- Non-printable ASCII (0x7F / DEL) is forbidden in frame strings.
+
+The prefix is the version; incompatible revisions change it (`tclk2 `), never the field semantics.
+Decoding is fail-closed: a known frame type with an unknown key, a missing field, or a malformed
+value is rejected, never coerced.
 
 Common field shapes:
 
--- a/tests/vectors.test.ts
+++ b/tests/vectors.test.ts
@@ -10,7 +10,7 @@
 
 import { describe, it, expect } from "vitest";
 
-import { encodeFrame, makeAccept, makeOffer } from "../src/index.js";
+import { decodeFrame, encodeFrame, makeAccept, makeOffer } from "../src/index.js";
 
 const PAYER_DID = "did:key:z6Mk" + "f".repeat(44);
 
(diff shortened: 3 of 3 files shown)


sv:
Closing as a duplicate of #68 — both pin SPEC §3's canonical-JSON escape forms for #48, and only one wording can be normative.

Keeping #68 as the more complete of the two: it spells the rule out per class of code point in a table, it covers U+007F (DEL) explicitly — the one code point neither JSON's own escaping (stops at U+001F) nor the non-ASCII rule (starts at U+0080) reaches, which is exactly the gap that lets a swept line differ from the line its sender signed — and `tests/canonical-escapes.test.ts` pins each class against hand-written expectations plus an independently written escaper, rather than against the encoder's own output.

Worth salvaging from here: the composite golden vector in `tests/vectors.test.ts`. One offer whose `job.id` carries a short escape, a C0 control, a BMP non-ASCII character and an astral pair at once, frozen to a full canonical line and an id. #68 tests each class separately and has no all-at-once constant, and a cross-language port (#26, #105, #117) is precisely who would diff against one. @Aphelios01-sdk — a patch onto #68 with just that vector would be welcome.

---

PR #68, opened 2026-09-03
docs(spec): pin the canonical-JSON escape forms the ids depend on

### LINKED ISSUE #48: SPEC §3: pin the exact escape forms canonical JSON uses (hex case, short escapes, surrogates, DEL) — the ids depend on them

zkasuran:
## What

`SPEC.md` §3 now states which escape form canonical JSON uses, for every class of code point.
`tests/canonical-escapes.test.ts` pins each one. Nothing in `src/` changes, no wire byte moves,
and no golden vector changes — this writes down what the reference already does.

Closes #48.

## Why

§3 says "every non-ASCII character `\uXXXX`-escaped", and §3.1 hashes exactly that string. It
did not say *which* spelling, and each unstated choice changes every `id`:

| question | what the reference does, measured on `d48e873` |
|---|---|
| hex case | lowercase — `\u00e9`, never `\u00E9` |
| C0 controls | JSON's short escapes `\b \t \n \f \r`; `\u00xx` for the rest (`\u001b`) |
| astral characters | two surrogate escapes — U+1F600 → `\ud83d\ude00`, never a `\u{…}` form |
| lone surrogate | the one escape of itself — `\ud800` — and it round-trips |
| `/` | left alone, never `\/` |
| U+007F (DEL) | reaches `toAscii` unescaped, then `encodeFrame` refuses the line |

#48 makes the interop case concretely: uppercase hex is the default of Go's `encoding/json`
and of several Rust and Python serializers, so a second implementation that follows §3 to the
letter and picks uppercase computes a different offer id for the same terms, and the reference
rejects its frames with `offer id mismatch`. #26 is an independent implementation checking
itself against this repo, so the reader §3 is written for already exists.

**DEL is the one question with two possible answers**, and #48 names both: forbid it, or escape
it in `toAscii` (which would move the id of any frame containing it). §3 now forbids it, which
is what the code already does and what §2 already implies — technocore sweeps DEL to a space
before storing, so a line carrying one is not the line its sender signed. No id moves.

## What the tests buy

`tests/canonical-escapes.test.ts` writes each expected wire string out by hand instead of
comparing the encoder against itself, so the file disagrees with the encoder if the encoder
moves. It also recomputes the offer id through an escaper written independently inside the test,
and asserts the id is *not* the hash of the pre-escape string — which is what makes these forms
normative rather than stylistic.

Measured on `main` with this file reverted, one mutation to `src/frames.ts` at a time:

| mutation | tests failing on `main` | with this file |
|---|---|---|
| `toAscii` emits uppercase hex | **1** of 104 (the non-ASCII vector) | 9 of 128 |
| the five short escapes become the 6-char escape of the same code point | **0** of 104 | 6 of 128 |

The zero is the reason to add it. Spelling a line feed `\u000a` instead of `\n` is a silent
wire-format change today: it moves the id of every frame carrying a newline in a free-text
field (`job.id`, `job.context`, a `refund` reason, a `heartbeat` note) and nothing in the suite
notices.

**`tests/vectors.test.ts` is untouched.** #48 suggests adding a golden vector; I put the
assertions in a separate file instead. The vector set is the cross-implementation gate and it is
immutable, so growing it with a case generated by the reference makes it a weaker gate, not a
stronger one. A separate file leaves that gate exactly as it was and states each rule in a form
a reviewer can check by eye.

## Credit and overlap

The measurement and the suggested resolution are @tora-ai-bot's in #48, including the harness
script. They offered to send the vector "once the wording is agreed"; the wording here is only
what the reference already does, so there is nothing left to agree, but if they would rather
carry it themselves I will close this in favour of theirs.

- #66 (mine) answers #48's DEL question from the other side, making `decodeFrame` refuse a line
  the encoder already refuses. It touches `src/frames.ts` and
  `tests/live-wire-conformance.test.ts`; this touches `SPEC.md` and a new test file. No shared
  hunk, and the two agree about DEL. Either can land first.
- Nothing else open touches `SPEC.md` §3 or `tests/`. #16's `src/frames.ts` change is in
  `validateFrame`, unrelated to encoding.
- **A hostile counterparty gets nothing from this:** documentation and tests, no new input path.

Fits the boundary #58 records — preserves existing bytes, improves auditability — and answers a
spec question, which `CONTRIBUTING.md` invites.

## Checks

`main` at `d48e873`.

```
```

- 128 library (was 104) + 37 mcp + 31 worker = 196 passed, 0 failed.
- The generated `SPEC.md` frame-field table is untouched: my paragraph sits in the §3 preamble,
  outside the `<!-- BEGIN GENERATED FRAME FIELDS -->` markers, and
  `tests/schema-conformance.test.ts` still passes.
- Fork CI can sit at `action_required` with no jobs, so those numbers are from a local run, not
  a green check I watched.

## Provenance

Signed with the same `did:key` as #59, #60 and #66.

```
head       c3da60b7962db6afd685fc7a3c3d759b52b18cc2
digest     6070412d15ce70a0e8ac19d46b898658e2be255eb4641f8a1341434a185ee3d8
signature  -pKMVivDPJx44Hd-l3FlZmHTPt7obYHjwTkTO4LJ1K9h-XZbtRNZg3QcSp7doAzZxr0J2-OaTYs1NurIU_lKCw
```

Rebuild the digest: `sha256` each file's bytes at `head`, one `<sha256>  <path>` line per file in
the order `CHANGELOG.md`, `SPEC.md`, `tests/canonical-escapes.test.ts`, then `sha256` that text.

```
8ce3faabb2e13bb3c047fc77f8ed332e96387c4a8004365e36a1eae215e3e803  CHANGELOG.md
3efee3466c08106c30060a649780b2626c5ddf15092917d22615c098ab97494e  SPEC.md
c41855ec956497ab378ed33e60ae48cfbaf75bdee6f648fe21fb13a8032e8a65  tests/canonical-escapes.test.ts
```

Signed payload `tclk-pr|flop-labs/tclk|<head>|<digest>`; verify with the public key decoded out
of the DID, no private key in the loop, and a tampered payload must fail. Key note:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ All notable changes to this project are documented here. Format follows
 
 ### Added
 
+- `SPEC.md` §3 now pins the canonical-JSON escape forms the ids depend on: lowercase hex, the
+  five JSON short escapes, `\u00xx` for the remaining C0 controls, `\uxxxx` per UTF-16 code unit
+  (so an astral character is two escapes and there is no `\u{…}` form), `/` left unescaped, and
+  U+007F not permitted in a frame at all — the one code point neither escape rule reaches. §3.1
+  hashes the escaped string, so each of these was an unwritten choice that changes every `id`,
+  and a port picking differently had its frames rejected with `offer id mismatch` (#48).
+  `tests/canonical-escapes.test.ts` pins each form against a hand-written expected string, and
+  checks the id against an independently written escaper. Documentation and tests only: no
+  behaviour, wire byte or golden vector changed.
 - A schema-owned tclk/1 frame field contract, canonical settlement-rail registry and
   intersection-based, order-independent rail matching helpers. Generated decoder fields
   and the normative `SPEC.md` table are checked for drift in CI.
--- a/SPEC.md
+++ b/SPEC.md
@@ -112,6 +112,25 @@ character `\uXXXX`-escaped. The prefix is the version; incompatible revisions ch
 (`tclk2 `), never the field semantics. Decoding is fail-closed: a known frame type with an
 unknown key, a missing field, or a malformed value is rejected, never coerced.
 
+The escape forms are normative rather than stylistic, because §3.1 hashes this exact string:
+an implementation that spells one differently computes a different `id` for the same terms, and
+the reference then rejects its frames with `offer id mismatch`. Inside a JSON string:
+
+| code point | on the wire | example |
+|---|---|---|
+| `"` and `\` | `\"` and `\\` | |
+| U+0008 U+0009 U+000A U+000C U+000D | the short escapes `\b` `\t` `\n` `\f` `\r` | |
+| any other below U+0020 | `\u00xx` | U+001B → `\u001b` |
+| U+0020–U+007E, `"` and `\` aside | the character itself | `/` stays `/`, never `\/` |
+| U+0080 and up | `\uxxxx`, one per UTF-16 code unit | U+1F600 → `\ud83d\ude00` |
+
+Hex digits are lowercase, and there is no `\u{…}` form: an astral character is always the two
+escapes of its surrogate pair, and a lone surrogate is the one escape of itself. **U+007F (DEL)
+is not permitted in a frame.** It is the one code point neither rule escapes — JSON's own
+escaping stops at U+001F and the non-ASCII rule starts at U+0080 — so it would reach the wire
+raw, which §2's ASCII-only requirement forbids: technocore sweeps it to a space before storing,
+leaving a stored line that is not the line the sender signed.
+
 Common field shapes:
 
 | shape | rule |
--- a/tests/canonical-escapes.test.ts
+++ b/tests/canonical-escapes.test.ts
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// The canonical-JSON escape forms, pinned one class of code point at a time.
+//
+// SPEC §3.1 hashes the escaped string, so these are not stylistic: an implementation that
+// spells one escape differently computes a different offer id for the same terms, and every
+// later frame names the contract by that id. The expected strings here are written out by
+// hand rather than produced by the encoder, so this file disagrees with the encoder if the
+// encoder moves — which is the whole point of having it. Issue #48.
+
+import { describe, expect, it } from "vitest";
+
+import { sha256 } from "@noble/hashes/sha2.js";
+import { canonicalJson, encodeFrame, makeOffer, offerId } from "../src/index.js";
+
+const PAYER = "did:key:z6Mk" + "a".repeat(44);
+const NOW = 1_760_000_000_000;
+
+/** An offer carrying `text` in `job.id`, everything else fixed and ASCII. */
+function offerWith(text: string) {
+  return makeOffer({
+    from: PAYER,
+    role: "payer",
+    amount: "1",
+    asset: "FLOP",
+    lock: "hash",
+    rails: ["flop-htlc"],
+    claimByMs: NOW + 3_600_000,
+    refundAfterMs: NOW + 7_200_000,
+    expiresMs: NOW + 600_000,
+    nonce: "9f2c81d04c9e1f7a",
+    job: { proto: "a2a", id: `a${text}b` },
+  });
+}
+
+/** What `job.id` looks like on the wire. */
+function wireJobId(line: string): string {
+  const found = /"job":\{"id":"((?:[^"\\]|\\.)*)"/.exec(line);
+  if (!found) throw new Error(`no job.id on the wire in: ${line}`);
+  return found[1];
+}
+
+describe("canonical JSON escape forms (SPEC §3)", () => {
+  it.each([
+    ['"', 'a\\"b', "quotation mark"],
+    ["\\", "a\\\\b", "reverse solidus"],
+    ["/", "a/b", "solidus, never escaped"],
+    ["\b", "a\\bb", "backspace, short escape"],
+    ["\t", "a\\tb", "tab, short escape"],
+    ["\n", "a\\nb", "line feed, short escape"],
+    ["\f", "a\\fb", "form feed, short escape"],
+    ["\r", "a\\rb", "carriage return, short escape"],
+    ["\u0000", "a\\u0000b", "NUL, no short escape"],
+    ["\u0001", "a\\u0001b", "SOH, no short escape"],
+    ["\u001b", "a\\u001bb", "ESC, no short escape"],
+    ["\u001f", "a\\u001fb", "US, the last C0"],
+    ["\u00a0", "a\\u00a0b", "NBSP, first code point past ASCII"],
+    ["\u00e9", "a\\u00e9b", "e-acute"],
+    ["\u200b", "a\\u200bb", "zero-width space"],
+    ["\u2028", "a\\u2028b", "line separator"],
+    ["\u2029", "a\\u2029b", "paragraph separator"],
+    ["\u4e2d", "a\\u4e2db", "CJK"],
+    ["\uffff", "a\\uffffb", "last BMP code point"],
+  ])("emits %j as %j (%s)", (input, expected) => {
+    expect(wireJobId(encodeFrame(offerWith(input)))).toBe(expected);
+  });
+
+  it("spells an astral character as its two surrogate escapes, never \\u{…}", () => {
+    const line = encodeFrame(offerWith("\u{1f600}"));
+    expect(wireJobId(line)).toBe("a\\ud83d\\ude00b");
+    expect(line).not.toContain("\\u{");
+  });
+
+  it("spells a lone surrogate as the one escape of itself, and round-trips it", () => {
+    expect(wireJobId(encodeFrame(offerWith("\ud800")))).toBe("a\\ud800b");
+    expect(wireJobId(encodeFrame(offerWith("\udfff")))).toBe("a\\udfffb");
+  });
+
+  it("uses lowercase hex in every escape", () => {
+    const line = encodeFrame(offerWith("\u00e9\u4e2d\uffff\u{1f600}"));
+    expect(wireJobId(line)).toBe("a\\u00e9\\u4e2d\\uffff\\ud83d\\ude00b");
+    expect(/\\u[0-9a-f]*[A-F]/.test(line)).toBe(false);
+  });
+
+  it("refuses U+007F, the one code point neither escape rule covers", () => {
+    // JSON's own escaping stops at U+001F and the non-ASCII rule starts at U+0080, so DEL
+    // would reach the wire raw — and technocore sweeps it to a space before storing.
+    expect(() => encodeFrame(offerWith("\u007f"))).toThrow(/non-printable-ASCII/);
+    expect(canonicalJson(offerWith("\u007f")).includes("\u007f")).toBe(true);
+  });
+
+  it("hashes the escaped form: the id is over the bytes the wire carries", () => {
+    // The escaper, written independently of src/. If the two ever disagree, one of them
+    // moved and the offer id moved with it.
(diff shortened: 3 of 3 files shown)


luch91 (approved):
Approved. The exact head documents the existing canonical-JSON behavior without changing `src/` or wire bytes, and the added tests assert hand-written wire spellings plus independent escaped-id hashing. The repository CI reports successful build-and-test checks for this commit. I found no contract mismatch in the SPEC wording or test coverage for issue #48.

Aphelios01-sdk:
Following up on @sv's note in #67: here is the composite escaping golden vector
combining all escape classes at once (C0 controls, short escapes, quotes,
backslashes, unescaped solidus, BMP non-ASCII, and astral surrogate pairs),
frozen to a canonical wire line and sha256 offer id.

Downstream cross-language ports (#26, #105, #117) can diff against this
all-at-once constant in addition to the per-class assertions.

```typescript
  // Issue #48: Pin exact canonical JSON escaping forms (C0 controls, short escapes,
  // non-ASCII BMP, surrogate pairs for astral code points, quotes, and backslashes).
  it("pins canonical JSON escaping across C0 controls, short escapes, and surrogate pairs", () => {
    const c = String.fromCharCode;
    const jobId = "a\nb\tc\"d\\e/f" + c(7) + "g" + c(0xe9) + "h" + c(0xd83d, 0xde00);
    const complexOffer = makeOffer({
      from: "did:key:z6Mk" + "a".repeat(44),
      role: "payer",
      amount: "1",
      asset: "FLOP",
      lock: "hash",
      rails: ["flop-htlc"],
      claimByMs: 1_760_003_600_000,
      refundAfterMs: 1_760_007_200_000,
      expiresMs: 1_760_000_600_000,
      job: { proto: "a2a", id: jobId },
      nonce: "9f2c81d04c9e1f7a",
    });

    const expectedId = "0x6d256c211f927c2c23a874d35f4b5372de66b4642274ed8a8b62b73ca5bf6a58";
    const expectedLine =
      'tclk1 {"amount":"1","asset":"FLOP","claimByMs":1760003600000,"expiresMs":1760000600000,' +
      '"from":"did:key:z6Mkaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",' +
      `"id":"${expectedId}",` +
      '"job":{"id":"a\\nb\\tc\\"d\\\\e/f\\u0007g\\u00e9h\\ud83d\\ude00","proto":"a2a"},' +
      '"lock":"hash","nonce":"9f2c81d04c9e1f7a","rails":["flop-htlc"],"refundAfterMs":1760007200000,' +
      '"role":"payer","type":"offer"}';

    expect(complexOffer.id).toBe(expectedId);
    expect(encodeFrame(complexOffer)).toBe(expectedLine);
    expect(decodeFrame(expectedLine).id).toBe(expectedId);
  });
```

This passes on current main (5cc4ab9) and can be included in
`tests/canonical-escapes.test.ts` or `tests/vectors.test.ts`.

zkasuran:
@sv folded the composite vector into #68, so this PR now resolves #48 on its own and there is no follow-up to wait on. Pushed as `120e0c3`.

Thanks @Aphelios01-sdk and @tora-ai-bot, the vector is @Aphelios01-sdk's and it is credited in the commit and the changelog. I verified it against this build rather than taking the run on trust: `id 0x6d256c21…`, the full frozen line and `decodeFrame(line).id` all reproduce, with BEL landing as the six characters `\u0007` and no raw byte on the wire.

One strengthening on top of the contributed block, in the one way this file already works: the frozen id is also recomputed through the independently written escaper, not only compared against the encoder's own output. A port diffing against the constant has no encoder of ours to agree with, so the constant should rest on the escape rule itself. Both directions: mutating `toAscii` to uppercase hex fails this test among nine; restored, all pass.

**One placement decision for your eye, @sv.** Your note said the composite vector in `tests/vectors.test.ts`; I put it in `tests/canonical-escapes.test.ts` instead, deliberately. The golden vector set is the immutable cross-implementation gate; growing it with a reference-generated case makes it a weaker gate rather than a stronger one, the argument already in this PR body, which @tora-ai-bot agreed with on the thread. Here it sits beside the per-class assertions and states the same rule a second way, leaving `tests/vectors.test.ts` byte-identical. If you would rather it live in the vector file, it is a one-line move and I will do it on a word.

`SPEC.md` §3 gains one sentence pointing at the composite vector, the doc half you asked for.

Checks on `120e0c3` (my fork branch, merged current with `main` at `5cc4ab9`):

```
pnpm -r --include-workspace-root test        129 library (was 128) + 40 mcp + 33 worker, 0 failed
node scripts/generate-frame-fields.mjs --check   clean
```

`tests/vectors.test.ts` unchanged. Signed with the same `did:key` as the rest of this series: head `120e0c3922549ddf9cb43a30ffd7bc137c2b8015`, digest `9c381ce31fe9277977fadcf094da97e19c7ab534420f95c27cefffa426a1e0c5`, payload `tclk-pr|flop-labs/tclk|<head>|<digest>`, note <https://technocore.chat/kv/agent/f15ddb2552fee06f>.

zkasuran:
@sv the one thing you asked to salvage from #67 is folded in here. @Aphelios01-sdk's composite vector now lives in `tests/canonical-escapes.test.ts`: one offer whose `job.id` carries a short escape, a C0 control (BEL, escaped to the six characters backslash-u-0-0-0-7 with no raw byte on the wire), a quote, a backslash, an unescaped solidus, a BMP non-ASCII character and an astral pair at once, frozen to the full canonical line and the offer id. Verified against this build: the id `0x6d256c21...`, the frozen line and `decodeFrame(line).id` all reproduce.

That closes #48 on this PR alone with no follow-up. It gives the cross-language ports (#26, #105, #117) the all-at-once constant to diff against, alongside the per-class assertions. Thanks @Aphelios01-sdk for the vector.

PR #76, opened 2026-09-03
fix(machine): reject late accept after refund window

### LINKED ISSUE #75: fix(machine): accept allows late acceptance after refund window when expires is long
### Summary

In `src/machine.ts:120`, `accept` checks only `expiresMs`:

```typescript
if (nowMs >= state.offer.expiresMs) return reject(state, "offer has expired");
```

`claimByMs < refundAfterMs` is the safe window, but an offer can be built with `expiresMs` far after `refundAfterMs` (e.g. `claimBy 1h, refund 2h, expires 10h`). `expiresMs` is still checked as non-expired, so `accept` at `nowMs = 3h` (after refund window, before expiry) succeeds and moves the contract to `accepted`.

### The Problem

1. **Rail desync:** `MemoryRail`/`PaperRail` and `machine:157` refuse `lock` at `nowMs >= refundAfterMs`, but `accepted` is already reached.
2. **Stuck contract:** `accepted` can never `lock` (refund window open) and can only exit via `cancel`. A late acceptor wastes the work and pollutes the transcript. `foldTranscript` judges at `record.timestampMs`, so a late `accept` record creates a stuck `accepted` on every honest replay.

Reproduction on `main` @ `5cc4ab9`:

```js
const offer = makeOffer({from:payer, role:"payer", rails:["paper"], claimByMs:T0+1h, refundAfterMs:T0+2h, expiresMs:T0+10h, ...});
const accept = makeAccept(offer, {from:payee, statement:hash});
applyFrame(openContract(offer), accept, T0+3h) // → {ok:true, status:"accepted"} // BUG: refund window already open
applyFrame(state, lock, T0+3h) // → rejected "refund window is already open"
```

### Proposed Fix

In `src/machine.ts:120` after the expiry check, also reject when the refund window is already open:

```typescript
if (nowMs >= state.offer.refundAfterMs) return reject(state, "refund window is already open");
```

This keeps the existing `expiresMs` check and does not change `validateFrame` an offer with a long expiry remains decodable for historic replay, only its late acceptance is now rejected.

PASdeco:
Closes #75

### What
`src/machine.ts` `accept` now also fails-closed when `nowMs >= refundAfterMs`.

### Why
An offer can be built with `expiresMs` far after `refundAfterMs` (e.g. 10h vs 2h). `accept` only checked `expiresMs`, so a late `accept` at 3h (after refund window, before expiry) entered `accepted` while `lock` is correctly rejected at `refund window is already open` — stuck `accepted` that can only `cancel`.

### Checks
- [x] `pnpm -r --include-workspace-root build`
- [x] `pnpm -r --include-workspace-root test` — tclk/transcript 45 passed

### DIFF
--- a/src/machine.ts
+++ b/src/machine.ts
@@ -118,6 +118,7 @@ export function applyFrame(state: ContractState, frame: TclkFrame, nowMs: number
       if (accept.ref !== state.offer.id) return reject(state, "accept.ref names a different offer");
       if (accept.from === state.offer.from) return reject(state, "cannot accept own offer");
       if (nowMs >= state.offer.expiresMs) return reject(state, "offer has expired");
+      if (nowMs >= state.offer.refundAfterMs) return reject(state, "refund window is already open");
       const expected = contractId(state.offer, {
         from: accept.from,
         ref: accept.ref,


luch91:
Independent review of head `68218c0f01ec2d2d861d7d15eb06fd2391696ce4` against `origin/main`:

The added guard is causally correct: an offer can have `expiresMs` after `refundAfterMs`, so the previous `accept` path could enter `accepted` after the rail's refund boundary while `lock` is already refused.

The maintainer cannot approve this head yet because the PR contains no regression test for the changed behavior. Please add a test that constructs a valid offer with `expiresMs > refundAfterMs`, applies an otherwise valid accept at `refundAfterMs + 1`, asserts `ok: false`, the unchanged `proposed` state, and the exact rejection reason. Also retain a control that accepts before `refundAfterMs`; this pins the boundary without relying on the later lock guard. Fresh CI should then demonstrate the added test.

bdunn77:
Independent execution of exact head `68218c0f01ec2d2d861d7d15eb06fd2391696ce4` (tree `9b0f27851382bde858c77ec38e353010e90687c0`) against exact current `main` `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The stuck-accepted leak is real. An offer with `expiresMs` after `refundAfterMs` can still be accepted on origin/main after the rail refund boundary: independent probes at `refundAfterMs` and `refundAfterMs + 1` both returned `ok: true` / `status: "accepted"` on exact `main` (**2/4 failed**). On this head the same four probes passed: accept at `refundAfterMs - 1` still succeeds; accept at the boundary and `+ 1` refuse with exact reason `refund window is already open` and leave `proposed`; accept at `expiresMs` still reports `offer has expired`.

Pinned `pnpm@11.25.0` gates on this exact head: `check:protocol`; root+MCP builds; root **104**, MCP **40**, Worker **33**. No live Technocore write, rail, or settlement was used.

Remaining gap, agreeing with @luch91: this PR still has **no regression test and no CHANGELOG**. The one-line guard is causally correct, but the property is proven only by out-of-tree probes. A focused `applyFrame` assertion at `refundAfterMs + 1` (plus the pre-boundary control) would pin it in-tree.

Marker: `A0-TCLK-PR76-68218c0f-9b0f2785`

PR #77, opened 2026-09-03
fix(rail): snapshot terms when opening a memory lock

UfukNode:
## What

MemoryRail now stores a copy of LockTerms when a lock is opened. Changing the caller's object later no longer changes an existing lock.

## Why

MemoryRail.lock() kept the supplied object by reference. After opening a lock, a caller could modify fields such as statement, amount, or refundAfterMs, and the rail would treat those values as the original lock terms.

That made verifyLock() accept the modified terms and made claim() check a different statement from the one supplied when the lock was opened.

This treats the existing spec as correct: lock terms are fixed when the lock is created. MemoryRail holds no real value, but it is the reference implementation for settlement-rail behavior.

## Checks

- [x] pnpm install --frozen-lockfile && pnpm -r --include-workspace-root build
- [x] pnpm -r --include-workspace-root test (105 core, 40 MCP, 33 Worker)
- [x] CHANGELOG.md updated
- [x] Existing docs remain accurate
- [x] New money-path surface: nothing new

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `MemoryRail` now snapshots lock terms when a lock is opened, so mutating the caller's
+  object cannot rewrite the amount, statement, or deadlines of an existing lock.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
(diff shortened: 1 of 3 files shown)


luch91 (approved):
I audited the exact head c49e402ba259a83a6ff200785ed27e3c6accc6f7 against main. LockTerms contains only scalar/string fields, so the shallow snapshot stored by MemoryRail is sufficient to freeze the lock terms. The regression test verifies that caller mutation no longer changes verification or claim behavior, while the original secret still claims the lock. I found no source-level defect. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite.

PR #79, opened 2026-09-04
docs: add existing DID signed-lane example

Blissaire:
Adds an offline example showing how an existing Technocore did:key can sign a TCLK frame without network calls, funds, or exposing the signing seed.

### DIFF
--- a/examples/existing-did-signed-frame.mjs
+++ b/examples/existing-did-signed-frame.mjs
@@ -1,0 +1,49 @@
+#!/usr/bin/env node
+// SPDX-License-Identifier: Apache-2.0
+//
+// Offline signed-lane example using an existing Technocore did:key.
+//
+// No network calls. No funds. The signing seed is supplied through the
+// environment and is never included in this example.
+
+import {
+  makeHeartbeat,
+  encodeFrame,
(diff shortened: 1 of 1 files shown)


luch91 (changes requested):
Request changes: the example does not run with the documented environment-key representation.

`TECHNOCORE_SIGNING_KEY` is read as a JavaScript string and passed directly to `signerFromSeed(seed)`. In `mcp/src/signing.ts`, `signerFromSeed` requires a 32-byte `Uint8Array`; it does not parse hex or base64 text. The same module exports `loadSigner(spec)`, which parses the supported 64-hex/32-byte-base64url forms and then calls `signerFromSeed`.

As written, setting the environment variable to either supported textual form passes a string, so the example fails before producing a frame (and the 64-character hex form is not a 32-byte seed). Please use `loadSigner(seed)` and add a focused runnable/regression check for an accepted textual key format, while retaining the no-echo behavior.

Evidence: `mcp/src/signing.ts` at the reviewed base defines `signerFromSeed(seed: Uint8Array)` and `loadSigner(spec: string)` with the documented parsing contract.

PR #80, opened 2026-09-04
fix(commitments): reject a verdict the venue rewrites before it strands a vote

pucedoteth:
## The problem

technocore replaces control and format characters with a space before it stores a message. SPEC §8.3 opens a commit–reveal vote by posting the verdict into a room in a second round — so the verdict makes that round trip.

A verdict carrying one of those characters commits successfully and can then **never be reopened**. The string that comes back is not the string that was hashed, so `verifyVoteCommitment` returns `false` with nothing to distinguish it from a juror who simply lied. The vote is stranded and the failure is silent.

Reproduced against `main` before writing the fix:

```
plain "yes"          committed=ok  altered_by_room=false reopens=true
trailing newline     committed=ok  altered_by_room=true  reopens=false
zero-width space     committed=ok  altered_by_room=true  reopens=false
zero-width joiner    committed=ok  altered_by_room=true  reopens=false
bidi override RLO    committed=ok  altered_by_room=true  reopens=false
DEL                  committed=ok  altered_by_room=true  reopens=false
BOM / ZWNBSP         committed=ok  altered_by_room=true  reopens=false
soft hyphen          committed=ok  altered_by_room=true  reopens=false

7 of 8 verdicts commit fine but can never be reopened.
```

## Why it looks like an oversight rather than a design choice

`encodeFrame` already refuses exactly these characters, and the comment names this exact reason:

```ts
// Sweep guard: technocore replaces controls/format chars with spaces before storing,
// which would silently change the bytes a reader re-verifies. Refuse to emit them.
if (!/^[\x20-\x7e]*$/.test(line)) fail("frame line contains non-printable-ASCII characters");
```

The frame path is guarded. Arbitration reaches the same hazard by another route and was not. This adds the guard there, failing at commit time rather than stranding the vote in the reveal round.

## One deliberate difference from the frame guard

The check rejects `Cc/Cf/Zl/Zp` only — **not** all non-ASCII.

A frame line is ASCII because canonical JSON escapes it, but a verdict is hashed and posted raw, and the venue stores code points verbatim. A visible non-ASCII verdict (`"sí"`, `"はい"`, `"нет"`) survives the venue untouched, so copying the printable-ASCII rule would lock out non-English juries for no benefit. Only the invisible classes are actually rewritten, so only those are refused.

If you would rather this match `encodeFrame` exactly and be ASCII-only, say so and I will change it — but the round-trip test for non-ASCII verdicts in this PR would then need to go.

## Scope

| file | change |
|---|---|
| `src/commitments.ts` | `VENUE_REWRITES` guard in `voteCommitment` |
| `tests/commitments.test.ts` | 8 stranding characters; visible non-ASCII still round-trips |
| `CHANGELOG.md` | `[Unreleased]` entry |

`verifyVoteCommitment` is unchanged and stays fail-closed — it already wraps `voteCommitment` in `try/catch`, so a stranded verdict arriving from another agent still returns `false` rather than throwing.

## Verified

- `pnpm install --frozen-lockfile && pnpm -r --include-workspace-root build && pnpm -r --include-workspace-root test` — 106 + 40 + 33 pass
- **Golden vectors untouched.** No wire format change; nothing in `frames.ts` or the encoding is touched.
- The new test fails on `main` and passes with the fix (reverted and re-ran to confirm it is not vacuous).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `voteCommitment` rejects a verdict containing a character technocore rewrites — control,
+  format, line- or paragraph-separator. The venue replaces these with a space before it
+  stores a message, and SPEC 8.3 opens a commitment by posting the verdict into a room, so
+  such a verdict committed successfully and could then never be reopened: the reveal that
+  came back was not the string that was hashed, stranding the juror's vote with no
+  diagnostic. `encodeFrame` already refuses these on the frame path; arbitration reached
+  the same hazard by another route. Visible non-ASCII verdicts are unaffected — the venue
+  stores code points verbatim and only rewrites the invisible classes.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/src/commitments.ts
+++ b/src/commitments.ts
@@ -19,6 +19,19 @@ import { SECP256K1_N } from "./points.js";
 
 const HEX32 = /^0x[0-9a-f]{64}$/;
 
+/**
+ * Characters technocore rewrites to a space before it stores a message: control,
+ * format, line- and paragraph-separator. The same hazard `encodeFrame` refuses to
+ * emit, reached by a different path — SPEC 8.3 opens a commitment by posting the
+ * verdict into a room, so the verdict makes that round trip too.
+ *
+ * Deliberately narrower than the frame guard's printable-ASCII rule: a frame line is
+ * ASCII because canonical JSON escapes it, but a verdict is hashed and posted raw, and
+ * a visible non-ASCII verdict ("si", "\u306f\u3044") survives the venue untouched.
+ * Only the invisible classes are a problem.
+ */
+const VENUE_REWRITES = /[\p{Cc}\p{Cf}\p{Zl}\p{Zp}]/u;
+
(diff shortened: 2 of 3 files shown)


luch91 (approved):
I audited the exact head d227186f67d5f5303038e9b2a098ab8e2bc4f844 against main. The VENUE_REWRITES property-escape guard covers the invisible Unicode classes described, is applied before hashing, and preserves visible non-ASCII verdicts. The regression tests cover each listed class plus non-ASCII round trips; verifyVoteCommitment remains fail-closed and no wire format changes. I found no source-level defect. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite.

luch91 (approved):
Approved. The exact head closes the commit–reveal round-trip gap by rejecting only control/format/line-separator characters that the venue rewrites, while retaining visible non-ASCII verdicts. The tests cover the rewrite classes and multilingual verdicts, and the guard is narrowly scoped to `voteCommitment`; no wire or frame behavior changes.

bdunn77:
Independent validation of head `d227186f` (base `5cc4ab93`):

- On `origin/main`, `voteCommitment` accepts a newline verdict (`yes\n`) and returns a hash. The venue rewrites Cc/Cf/Zl/Zp to a space before store, so SPEC 8.3 cannot reopen that commitment.
- Head rejects those invisible classes at commit time and still accepts visible non-ASCII verdicts (`si`, `はい`, `нет`, `✓`).
- Head `tests/commitments.test.ts`: **11/11 pass**, including the rewrite-refusal and non-ASCII accept cases.

pucedoteth:
Approved twice on `d227186` and still merging cleanly. GitHub shows it as blocked, which from the outside looks like a required check that skipped rather than anything in the diff — is there something you'd like me to do, or is it just waiting on someone with merge rights?

Issue #78, opened 2026-09-04
A venue nonce above 2^53 is a valid record the read path cannot fold — 40% of the live `tclk-offers` export, and one row refuses the whole file


nbo01:
### Commit verified against

5cc4ab9

### Where

Something else

## What I observed

Technocore's signed lane accepts a nonce of 1–19 decimal digits (`400 nonce must be 1-19 digits` in technocore-chat `src/app.py`, `_signer`) and serializes it back as a **bare JSON number** in both `/export` rows and `?format=json` views. Signers on a nanosecond clock mint 19 digits. That is past `Number.MAX_SAFE_INTEGER` (2^53 − 1 ≈ 9.0e15), so `JSON.parse` rounds the value before tclk ever sees it.

`transcriptRecord` then takes the number branch, `Number.isSafeInteger` is false, and it throws `transcript message nonce must be decimal text`. Consequences, all reproduced against the live venue:

- `parseTranscriptExport("tclk-offers", <live export>)` throws at **line 3** — the whole export is refused (by design for a malformed row, but this row is not malformed).
- `examples/live-deal.mjs` therefore cannot finish step 5 (`exportRoom(OFFER_ROOM)`) against `technocore.chat`, and `examples/audit-export.mjs` cannot audit the board at all.
- `tclk_read_room` with `full: true` rejects for the same reason; without it, the window read sets every such record aside as `malformed` — 38 of the last 200 messages when I read it.
- Even without the throw, the signature could not verify: it covers `<room>|<exact digits>|<text>`, and the digits are gone.

Measured on the export taken 2026-09-03 ~23:50Z:

| | |
|---|---|
| rows | 8,953 |
| distinct signing DIDs | 999 |
| rows whose nonce > 2^53 − 1 | 3,619 (40%) |
| …from distinct DIDs | 291 |
| …that are `tclk1` frames | 3,589 |
| of those 3,619, Ed25519 verifies against the **exact** digits | 3,619 (all) |
| of those 3,619, verifies against the digits `JSON.parse` yields | 1,331 (only where the value happens to be a representable double) |

So these are venue-verified signed records from a third of the active signers, not hostile input, and the fold cannot see any of them. A deal with such a counterparty folds to `accepted` forever on this side.

## What the code promises instead

[...]

PR #81, opened 2026-09-04
fix(mcp): a string nonce must be canonical decimal, as a fold requires

Hyeongpil:
## What

`tclk_post_frame`'s string nonce must now be canonical decimal — no leading zero — at the
schema and in the shared handler. A caller passing `0000001730000000001` is refused before any
network call. `0` is still valid, the 19-digit bound is unchanged, and a numeric nonce is
unaffected.

## Why

The caller-signed tier sends a string nonce to the venue verbatim (`postSigned` does
`String(post.nonce)`), but the pattern added in #55, `^[0-9]{1,19}$`, admits a leading zero.
`verifyTranscriptRecord` accepts `^(?:0|[1-9][0-9]*)$` and nothing else, so the tool wrote a
record this same package refuses to fold — either the venue echoes the padded nonce back and
the read path calls it "not canonical decimal", or it normalizes to the bare number and the
signature over `room|0000001730000000001|text` stops verifying. Neither is recoverable: the
frame is on the board. Same shape as #59, one layer up — a write path looser than the read path
that consumes it.

Found while auditing our agent's live `tclk-offers` traffic (2026-09-04), where 19-digit string
nonces are routine; the reproduction is the added test, which fails on `main`. Leaves #78 (a
numeric nonce above 2^53 on the read path) alone. I treated the library's read path as correct.

## Checks

- [x] Docs that would now be wrong are updated: none — no doc stated the old pattern
- [x] New surface on the money path: nothing new; this narrows an input.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `tclk_post_frame`'s string nonce must now be canonical decimal — no leading zero — the same
+  spelling `verifyTranscriptRecord` accepts. The 1-19 digit pattern admitted `0000001730000000001`,
+  and the caller-signed tier sends a string nonce to the venue verbatim, so the tool wrote a record
+  this package then refuses to fold: either the venue echoes the padded nonce back and the read path
+  calls it "not canonical decimal", or it normalizes to the bare number and the signature over
+  `room|0000001730000000001|text` no longer verifies. A posted frame cannot be taken back, so the
+  write path is now no looser than the read path. `0` remains valid; a numeric nonce is unaffected.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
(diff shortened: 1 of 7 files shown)


luch91 (approved):
I audited the exact head fb8f3d1e18e7f0d8490f0964a9ac9cd5873d5fd4 against main. The canonical nonce rule is applied consistently in the MCP schema, shared handler, generated worker manifest, and tests: zero remains valid, safe numeric nonces remain supported, and string nonces are sent verbatim only when the fold can accept their spelling. This complements rather than duplicates PR #82's read-path precision fix. I found no source-level defect. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite.

luch91 (approved):
Approved. The exact head makes MCP string-nonce validation consistent with `verifyTranscriptRecord`: canonical decimal text only, no leading zero, with `0` retained and safe numeric nonces unchanged. The shared handler, server schema, generated Worker manifest, and tests agree; the regression confirms a rejected padded nonce makes no network call. GitHub build-and-test passed for this commit.

bdunn77:
Independent validation of head `fb8f3d1e` (base `5cc4ab93`):

- On `origin/main`, `tclk_post_frame` admits `/^[0-9]{1,19}$/`, so `007` and `0000001730000000001` pass. The caller-signed tier sends that spelling verbatim; `verifyTranscriptRecord` then refuses the padded form or the signature over it.
- Head uses `CANONICAL_NONCE = /^(?:0|[1-9][0-9]{0,18})$/` in both the handler and the MCP schema. `0` remains valid; a numeric nonce is unchanged.
- Head MCP tests: **19/19 pass** (`transport.test.ts` + `server.test.ts`).

PR #82, opened 2026-09-04
fix(transcript): preserve exact nonces above 2^53 in export and room parsing (#78)

### LINKED ISSUE #78: A venue nonce above 2^53 is a valid record the read path cannot fold — 40% of the live `tclk-offers` export, and one row refuses the whole file

Aphelios01-sdk:
## What

1. Adds `parseRecordJson` in `src/transcript.ts` (and exports it in `src/index.ts`) which preserves exact decimal digit string tokens for bare numeric nonces above `Number.MAX_SAFE_INTEGER` (2^53 - 1, up to 19 digits) using ES2023 `JSON.parse` reviver `context.source` with a safe token-boundary fallback for environments without context.
2. Updates `transcriptRecord` to support `bigint` nonces alongside string and safe integers.
3. Updates `parseTranscriptExport` to use `parseRecordJson` so full JSONL exports containing nanosecond-clock nonces parse without precision loss or throws.
4. Updates MCP `readRoom` in `mcp/src/technocore.ts` to parse GET response bodies with `parseRecordJson`, ensuring window reads preserve large nonces instead of marking them `malformed`.
5. Adds regression tests in `tests/transcript.test.ts` and `mcp/tests/transport.test.ts`.
6. Adds `CHANGELOG.md` entry.

Closes #78.

## Why

Technocore serializes nonces as bare JSON numbers. Signers on nanosecond clocks emit up to 19 decimal digits (exceeding `Number.MAX_SAFE_INTEGER`). Standard `JSON.parse` in JavaScript rounded these numbers, causing `Number.isSafeInteger(message.nonce)` in `transcriptRecord` to fail with `transcript message nonce must be decimal text`. On the live venue, this caused `parseTranscriptExport` to fail on valid board rows (around 40% of the live `tclk-offers` board export) and caused `readRoom` window reads to mark those records `malformed`.

## Checks

- [x] `tsc -p tsconfig.json`
- [x] `tsc -p mcp/tsconfig.json`
- [x] `pnpm test` (105/105 passed in root)
- [x] `pnpm --filter @flop-labs/tclk-mcp test` (41/41 passed in MCP)
- [x] Worker test suite (33/33 passed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,11 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- Preserve exact unrounded digit strings for bare numeric nonces above Number.MAX_SAFE_INTEGER
+  (2^53 - 1, up to 19 digits) in `parseTranscriptExport`, `parseRecordJson`, and `readRoom`.
+  Previously, standard `JSON.parse` rounded nanosecond-clock nonces and caused
+  `transcriptRecord` to reject valid rows as non-safe integers, refusing full-board exports
+  on the live venue (#78).
 - `tclk_read_room`'s bounded window read no longer fails the whole call when the venue
   returns one message with a malformed envelope. A room is world-writable, so a single
   hostile line — a `nonce` that is not decimal text, a non-string `from` — used to throw
--- a/mcp/src/technocore.ts
+++ b/mcp/src/technocore.ts
@@ -7,7 +7,7 @@
 // Fail-closed on the wire: a non-2xx answer throws with the status and the venue's own
 // first line (its 400s name the offending field), never a silently empty result.
 
-import { parseTranscriptExport, type TranscriptRecord } from "@flop-labs/tclk";
+import { parseRecordJson, parseTranscriptExport, type TranscriptRecord } from "@flop-labs/tclk";
 
 const NAME_RE = /^[a-z0-9][a-z0-9_-]{0,47}$/;
 
@@ -99,7 +99,8 @@ export function createClient(opts: { baseUrl?: string; fetch?: FetchLike } = {})
       if (!response.ok) await fail(response, `GET /r/${room}`);
       let view: unknown;
       try {
-        view = await response.json();
+        const text = await response.text();
+        view = parseRecordJson(text);
       } catch {
         throw new Error(`tclk-mcp: GET /r/${room} did not return JSON`);
       }
--- a/mcp/tests/transport.test.ts
+++ b/mcp/tests/transport.test.ts
@@ -264,6 +264,25 @@ describe("tclk_read_room", () => {
     expect(result.malformed).toEqual([{ seq: null, reason: expect.stringMatching(/seq/) }]);
   });
 
+  it("preserves nonces above 2^53 in window reads without marking them malformed (#78)", async () => {
+    const rawBody = JSON.stringify({
+      room: "lobby",
+      count: 1,
+      last_seq: 1,
+      messages: [],
+    }).replace(
+      '"messages":[]',
+      '"messages":[{"seq":1,"ts":"2026-01-01T00:00:00Z","from":"did:key:z6Mkaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa","text":"gm","nonce":1730000000000000001,"sig":"AQ"}]',
+    );
+
+    const { fetchLike } = fakeFetch([{ body: rawBody }]);
+    const h = createHandlers({ env: {}, fetch: fetchLike });
+    const result = await h.tclk_read_room({ room: "lobby" });
+    expect(result.malformed).toHaveLength(0);
+    expect(result.records).toHaveLength(1);
+    expect(result.records[0].nonce).toBe("1730000000000000001");
+  });
+
   it("reads and strictly parses the full JSONL export", async () => {
     const line = offerLine();
     const nonce = 17;
--- a/src/index.ts
+++ b/src/index.ts
@@ -35,7 +35,7 @@ export { TCLK_TERMINAL_STATUSES, openContract, applyFrame } from "./machine.js";
 export type { TclkStatus, ContractState, StepResult } from "./machine.js";
 
 export {
-  verifyTranscriptRecord, transcriptRecord, parseTranscriptExport,
+  verifyTranscriptRecord, transcriptRecord, parseTranscriptExport, parseRecordJson,
   findContractHandshake, foldTranscript,
 } from "./transcript.js";
 export type {
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -110,6 +110,33 @@ export function verifyTranscriptRecord(record: TranscriptRecord): TranscriptReco
   return { ok: true };
 }
 
+let supportsReviverContext = false;
+try {
+  JSON.parse('{"a":1}', ((_k: string, v: unknown, ctx?: { source?: string }) => {
+    if (ctx && typeof ctx.source === "string") supportsReviverContext = true;
+    return v;
+  }) as any);
+} catch {
+  // ignore
+}
+
+/**
+ * Parse a technocore record line or view payload to JSON while preserving exact
+ * decimal digit strings for numeric nonces above Number.MAX_SAFE_INTEGER (up to 19 digits).
+ */
+export function parseRecordJson(text: string): unknown {
+  if (supportsReviverContext) {
+    return JSON.parse(text, ((key: string, value: unknown, context?: { source?: string }) => {
+      if (key === "nonce" && typeof value === "number" && context && typeof context.source === "string") {
+        return context.source.trim();
+      }
+      return value;
+    }) as any);
+  }
+  const safe = text.replace(/(^|[{,]\s*)"nonce"\s*:\s*([0-9]{1,19})\s*([,}])/g, '$1"nonce":"$2"$3');
+  return JSON.parse(safe);
+}
+
 function object(value: unknown, where: string): Record<string, unknown> {
   if (value === null || typeof value !== "object" || Array.isArray(value)) {
     throw new Error(`tclk: ${where} is not a JSON object`);
@@ -139,6 +166,8 @@ export function transcriptRecord(room: string, value: unknown): TranscriptRecord
   if (typeof message.nonce === "string") nonce = message.nonce;
   else if (typeof message.nonce === "number" && Number.isSafeInteger(message.nonce)) {
     nonce = String(message.nonce);
+  } else if (typeof message.nonce === "bigint") {
+    nonce = String(message.nonce);
   } else if (message.nonce !== undefined && message.nonce !== null) {
     throw new Error("tclk: transcript message nonce must be decimal text");
   }
@@ -167,7 +196,7 @@ export function parseTranscriptExport(room: string, jsonl: string): TranscriptRe
     if (line.trim() === "") return;
     let value: unknown;
     try {
-      value = JSON.parse(line);
+      value = parseRecordJson(line);
     } catch {
       throw new Error(`tclk: transcript export line ${index + 1} is not JSON`);
     }
--- a/tests/transcript.test.ts
+++ b/tests/transcript.test.ts
@@ -13,6 +13,8 @@ import {
   makeAccept,
   makeOffer,
   parseTranscriptExport,
+  parseRecordJson,
+  verifyTranscriptRecord,
   type TranscriptRecord,
 } from "../src/index.js";
 
@@ -215,6 +217,19 @@ describe("trusted transcript records", () => {
     });
     expect(() => parseTranscriptExport(BOARD, withoutTimezone)).toThrow(/timezone-qualified/);
     expect(() => parseTranscriptExport(BOARD, localeTimestamp)).toThrow(/timezone-qualified/);
+  });
+
+  it("preserves exact nonces above 2^53 in export rows (#78)", () => {
+    const { offer } = deal();
+    const line = encodeFrame(offer);
+    const largeNonce = "1730000000000000001";
+    const sig = payer.sign(`${BOARD}|${largeNonce}|${line}`);
+    const rawExportLine = `{"seq":1,"ts":"2026-09-03T23:50:00Z","from":"${payer.did}","nonce":1730000000000000001,"sig":"${sig}","text":${JSON.stringify(line)}}`;
+
+    const records = parseTranscriptExport(BOARD, `${rawExportLine}\n`);
+    expect(records).toHaveLength(1);
+    expect(records[0].nonce).toBe(largeNonce);
+    expect(verifyTranscriptRecord(records[0]).ok).toBe(true);
   });
 
   it("never synthesizes offer-before-accept order while selecting a board handshake", () => {


thangvmt:
Ran this against a live full-board export rather than a synthetic row, because the failure it
fixes is one I hit from the outside: `examples/audit-export.mjs` cannot read the real
`tclk-offers` export on `main` at all.

Corpus: `GET /r/tclk-offers/export`, fetched 2026-09-04T07:48Z. 16,104 records, 9.56 MiB, seq
26,122 to 42,225, covering 2026-09-03T23:47:00Z to 2026-09-04T07:43:43Z. Whole file, no
filtering.

## Before

`main` @ `5cc4ab9`:

```
parseTranscriptExport(OFFER_ROOM, file)
→ tclk: transcript export line 7: transcript message nonce must be decimal text
```

Line 7 carries `"nonce": 1788479229609853000`. That is not an edge case in this corpus:

```
nonce is a safe integer      10,605
nonce above 2^53              5,499     34.1%
nonce as string, float, absent    0
```

A third of the live board is unreadable by the shipped audit path, and it fails on the seventh
row, so nothing downstream of it runs.

## After

`pr/82` @ `306082b`, built with the CI trio:

```
parseTranscriptExport → 16,104 records, no throw
nonces longer than 16 digits          5,499
of those, digits preserved exactly    5,499
```

The digit-for-digit comparison is against the raw `"nonce":<digits>` substring in the file, not
against a re-serialization, so it tests what the fix claims.

## The check that actually settles it

Digit preservation is not really provable by string comparison, because the question is whether
the recovered nonce is the one the sender signed. `room|nonce|line` is the signed message, so a
rounded nonce cannot verify. Running `verifyTranscriptRecord` over the whole corpus on this
branch:

```
records            16,104
signature valid    16,104
unsigned                0
signature failed        0     (18.8s)

of the 5,499 above 2^53: 5,499 valid, 0 failed
```

Every record in an eight-hour slice of the live board verifies, including every large nonce. On
`main` those 5,499 rows could not be reached at all.

One caveat on reproducing this: `tclk-offers` is a ring and drops history within hours, so this
exact file is not re-fetchable. The shape is: any full export taken now will contain a large
fraction of nanosecond-clock nonces and will fail on `main` within the first ten rows.

Aphelios01-sdk:
@thangvmt Thank you for running this comprehensive verification against the live production board!

Testing against the full 16,104-record export (with 5,499 nonces above $2^{53}$) and confirming that all 16,104 Ed25519 signatures verify cleanly with zero failures confirms that the recovered nonces match the exact signed bytes on the wire. Appreciate you sharing the benchmark and verification data!

zuli2021:
Independently verified PR #82 at `306082be1199dcdc6d2a8858828611f5fd3b7239` against a fresh full `tclk-offers` export from technocore.chat.

Corpus:
- 11,949 records total
- 11,947 signed/numeric-nonce records
- 930 numeric nonces above `Number.MAX_SAFE_INTEGER`

Results:

**Node v22.23.2**
- `JSON.parse` reviver `context.source`: available
- full export parsed successfully
- exact large nonces preserved: 930/930
- Ed25519 verification: 11,947/11,947 valid, 0 failures

**Node v20.20.2**
- `JSON.parse` reviver `context.source`: unavailable, so the fallback path was exercised
- full export parsed successfully
- exact large nonces preserved: 930/930
- Ed25519 verification: 11,947/11,947 valid, 0 failures

This gives live-corpus coverage for the Node 20 fallback path that the current Node 22 CI does not exercise.

Local evidence hashes:
- export: `13319dbd8f4d2ded10c3681ebed3b015a6245ffb0f46c56d034ea56fe901ca7c`
- Node 22 result: `a116166ad012e0e3f877c8f3bd69e4646872eb0c87f77649e9ac0d6d0ea9c5dc`
- Node 20 result: `7fef4d61ff980bfc6a92140a00591e08ab9b732adaaafa6fbdc778c1b25978d6`
- summary: `938327a674190eb8c653296658f791a6c00ee8dd9ab95776606fc846ab5a387d`

Aphelios01-sdk:
@zuli2021 Thank you for the independent verification against the live production export! Testing across both Node v22 (native JSON.parse context.source) and Node v20 (token fallback path) on the 11,949-record corpus provides valuable live evidence for both branches of the parser. Appreciate you sharing the results and evidence hashes!

luch91 (approved):
I audited the exact head 306082be1199dcdc6d2a8858828611f5fd3b7239 against main. The parser preserves the raw decimal nonce token through the context.source path and the bounded fallback, while normalized TranscriptRecord values remain canonical strings; readRoom uses the same parser. The focused regression coverage and the reported independent Node 20/22 live-export checks address both runtime branches, and GitHub build-and-test plus commit-policy checks pass. I found no remaining source-level defect and did not rerun the full suite locally.

0xjorgen:
Independent Codex-assisted validation of this fix on 2026-09-06. Checked PR head `306082be1199dcdc6d2a8858828611f5fd3b7239`; freshly fetched upstream `main` remains `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The required commands passed locally: `pnpm install --frozen-lockfile`, `pnpm -r --include-workspace-root build`, and `pnpm -r --include-workspace-root test` (105 library + 41 MCP + 33 Worker tests; no skips reported).

I additionally exercised fresh offline Ed25519 signatures at five nonce boundaries: `9007199254740991`, `9007199254740992`, `9007199254740993`, `1730000000000000001`, and `9223372036854775807`. All five retain exact nonce text and verify after `parseRecordJson` → `transcriptRecord`. Incrementing each nonce without resigning is rejected. The plain `JSON.parse` baseline fails for all four values above MAX_SAFE_INTEGER.

The same matrix passed after suppressing reviver context before importing the library, exercising the fallback path on Node 22.22.3. This simulates absence of context; it is not a test on an older runtime. No live messages or project signing keys were used. This is a bounded check of nonce preservation, not approval of the entire parser or protocol.

Reproducer: save as `review-82.mjs` in the checked-out PR after building, then run `node review-82.mjs` and `node review-82.mjs --fallback`:

```js
import assert from 'node:assert/strict';
import { generateKeyPairSync, sign } from 'node:crypto';

const nativeParse = JSON.parse;
if (process.argv.includes('--fallback')) {
  JSON.parse = (text, reviver) => nativeParse(text, reviver && function (key, value) {
    return reviver.call(this, key, value);
  });
}
const { parseRecordJson, transcriptRecord, verifyTranscriptRecord } = await import('./dist/index.js');
const { privateKey, publicKey } = generateKeyPairSync('ed25519');
const tagged = Buffer.concat([Buffer.from([0xed, 1]), publicKey.export({ format: 'der', type: 'spki' }).subarray(-32)]);
let number = BigInt('0x' + tagged.toString('hex'));
let encoded = '';
while (number) {
  encoded = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'[Number(number % 58n)] + encoded;
  number /= 58n;
}
const from = 'did:key:z' + encoded;
for (const nonce of ['9007199254740991', '9007199254740992', '9007199254740993', '1730000000000000001', '9223372036854775807']) {
  const text = 'offline nonce precision review';
  const sig = sign(null, Buffer.from(`lobby|${nonce}|${text}`), privateKey).toString('base64url');
  const raw = JSON.stringify({ seq: 1, ts: '2026-09-06T00:00:00Z', from, text, sig, nonce: 'REPLACE' }).replace('"REPLACE"', nonce);
  const row = transcriptRecord('lobby', parseRecordJson(raw));
  assert.equal(row.nonce, nonce);
  assert.equal(verifyTranscriptRecord(row).ok, true);
  assert.equal(verifyTranscriptRecord({ ...row, nonce: String(BigInt(nonce) + 1n) }).ok, false);
  let baseline;
  try { baseline = verifyTranscriptRecord(transcriptRecord('lobby', nativeParse(raw))).ok; }
  catch { baseline = false; }
  console.log(JSON.stringify({ nonce, baseline, patched: true, tamperedRejected: true }));
}
```

luch91 (approved):
Approved. The exact head preserves large bare numeric nonces without rounding in transcript exports and MCP room reads, while keeping the exact decimal text used by signature verification. It adds the parser as a public helper, supports bigint callers, and includes regression coverage for a 19-digit nonce and verified signature. GitHub build-and-test passed for this commit.

bdunn77:
Independent validation of head `306082be` (base `5cc4ab93`):

- On `origin/main`, `JSON.parse('{"nonce":9007199254740993}')` yields `9007199254740992` (not a safe integer). `parseTranscriptExport` / `readRoom` then refuse a valid 19-digit venue nonce.
- Head adds `parseRecordJson` (reviver `context.source` when available, otherwise a bounded nonce-digit rewrite) and uses it on export lines and MCP room JSON. `transcriptRecord` also accepts bigint nonces as decimal text.
- Head `tests/transcript.test.ts`: **8/8 pass**.

Aphelios01-sdk:
Pushed `702c8a9` in response to @sv's triage note on #78:

- Added the suggested normative rule to `SPEC.md` §2 ("Fold records, not detached lines"): transport `nonce` is decimal text, readers MUST recover its exact digits from the wire bytes and never from a parsed JSON number, and producers SHOULD emit it as a JSON string.
- Updated `CHANGELOG.md` to document the spec clarification.

All 105 tests pass locally, `generate-frame-fields --check` clean.

almondous:
I rechecked the current head `702c8a91203e73b9ce92e78c27d1cb022ea3ca90` and found one bounded parity gap in the old-runtime fallback. This is a follow-up to #82, not a competing fix.

The native `context.source` branch sees decoded member names and the complete numeric source token. The fallback regex currently does not:

- valid escaped member names such as `"no\u006ece"` and `"\u006eonce"` lose a large nonce;
- invalid JSON such as a leading-zero nonce or NBSP after the colon can be repaired into valid JSON by the rewrite;
- exponent, fraction, negative-zero and negative tokens do not retain the same source text as the native branch;
- a JSON-looking `"nonce":...` sequence inside signed text needs to remain untouched.

I put the proposed fallback change, changelog entry and permanent root/MCP/Worker regressions in one commit:

https://github.com/almondous/tclk/commit/6f63d08

The change validates the original JSON before rewriting, consumes string tokens whole, decodes candidate member names with `JSON.parse(key)`, and preserves the complete JSON number token. It does **not** broaden the accepted nonce grammar: `verifyTranscriptRecord` still rejects `1e3`, `1.0`, `-0`, etc. Last-key-wins behavior remains the underlying `JSON.parse` behavior.

### Red / green

Against the unmodified #82 head:

- built-library driver: native 14/14, forced fallback 5/14;
- new root regression: 16 failed / 30 passed;
- MCP fallback regression: 14 failed / 2 passed;
- Worker fallback regression: 6 failed / 0 passed.

With the follow-up:

- built-library driver: native 14/14 and forced fallback 14/14;
- Node 22.22.2 full suite: library 151, MCP 57, Worker 39 passed;
- Node 22.22.2 with fallback forced for the whole suite: 151 / 57 / 39 passed;
- actual Node 20.20.2, where reviver context is unavailable: 151 / 57 / 39 passed;
- leaving the tests in place and reverting only the fallback implementation gives 32 failures / 14 passes on Node 20; restoring it gives 46/46;
- MCP `tclk_read_room` window and full export both preserve and verify a synthetic signed record; line, nonce and room tampering fail verification;
- the dry-run Worker bundle succeeded, then the actual bundle passed 12 Miniflare/workerd cases across native/forced-fallback and window/export. Every outbound fetch was intercepted; no venue call or deployment was made;
- protocol generation check is clean; regenerating the Worker manifest produces no diff; golden vectors, SPEC, lockfile and generated frame fields are unchanged;
- a deterministic bounded 1,000-case JSON representation comparison matched native parsing exactly.

All signatures use a fixed synthetic test-only identity.

### Explicit performance tradeoff

The extra validation parse and whole-string token scan have a measurable cost on the fallback path. Node 22.22.2, real built module, seven-run median after warm-up:

| synthetic input | bytes | current fallback | follow-up fallback | ratio |
| --- | ---: | ---: | ---: | ---: |
| one export row | 767 | 0.00125 ms | 0.00407 ms | 3.26x |
| 200-row window | 153,614 | 0.21655 ms | 0.74138 ms | 3.42x |
| escaped-key 200-row window | 154,614 | 0.18698 ms | 0.75050 ms | 4.01x |
| large escaped string | 1,400,021 | 1.71853 ms | 6.73379 ms | 3.92x |

So this is roughly 3.3-4.0x slower than the current fallback in these synthetic cases, although the 200-row example remains under 1 ms. The native-context path is unchanged. The escaped-key baseline is not an equal-correctness comparison because it returns the wrong representation today.

I did not find an existing lossless-number primitive in the current code/dependencies that would remove this tradeoff without adding a dependency or a larger JSON parser. I also rechecked the current comments/reviews and searched fallback/nonce/escaped overlaps: #159 is the separate direct-runtime-type guard, #116 is the closed duplicate of #82, and #48/#66/#68 concern canonical frame escaping rather than the transport envelope.

I did not open another PR or write to the contributor's branch. The commit is intended to be reviewed/cherry-picked into this existing PR if the approach and fallback-only cost are acceptable.

PR #84, opened 2026-09-04
fix(transcript): verify contract binding in findContractHandshake

### LINKED ISSUE #83: findContractHandshake returns accepts that name a contract id the offer+accept do not derive
## Summary

`findContractHandshake(records, contract)` returns a signed accept naming the queried contract id as the authenticated handshake even when that contract id is not the one the offer + acceptance derive. An attacker can name an arbitrary contract id over a victim's public offer in `tclk-offers` and be reported as the authenticated pair, while `applyFrame` later rejects the same accept with `contract id mismatch`.

## Affected component

`src/transcript.ts` — `findContractHandshake` (used by `examples/live-deal.mjs` and `examples/audit-export.mjs` before folding, and by any downstream integrator selecting a handshake).

## Root cause

The helper checks transport authentication (valid Ed25519 signature, `frame.from === record.sender`, room is `tclk-offers`, accept follows its referenced offer in append order) but never recomputes the contract binding from SPEC §3.2. The contract id commits to the full offer plus the acceptance core (`from`, `ref`, `statement`, `paymentKey`, `nonce`); both sides must recompute it and reject on mismatch. The finder skips that recomputation entirely.

SPEC §3.2: the contract id binds the full offer and the acceptance, so either side tampering with any term yields a different contract, and a mismatch rejects the frame. A signature alone is not an agreement.

## Attack scenario

1. Victim posts an offer in the public, world-readable `tclk-offers` room (normal rendezvous).
2. Attacker posts an accept referencing the victim's offer id (`ref`), with their own statement/nonce, but with `contract` set to an arbitrary value (e.g. `0xaa…aa`) rather than the derived id.
3. Any consumer calling `findContractHandshake(board, "0xaa…aa")` is handed `{ offer: victim, accept: attacker }` as the authenticated pair.

A self-accept variant works the same way (acceptor equals offerer), which `applyFrame` rejects with `cannot accept own offer` but the finder returns.

## Prerequisites

Read access to `tclk-offers` (public by design) and the ability to post one signed message there (any `did:key`).

## Exact reproduction steps

```js
import { makeOffer, encodeFrame, findContractHandshake, contractId, openContract, applyFrame } from "@flop-labs/tclk";
// victim offer O (public), attacker accept with fake contract 0xaa…aa, ref = O.id
const hs = findContractHandshake([offerRec, fakeAcceptRec], "0x" + "aa".repeat(32));
// BEFORE fix: returns { offer, accept }. AFTER fix: null.
const expected = contractId(offer, { from, ref, statement, nonce });
// expected (0x0695…) !== queried (0xaaaa…), and:
applyFrame(openContract(offer), fakeAccept, now);
// → { ok: false, reason: "contract id mismatch" }
```

Verified locally: finder returned the spoofed pair while `applyFrame` rejected it.

## Observed result

`findContractHandshake` returns a handshake for a contract id the inputs do not derive.

## Expected secure result

No handshake: `null` (or the missing-offer throw when the referenced offer is absent). Only an accept whose recomputed contract id equals the queried id counts.

## Security / user impact

In-repo callers fold after finding, so `foldTranscript` catches it — but the function's contract says "authenticated offer/accept pair", and downstream integrators will reasonably trust it alone. That enables handshake spoofing (a fabricated agreement over a victim's offer) and self-accept laundering. No funds move on frames alone (the rail enforces independently), so this is a trust-boundary/authentication bug, not direct theft.

`authenticatedFrame` verifies the transport signature and sender binding, and the loop checks ordering and `frame.contract === query` — none of which covers the contract derivation. `applyFrame` does enforce it (`machine.ts`), but only runs if the caller folds after finding.

## Potential escalation

A spoofed handshake fed to accounting/reputation layers (receipts consume terminal states) or to a deal-room derivation (`dealRoom(fakeContract)` yields an attacker-chosen room) without a fold in between.

## Relevant code locations

- `src/transcript.ts` — `findContractHandshake` (missing `contractId` recompute; cf. `src/machine.ts` accept guard which enforces it)
- SPEC §3.2 (contract derivation), `src/frames.ts` `contractId`

## Recommended remediation

Recompute `contractId(offerFrame, acceptCore)` inside the finder and require equality with the queried id, plus the contextual checks `applyFrame` enforces (acceptor differs from offerer, statement fits the lock kind, point locks carry `paymentKey`). Skip non-matching accepts rather than returning them. No wire-format change; honest handshakes are unaffected. Code follows SPEC §3.2 (spec already requires recompute-and-reject).

## Deduplication note

Checked open issues (#78, #75, #61, #57, #49, #48, #41, #36, #26, #12, #5, #4, #3, #2): none cover handshake contract-id verification. #5 discusses `cancel`'s contract field while proposed; this is the accept-side binding in the finder.

## Verification

- New regression test `refuses a signed accept that names a contract id the offer+accept do not derive` fails before the fix, passes after.
- `pnpm install --frozen-lockfile`, `pnpm -r --include-workspace-root build`, `pnpm -r --include-workspace-root test` all pass (105 lib + 40 mcp + 33 worker).

Jr-kenny:
Fixes #83.

`findContractHandshake` now recomputes the contract id from the offer and the acceptance core and only returns the pair when it equals the queried id (plus acceptor-d differs, statement fits the lock kind, point locks carry `paymentKey`). Previously any signed accept naming the queried contract that followed its referenced offer was returned, so an attacker could name an arbitrary contract id over a victim's public offer and be reported as the authenticated handshake while `applyFrame` later rejects it.

Callers that fold after finding (`examples/live-deal.mjs`, `examples/audit-export.mjs`) see no behavior change on honest transcripts; callers that trusted the finder alone stop accepting spoofed/self-accept handshakes (now `null`).

## Spec agreement

Code follows SPEC §3.2: both sides recompute the contract id and a mismatch rejects. The spec already requires this; the finder was the one not doing it. No wire-format change, no golden-vector change.

## Tests

- New regression test in `tests/transcript.test.ts` (`refuses a signed accept that names a contract id the offer+accept do not derive`) fails before the fix, passes after.
- `pnpm install --frozen-lockfile`, `pnpm -r --include-workspace-root build`, `pnpm -r --include-workspace-root test` pass locally (105 lib + 40 mcp + 33 worker). `check:protocol` clean, commit-message scan clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `findContractHandshake` now recomputes the contract id from the offer and acceptance
+  core before returning a pair. It previously returned any signed accept naming the
+  queried contract that followed its referenced offer, so an attacker could name an
+  arbitrary contract id over a victim's public offer and be reported as the
+  authenticated handshake while `applyFrame` later rejects it. Code follows SPEC §3.2;
+  the spec already requires both sides to recompute and reject on mismatch.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -8,7 +8,8 @@
 import { ed25519 } from "@noble/curves/ed25519.js";
 import { base58, base64urlnopad } from "@scure/base";
 
-import { decodeFrame, tryDecodeFrame } from "./frames.js";
+import { contractId, decodeFrame, isValidStatement, tryDecodeFrame } from "./frames.js";
+import type { OfferFrame } from "./frames.js";
 import { applyFrame, openContract, type ContractState } from "./machine.js";
 import { dealRoom, OFFER_ROOM } from "./technocore.js";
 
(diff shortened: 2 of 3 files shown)


luch91 (approved):
I audited the exact head 01df94a4e31b55085462ef3331a3adc9fdffdef2 against main. findContractHandshake now applies the same contract-binding predicates required by SPEC §3.2 before returning a pair: distinct parties, lock-compatible statement, required point paymentKey, and recomputed offer/accept contract id. The added signed-spoof regression test exercises the previously missing rejection. I found no source-level defect. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite.

luch91 (approved):
Approved. The exact head closes the trust-boundary gap in `findContractHandshake`: it recomputes the contract ID from the referenced offer and acceptance core, rejects self-accepts and lock-kind-invalid statements, and enforces point-lock payment-key presence before returning a handshake. The regression covers a signed accept naming an unrelated contract, and the behavior matches `applyFrame` and SPEC §3.2.

PR #86, opened 2026-09-04
fix(machine): compare receipt rails canonically like the lock path

### LINKED ISSUE #85: Lock accepts registry-alias rail spellings but receipt compares rails exactly, censoring canonical receipts
## Summary

`applyFrame` accepts a `lock` rail under any registry-alias spelling (normalization) but compares `receipt.rail` by exact string. A lock stored as `paperrail` rejects the honest canonical receipt `paper` (and vice versa), censoring the post-terminal acknowledgment that reputation/spend-accounting layers consume.

## Affected component

`src/machine.ts` — `applyFrame`, `receipt` case (rail comparison). Lock path (`tclk1OfferIncludesRail`) is the inconsistent counterpart.

## Root cause

Two comparison rules for one identifier: the lock path normalizes (`normalizeRailId` lowercases plus maps `paperrail`/`paper-rail`/`PaperRail` → `paper`), while the receipt path used `frame.rail !== state.rail`. `encodeFrame` enforces canonical on emission, but `decodeFrame`/`foldTranscript` accept legacy spellings for replay — so an alias lock reaches the state machine and poisons the stored spelling.

SPEC §4–§5: rail lists are sets; match means non-empty intersection **after normalization**, independent of spelling or order. One rail, one identity.

## Attack scenario

1. Offer lists `rails: ["paper"]`.
2. Payer hand-crafts a `lock` line with `rail: "paperrail"` (decoder-legal legacy spelling, encoder would reject it) and posts it. `applyFrame` accepts: `locked, rail = "paperrail"`.
3. Deal completes (`reveal` → `claimed`).
4. Payee (or payer) posts the honest canonical receipt `rail: "paper"`. `applyFrame` rejects: `receipt rail paper does not match contract rail paperrail`.
5. Only a receipt echoing the alias spelling passes — which a canonical-only builder will never emit.

No payer/payee privilege needed beyond posting one's own frames; the alias lock can also arrive via a replayed historical line.

## Prerequisites

A contract whose lock was stored under an alias spelling (hand-built or historical). Honest canonical-only flows are unaffected.

## Exact reproduction steps

```js
const offer = makeOffer({ ..., rails: ["paper"], ... });
const accept = makeAccept(offer, ...);
let st = applyFrame(openContract(offer), accept, T0).state;
st = applyFrame(st, { type: "lock", from: PAYER, contract, rail: "paperrail", ref: "escrow-42" }, T0+1).state;
// st.rail === "paperrail", ok: true
st = applyFrame(st, { type: "reveal", ... }, T0+2).state; // claimed
applyFrame(st, { type: "receipt", outcome: "claimed", rail: "paper", ref: "escrow-42" }, T0+3);
// BEFORE fix: { ok: false, reason: "receipt rail paper does not match contract rail paperrail" }
// AFTER fix: { ok: true }
```

Verified locally, including the `paper-rail` alias in both directions.

## Observed result

Canonical receipt rejected for an alias lock; alias receipt required.

## Expected secure result

Receipt rails compared the same way locks are: equal when they normalize to the same registered id, exact-only for unnormalizable legacy customs.

## Security / user impact

`receipt` is post-terminal and never moves funds, so no theft. Impact is accounting/DoS plus interop split: the reputation/spend-accounting input SPEC §3.5 names becomes censorable by the payer's spelling choice, and canonical-only implementations disagree with alias-tolerant ones about whether a valid receipt exists.

`validateEmissionRails` enforces canonical only inside `encodeFrame`, not inside `decodeFrame`/`applyFrame` (deliberately, for replay). The fold path therefore admits the alias, and the receipt guard was never updated to match.

## Potential escalation

None to funds; persistent disagreement about terminal acknowledgment across implementations watching the same room.

## Relevant code locations

- `src/machine.ts` — `tclk1OfferIncludesRail` (normalizing) vs `receipt` rail check (exact)
- `src/frames.ts` — `validateEmissionRails` (encode-only canonical enforcement)
- `src/rails.ts` — `normalizeRailId` + aliases; SPEC §4–§5

## Recommended remediation

Compare receipt rails with the lock path's rule: exact match wins, else both sides normalized and compared, unnormalizable ids never equal except exactly. No wire change; canonical-only flows byte-identical.

## Deduplication note

Checked open issues (#78, #75, #61, #57, #49, #48, #41, #36, #26, #12, #5, #4, #3, #2): rail-alias handling is discussed for builders/decoders generally, but no issue covers the lock-vs-receipt comparison split.

## Verification

- New test `accepts a canonical receipt rail for a lock stored under a registry alias` fails before, passes after (both `paperrail` and `paper-rail` directions).

Jr-kenny:
Fixes #85.

Receipt rail comparison now uses the lock path's canonical rule (`sameTclkRail`): exact match wins, otherwise both sides are normalized (`paperrail`/`paper-rail`/`PaperRail` → `paper`) and compared; unnormalizable legacy customs only ever equal their exact spelling. Previously a lock stored under an alias accepted by normalization rejected the honest canonical receipt by exact compare, censoring the post-terminal acknowledgment.

Canonical-only flows are byte-identical; only alias-locked contracts change (their canonical receipts now pass, as the lock already implied).

## Spec agreement

Code follows SPEC §4–§5: rails are sets compared after normalization. The receipt guard was the one not doing it. No wire-format change, no golden-vector change.

## Tests

- New test in `tests/tclk.test.ts` (`accepts a canonical receipt rail for a lock stored under a registry alias`) fails before the fix, passes after (both alias directions).
- `pnpm install --frozen-lockfile`, `pnpm -r --include-workspace-root build`, `pnpm -r --include-workspace-root test` pass locally (105 + 40 + 33). `check:protocol` clean, commit-message scan clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `applyFrame` now compares receipt rails canonically, matching the lock path. A lock
+  stored under a registry alias (e.g. `paperrail`) previously rejected the canonical
+  receipt (`paper`) while the lock itself had been accepted by normalization. Code
+  follows SPEC §4–§5; rails are sets compared after normalization.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/src/machine.ts
+++ b/src/machine.ts
@@ -94,6 +94,21 @@ function tclk1OfferIncludesRail(offered: readonly string[], selected: string): b
 }
 
 /**
+ * Rail equality for receipt acknowledgments. Mirrors the lock path's canonical
+ * comparison: two spellings that normalize to the same registered id (e.g.
+ * `paperrail` vs `paper`) name the same rail, while historical custom ids that
+ * cannot be normalized only ever equal their exact spelling.
+ */
+function sameTclkRail(left: string, right: string): boolean {
+  if (left === right) return true;
+  try {
+    return normalizeRailId(left) === normalizeRailId(right);
+  } catch {
+    return false;
+  }
+}
+
+/**
  * Apply one frame at wall-clock `nowMs`. Clock and structural validation run first
  * (bad input is rejected, not thrown on); then the transition guards.
  */
@@ -213,7 +228,7 @@ export function applyFrame(state: ContractState, frame: TclkFrame, nowMs: number
       if (frame.outcome !== state.status) {
         return reject(state, `receipt outcome ${frame.outcome} does not match ${state.status}`);
       }
-      if (frame.rail !== undefined && state.rail !== undefined && frame.rail !== state.rail) {
+      if (frame.rail !== undefined && state.rail !== undefined && !sameTclkRail(frame.rail, state.rail)) {
         return reject(state, `receipt rail ${frame.rail} does not match contract rail ${state.rail}`);
       }
       if (frame.ref !== undefined && state.railRef !== undefined && frame.ref !== state.railRef) {
--- a/tests/tclk.test.ts
+++ b/tests/tclk.test.ts
@@ -280,6 +280,34 @@ describe("tclk state machine", () => {
     expect(contradictoryReceipt.state).toBe(claimed.state);
   });
 
+  it("accepts a canonical receipt rail for a lock stored under a registry alias", () => {
+    const offer = baseOffer({ rails: ["paper"], nonce: "aabbccddeeff0011" });
+    const { lock, state } = accepted(offer);
+    const locked = applyFrame(state, {
+      type: "lock", from: PAYER_DID, contract: state.contract!, rail: "paperrail", ref: "escrow-42",
+    }, T0 + 1);
+    expect(locked.ok).toBe(true);
+    expect(locked.state.rail).toBe("paperrail");
+
+    const claimed = applyFrame(locked.state, {
+      type: "reveal", from: PAYEE_DID, contract: state.contract!, ref: "escrow-42",
+      secret: lock.preimage,
+    }, T0 + 2);
+    expect(claimed.ok).toBe(true);
+
+    const canonicalReceipt = applyFrame(claimed.state, {
+      type: "receipt", from: PAYER_DID, contract: state.contract!, outcome: "claimed",
+      rail: "paper", ref: "escrow-42",
+    }, T0 + 3);
+    expect(canonicalReceipt.ok).toBe(true);
+
+    const aliasReceipt = applyFrame(claimed.state, {
+      type: "receipt", from: PAYER_DID, contract: state.contract!, outcome: "claimed",
+      rail: "paper-rail", ref: "escrow-42",
+    }, T0 + 3);
+    expect(aliasReceipt.ok).toBe(true);
+  });
+
   it.each([
     ["NaN", Number.NaN],
     ["Infinity", Number.POSITIVE_INFINITY],


stupeterwilliams-ui (commented):
This is the one @sv is keeping out of the two that fix #85, and it is the right choice — the
`sameTclkRail` fallback is what stops a receipt against a historical custom rail turning into a
thrown exception out of `applyFrame`, which `foldTranscript` does not catch.

**That fallback is currently the only untested line in the PR.** Your alias test covers
`paperrail`/`paper-rail`, which is the bug. Nothing covers the `catch`. I removed the exact-match
short circuit and the `try`/`catch` from `sameTclkRail` and **all 39 of your other tests still
pass** — so the next person who sees a two-branch helper wrapping a one-line comparison and
"simplifies" it gets #102's behaviour back with a green suite.

Here is a test that fails without the fallback and passes with it. Yours to take, alter or
ignore — no PR from me, it belongs on this branch.

```ts
  it("a historical custom rail id rejects cleanly rather than throwing out of applyFrame", () => {
    // The offer is DECODED, not built: `makeOffer` requires canonical rails, while
    // `validateFrame` admits a pre-registry spelling on the decode path (LEGACY_RAIL), which
    // is how such a frame reaches a fold — off the wire. `lock` then stores `frame.rail`
    // verbatim, so `state.rail` need not be canonical.
    //
    // This pins the fallback in `sameTclkRail`. Calling `normalizeRailId` on `state.rail`
    // unguarded throws, and `foldTranscript` calls `applyFrame` with no try/catch
    // (src/transcript.ts:308) — so the throw ends the fold of an entire room instead of
    // marking one record bad, which is worse than the mismatch being fixed. Without the
    // fallback every other test here still passes.
    const CUSTOM = "acme-custom-rail";
    const core = {
      amount: "1000000", asset: "FLOP", claimByMs: CLAIM_BY, expiresMs: EXPIRES,
      from: PAYER_DID, lock: "hash", nonce: "0011223344556677",
      rails: [CUSTOM], refundAfterMs: REFUND_AFTER, role: "payer", type: "offer",
    } as const;
    const offer = decodeFrame(
      TCLK_PREFIX + JSON.stringify({ ...core, id: offerId(core as never) }),
    ) as OfferFrame;
    expect(offer.rails).toEqual([CUSTOM]);

    const { lock, state } = accepted(offer);
    const locked = applyFrame(state, {
      type: "lock", from: PAYER_DID, contract: state.contract!, rail: CUSTOM, ref: "escrow-9",
    }, T0 + 1);
    expect(locked.ok).toBe(true);
    expect(locked.state.rail).toBe(CUSTOM);

    const claimed = applyFrame(locked.state, {
      type: "reveal", from: PAYEE_DID, contract: state.contract!, ref: "escrow-9",
      secret: lock.preimage,
    }, T0 + 2);
    expect(claimed.ok).toBe(true);

    // Same spelling: equal by the exact-match short circuit, never reaching the registry.
    const same = applyFrame(claimed.state, {
      type: "receipt", from: PAYER_DID, contract: state.contract!, outcome: "claimed",
      rail: CUSTOM, ref: "escrow-9",
    }, T0 + 3);
    expect(same.ok).toBe(true);

    // A different unregistered spelling: a clean rejection, and above all not a throw.
    const other = applyFrame(claimed.state, {
      type: "receipt", from: PAYER_DID, contract: state.contract!, outcome: "claimed",
      rail: "acme-other-rail", ref: "escrow-9",
    }, T0 + 3);
    expect(other.ok).toBe(false);
    expect(other.reason).toContain("does not match");
  });

```

It also needs `offerId` in the import block at the top of the file.

**One detail worth the comment it carries:** the offer is *decoded* rather than built. `makeOffer`
calls `normalizeRailIds` and refuses a custom rail outright, so the state cannot be reached
through construction — only off the wire, where `validateFrame` admits it via `LEGACY_RAIL`. That
is exactly how a fold receives frames, and it is why the case is reachable in production rather
than only in a hand-built fixture. I had to find that out the hard way: my first version of this
test used `baseOffer({ rails: [CUSTOM] })` and failed in setup, which for a few minutes looked
like your branch was broken.

Verified on `557f5a2`: 40 passing in this file, 106 across the suite, and with the fallback
stripped the new test alone fails with `tclk: unknown rail id: acme-custom-rail`.

Jr-kenny:
Added your regression test as-is (plus the `offerId` import) — verified it fails without the fallback with the thrown `unknown rail id` and passes with it. 40 in file, 106 suite. Thanks for the decode-not-build detail, that would have bitten.

luch91 (approved):
I audited the exact head 6caa5c6ed385c223482aa955408116451788af63 against main. Receipt comparison now mirrors the lock path: registered aliases normalize to the same rail, exact historical custom ids remain comparable, and differing unnormalizable custom ids reject without throwing. The added tests cover both alias directions and the historical custom-rail fallback requested in review. I found no source-level defect. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite.

luch91 (approved):
Approved. The exact head makes receipt rail comparison use the same canonical normalization as lock admission, while retaining exact-match handling for historical custom IDs and fail-closed behavior for unrelated spellings. The added tests cover alias acceptance and custom-rail non-throwing rejection, and the change matches the SPEC’s normalized rail-set rule.

PR #88, opened 2026-09-04
fix(paper-rail): harden case handling and fail-closed reads

### LINKED ISSUE #87: PaperRail stores what verifySecret accepts but decodePaperRecord rejects: uppercase claims become unauditable, malformed ids throw
## Summary

`PaperRail` writes records in canonical lowercase but `verifySecret` accepts uppercase hex, so an uppercase claim succeeds while writing a record no reader can parse back — a successful claim becomes unauditable. Separately, `read`/`verifyLock` threw on malformed ids instead of failing closed, and `lock` accepted non-canonical statements it could never make readable.

## Affected component

`src/paper-rail.ts` — `PaperRail.lock` / `claim` / `read` / `verifyLock` (the rehearsal rail; explicitly holds no value).

## Root cause

Three half-enforced invariants around one encoding rule (`decodePaperRecord` requires lowercase `0x[0-9a-f]{64}`):

1. `claim` passed the caller-supplied `secret` straight to the record after `verifySecret` approved it — but `verifySecret` lowercases before comparing, so uppercase passes verification yet is stored verbatim and fails `decodePaperRecord`.
2. `lock` stored `terms.statement` without `isValidStatement`, so an uppercase/hand-built statement produced an unreadable `locked` record (`verifyLock` → `false` forever).
3. `read` let `paperNote`'s throw escape (should be `null` per its own docstring: "null when absent or unparseable"), and `verifyLock` let it escape as a throw instead of the documented fail-closed `false`.

Fail closed everywhere; every read out of the world-writable note namespace is anonymous input that must never throw out of a polling loop. Plus the record format's own rule: emitted lines must round-trip through the parser.

## Attack scenario / prerequisites

No privilege needed beyond calling the rail (rehearsal surface):

```js
const { hash, preimage } = generateHashLock();
await paper.lock({ ..., statement: hash, ... }); // honest, lowercase
await paper.claim(contract, preimage.toUpperCase().replace("0X", "0x"));
// BEFORE fix: succeeds; raw line ends `... 0x6B14…FB6`; read() → null; decode → null.
// AFTER fix: succeeds; stored lowercase; read() → claimed.
await paper.verifyLock({ contract: "0xZZZ", ... }, "0xZZZ");
// BEFORE: throws "malformed contract id". AFTER: false.
await paper.lock({ ..., statement: UPPER, ... });
// BEFORE: succeeds, then unreadable. AFTER: throws "does not fit its lock kind".
```

A counterparty that claims with an uppercase spelling can later point at the unreadable note and deny the claim's contents; a malformed id can throw out of a reader that expected `false`/`null`.

## Observed result

Successful claim unauditable (`read` null); malformed ids throw; non-canonical locks writable-but-unreadable.

## Expected secure result

- `lock` refuses non-canonical statements (and non-positive `refundAfterMs`) rather than writing unparseable records.
- `claim` stores the canonical lowercase spelling it verified.
- `read` returns `null` and `verifyLock` returns `false` on malformed ids; transport failures still throw (outage, not a negative answer).

## Security / user impact

The paper rail settles nothing by design ("evidence of a rehearsal, never of a payment"), so no funds at risk. Impact is auditability/availability of the rehearsal path: the one honest job the rail claims ("a client written against this rail is written correctly") breaks when secrets change case, and readers must catch throws the interface says will not happen. Low severity, high correctness signal for anyone porting predicates to a value-bearing rail.

`decodePaperRecord` is strict (correct), but the writers were not: `verifySecret`'s case-insensitivity was treated as storage permission. `read`'s docstring promises `null`, the code threw.

## Potential escalation

None to funds on this rail. The pattern matters for real rails: store exactly what you verify, in the spelling you read back.

## Relevant code locations

- `src/paper-rail.ts` — `lock` (missing `isValidStatement`), `claim` (verbatim `secret`), `read`/`verifyLock` (throw vs null/false)
- `src/locks.ts` `verifySecret` (case-insensitive by design), `src/frames.ts` `isValidStatement` (lowercase canonical)

## Recommended remediation

Validate at write time, normalize the secret to lowercase on claim, return `null`/`false` on malformed ids. No record-format change; honest lowercase flows byte-identical. Code follows the existing fail-closed + canonical-encoding rules.

## Deduplication note

Checked open issues (#78, #75, #61, #57, #49, #48, #41, #36, #26, #12, #5, #4, #3, #2): none cover paper-rail case handling or fail-closed reads.

## Verification

- New test `stays fail-closed and readable on non-canonical case` fails before, passes after.

Jr-kenny:
Fixes #87.

- `lock` now refuses a statement that does not fit its lock kind (and a non-positive `refundAfterMs`) instead of writing a note line no reader can parse back. Honest terms come from `lockTerms()` over an accepted contract and already pass, so honest flows are unchanged.
- `claim` stores the canonical lowercase spelling of the secret it verified (`verifySecret` accepts uppercase by design). Previously an uppercase claim succeeded and then read back as `null`.
- `read` returns `null` and `verifyLock` returns `false` on malformed ids per their documented fail-closed contracts. A transport failure from the note store still throws — that is an outage, not a negative answer.

## Spec agreement

No record-format change; the parser (`decodePaperRecord`) is untouched and honest lowercase flows are byte-identical. Code follows the existing fail-closed + canonical-encoding rules the module already states.

## Tests

- New test in `tests/paper-rail.test.ts` (`stays fail-closed and readable on non-canonical case`) fails before the fix, passes after.
- `pnpm install --frozen-lockfile`, `pnpm -r --include-workspace-root build`, `pnpm -r --include-workspace-root test` pass locally (105 + 40 + 33). `check:protocol` clean, commit-message scan clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `PaperRail` now refuses a non-canonical lock statement, stores claims in canonical
+  lowercase, and reads malformed ids as absent. `verifySecret` accepts uppercase hex,
+  so an uppercase claim previously succeeded while writing a record no reader could
+  parse back; `verifyLock` threw on a malformed id instead of returning `false`.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/src/paper-rail.ts
+++ b/src/paper-rail.ts
@@ -114,6 +114,16 @@ export class PaperRail implements SettlementRail {
     if (this.clock() >= terms.refundAfterMs) {
       throw new Error("tclk: refusing to lock into an already-open refund window");
     }
+    // The record only stores lowercase canonical spellings (decodePaperRecord enforces
+    // them), so refuse a non-canonical statement here rather than writing a record no
+    // reader can parse back. Honest terms always come from lockTerms() over an accepted
+    // contract, where the statement already passed isValidStatement.
+    if (!isValidStatement(terms.lock, terms.statement)) {
+      throw new Error("tclk: paper rail lock statement does not fit its lock kind");
+    }
+    if (!Number.isSafeInteger(terms.refundAfterMs) || terms.refundAfterMs <= 0) {
+      throw new Error("tclk: paper rail lock refundAfterMs must be a positive unix-ms integer");
+    }
     const { ns, key } = paperNote(terms.contract);
     const record: PaperRecord = {
       status: "locked",
@@ -128,6 +138,9 @@ export class PaperRail implements SettlementRail {
 
   async verifyLock(terms: LockTerms, ref: string): Promise<boolean> {
     if (ref !== terms.contract) return false;
+    // read() returns null for malformed ids and unparseable lines, so a malformed
+    // input ends here as false. A transport failure from notes.get still throws,
+    // which is correct: that is an outage, not a negative answer.
     const record = await this.read(ref);
     return (
       record !== null &&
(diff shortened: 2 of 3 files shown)


luch91 (approved):
I audited the exact head c5235ccca8c143c7abe92f8c1d73469d69fb3d68 against main. PaperRail now validates canonical statements and positive safe deadlines before writing, normalizes a verified uppercase secret before encoding a claimed record, and treats malformed ids as absent on read/verification while preserving transport failures. The added regression test covers these boundaries. I found no source-level defect. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite.

bdunn77:
Independent validation of head `c5235ccc` (base `5cc4ab93`):

- On `origin/main`, `read("0xZZZ")` and `verifyLock(..., "0xZZZ")` throw `malformed contract id` instead of returning null/false. `claim` stores the caller-supplied secret; a probe with uppercase hex stored a secret that `decodePaperRecord` cannot parse (`/^0x[0-9a-f]{64}$/`), so `canonical` was false.
- Head catches a malformed id in `read` as null (verifyLock then false), refuses a non-canonical statement at lock, and stores `secret.toLowerCase()` so a successful claim stays readable.
- Head `tests/paper-rail.test.ts`: **12/12 pass**, including fail-closed non-canonical case.

PR #91, opened 2026-09-04
fix(schema): compare generated content, not checkout line endings

### LINKED ISSUE #90: Windows CRLF checkout fails generated protocol field freshness check without content drift
### Commit verified against

5cc4ab9

### Where

Build, packaging, CI

### Summary

### Smallest reproduction

```powershell
cd tclk
git rev-parse --short HEAD
```

```text
```

### LF control

- build: passes
- root tests: 104/104 pass
- MCP tests: 40/40 pass
- Worker tests: 33/33 pass

### Related prior observation

### Cause boundary

### Expected behavior

### Environment

- Windows
- Git `2.55.0.windows.3`
- Node `v24.19.0`

Orvynel:
Fixes #90.

## What breaks

`scripts/generate-frame-fields.mjs --check` compares the generated field contract and the SPEC table byte for byte, and nothing pins the checkout's line endings. On a clone made with `core.autocrlf=true` both files hold CRLF while the generator emits LF, so the freshness check fails on an unmodified `main`:

```
```

Measured at `5cc4ab9`: root suite 103/104, the one failure being `tests/schema-conformance.test.ts > has current generated decoder fields and SPEC documentation`; build clean; MCP 40/40; Worker 33/33.

Following the message's advice does not settle it. `pnpm generate:protocol` on that clone rewrites the generated block of `SPEC.md` with LF and leaves the rest of the file CRLF (CR 471 to 460). `git status` then reports both files modified with an **empty** diff, alongside `warning: LF will be replaced by CRLF the next time Git touches it` — and the next checkout restores CRLF and the failure with it. CI is `ubuntu-latest` only, so nothing upstream sees any of this.

## What I treated as correct

The schema and the committed artefacts. Their content is identical across both checkouts — #90's normalised hashes show that, and I reproduced it. The comparison is the part that was wrong: it asserted a property of the working tree that the schema does not own.

## The change

Two layers, one problem. The attribute alone would leave every existing Windows clone failing until it was renormalised; the fold alone would leave the repository's checkout unspecified.

- `--check` folds CRLF to LF on both sides before comparing. Writes are untouched and still emit LF.
- `.gitattributes` pins `* text=auto eol=lf`. Every tracked file here is text — 43 `.ts`, 10 `.md`, 7 `.json`, 5 `.yml`, 5 `.mjs`, 2 `.yaml`, one `.sh`, one `.mts`, one `.jsonc` — with no binaries and no `.bat`, `.cmd` or `.ps1`, so nothing in the tree wants CRLF.

## Regression test

#90 asks for a CRLF fixture, and the failure is otherwise invisible to Linux CI. The new case copies the script, schema, generated contract and SPEC into a temp directory with CRLF endings and runs `--check` against that root, so a Windows-only failure is reproducible on `ubuntu-latest`. It touches no tracked file, and its fixture is idempotent (`/\r?\n/g` to CRLF) so it behaves the same on a CRLF working tree.

Three rows, run on this branch:

| script under test | the new case |
|---|---|
| this branch | passes |
| `main`'s byte comparison | fails — `expected [Function check] to not throw` |
| `sameContent` forced to `true` | fails — `expected [Function check] to throw` |

The third row is what the test's second half is for: an unregenerated schema must still be caught, so the fold cannot be widened into a comparison that always agrees.

## Verification

- LF clone, the three CI commands: `install --frozen-lockfile` clean, both `tsc` builds clean, then `9 passed (9)` / `105 passed (105)`, `6/6` / `40/40`, `2/2` / `33/33`.
- The same three commands on Linux x86_64 with node `v22.22.1`, since the new case spawns a child process and writes to a temp directory and everything above was measured on Windows: identical `105` / `40` / `33`, and `scripts/no-agent-trailers.sh` passes on the commit.
- The CRLF clone that was failing, with `src/frame-fields.generated.ts` and `SPEC.md` still CRLF on disk and nothing renormalised: the same 105 / 40 / 33.
- A fresh `git -c core.autocrlf=true clone` of this branch checks out CR 0 across the tree, and `main`'s pre-fix script exits 0 there — the attribute alone is enough for a new clone.
- `scripts/no-agent-trailers.sh` passes on the commit message, with `core.hooksPath` set to `.githooks`.

No golden vectors, wire encoding, frame semantics or rail registry are touched.

Technocore identity: `did:key:z6MkiusXWVcKL5PWPXpJp1HTLxrriYsWPyvyTdvvsHTfYCDE`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Every tracked file here is text, and two of them are compared byte for byte against
+# freshly generated output by `scripts/generate-frame-fields.mjs --check`. Pin the checkout
+# so that comparison does not depend on a contributor's `core.autocrlf`.
+* text=auto eol=lf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ All notable changes to this project are documented here. Format follows
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
   boundary rather than silently rounded.
+- The generated-protocol drift check now compares content rather than line endings, and a
+  `.gitattributes` pins the checkout to LF, so a clone made with `core.autocrlf=true` no
+  longer fails `pnpm test` with `generated protocol fields are stale` when nothing has
+  drifted. Regenerating on such a clone previously rewrote `SPEC.md` with mixed line endings.
 
 ### Added
 
--- a/scripts/generate-frame-fields.mjs
+++ b/scripts/generate-frame-fields.mjs
@@ -56,8 +56,12 @@ if (nextSpec === spec && !spec.includes(table)) {
   throw new Error("SPEC.md has no generated frame-field markers");
 }
 
+// Line endings belong to the checkout, not to the schema. Compare with CRLF folded to LF so
+// a working tree that stores them that way does not read as protocol drift; writes stay LF.
+const sameContent = (a, b) => a.replaceAll("\r\n", "\n") === b.replaceAll("\r\n", "\n");
+
 if (process.argv.includes("--check")) {
-  if (readFileSync(generatedPath, "utf8") !== generated || spec !== nextSpec) {
+  if (!sameContent(readFileSync(generatedPath, "utf8"), generated) || !sameContent(spec, nextSpec)) {
     console.error("generated protocol fields are stale; run pnpm generate:protocol");
     process.exit(1);
   }
--- a/tests/schema-conformance.test.ts
+++ b/tests/schema-conformance.test.ts
@@ -3,7 +3,9 @@
 // generated decoder contract or the normative SPEC table from drifting away unnoticed.
 
 import { execFileSync } from "node:child_process";
-import { readFileSync } from "node:fs";
+import { copyFileSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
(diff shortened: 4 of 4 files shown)


ChozetsuBlack:
Independent check from a Windows checkout, since #90 is about CRLF and CI only runs on ubuntu.

Windows 11 (10.0.26200), Node v24.14.0, pnpm 11.25.0 (corepack), Git 2.53 with `core.autocrlf=true`.

- `main` at 5cc4ab9: `tests/schema-conformance.test.ts` fails through `scripts/generate-frame-fields.mjs --check` — root suite 103/104. `pnpm -r` stops there, so the `mcp/` suites do not run.
- This branch at 42c6f93, same checkout, same settings: root 105/105, mcp 40/40, mcp/worker 33/33. All three CI commands exit 0.

No content drift on my side either — `git diff` on the generated file is empty once line endings are normalized, which is what this change compares.

luch91 (approved):
I audited the exact head 42c6f93063270f5a86fcf350fc29b2b4a43a1573 against main. The check now compares normalized content while preserving LF writes, .gitattributes establishes LF for fresh checkouts, and the regression test proves both CRLF-only equivalence and genuine schema drift detection. The independent Windows verification confirms the original failure on main and passing root/MCP/worker suites on this head. GitHub exposes only the skipped optional [code]smith check; I did not independently rerun the full suite.

luch91 (approved):
Approved. The exact head separates checkout line endings from generated-content drift: `--check` folds CRLF only for comparison, writes remain LF, and `.gitattributes` pins future checkouts. The regression test verifies both CRLF-equivalent content and genuine schema drift. No wire or protocol semantics change, and the scope is distinct from PR #121’s generator-output freshness fix. GitHub’s available checks report success for this head.

Issue #41, opened 2026-09-03
SPEC question: make accepted-but-unlocked abandonment observable


sv:
## Spec question

Live-wire measurement found 663 accepted contracts but only 14 lock frames and one cancel.
Today an accepted contract can remain in `accepted` forever: disappearing is free and leaves no
signed evidence distinguishing "declined after accept" from "temporarily offline".

This is not a state-machine bug with an obvious local fix. SPEC §4 deliberately leaves value
timing to the settlement rail, while any coordination deadline added to `accept` becomes part of
the signed contract and its id. The protocol should choose the observable semantics before code
changes.

## Candidate A: bind `lockByMs` into `accept`

Add a coordination deadline after which a lock frame is no longer accepted.

- It gives both parties a deterministic upper bound and makes an abandoned acceptance observable
  from the transcript plus a clock.
- It needs a defined relationship to `offer.expiresMs`, `claimByMs`, and each rail's setup/finality
  latency. A deadline safe for `paper` may be unsafe for BTC.
- Adding a required accept field changes the bytes committed by `contractId`; that is likely a
  `tclk2 ` wire decision, not a compatible tclk/1 validator tweak.
- Clock-only expiry is not signed evidence that either party intentionally aborted.

## Candidate B: require an `abort` frame after the lock window lapses

Add an explicit terminal coordination frame for accepted-but-unlocked contracts.

- It produces attributable evidence for reputation and funnel accounting.
- A vanished party cannot be forced to sign, so "mandatory" needs an enforcement consequence;
  otherwise absence remains ambiguous.
- The protocol must say who may abort (payer, payee, or either), what timestamp makes it valid,
  and whether the other party can synthesize closure after timeout. Without a deadline this frame
  alone does not close the cliff.

## Decisions needed

1. Is the goal a deterministic timeout, attributable intent, or both (`lockByMs` plus `abort`)?
2. Who chooses the deadline, and what minimum rail-specific safety margin constrains it?
3. Does a lapse create a local terminal state automatically, or only authorize a signed abort?
4. Who may sign abort, and can one party close unilaterally after the deadline?
5. Is accepted-but-unlocked abandonment reputation evidence, or merely missing liveness?
6. Does the wire change wait for the payee-opened secret-custody decision in #12 so one version
   bump can settle both handshake questions?

Related: #12 covers the payee-opened secret-assignment seam and should remain the source of truth
for that handshake decision. #31 explains why board-only observation misses deal-room locks; this
[...]

PR #92, opened 2026-09-04
feat(machine): expose whether an accepted contract can still be funded (#41)

zkasuran:
## What

`fundingOutlook(state, nowMs)` — a derived reading of whether a contract can still be funded.
No new frame, no new field, no new status, no wire byte, no golden vector.

Answers #41.

## Why the deadline already exists

#41 offers two candidates and both need a wire change: bind a `lockByMs` into `accept`, or add an
`abort` frame. Neither is necessary, because `applyFrame` already refuses a lock past a deadline
both parties signed:

```ts
case "lock": {
  …
  if (nowMs >= state.offer.refundAfterMs) return reject(state, "refund window is already open");
```

`refundAfterMs` is an offer field committed by `contractId`, so the bound was agreed before either
party acted. SPEC §4's own diagram already prints it: `accepted ──lock(payer, rail ∈ offer.rails,
now < refundAfterMs)──▶ locked`.

What was missing is that nothing exposed it. #41's measurement — 663 accepted contracts, 14 lock
frames, one cancel — cannot say which of those acceptances could still be funded, so an
abandonment and a party being briefly offline read identically. That is an observability gap, not a
state-machine gap.

## What it reports

The four buckets #41's acceptance criteria ask a funnel to distinguish, plus the proposed side:

| status | clock | `code` | `actionable` |
|---|---|---|---|
| `accepted` | `now < refundAfterMs` | `open` | **true** — merely unobserved |
| `accepted` | `now >= refundAfterMs` | `lapsed` | false — the window shut, no lock arrived |
| `locked` | any | `funded` | false — already happened |
| `cancelled` | any | `cancelled` | false — explicitly closed |
| `claimed`/`refunded` | any | `settled` | false |
| `proposed` | `now < expiresMs` | `awaiting-accept` | true |
| `proposed` | `now >= expiresMs` | `offer-expired` | false |
| any | not a finite non-negative clock | `unreadable-clock` | false |

Two details worth naming. Nothing requires `expiresMs < refundAfterMs`, so an offer can still be
acceptable at a clock where no lock could follow — that is `lapsed`, a closed funding window, not
an expired offer, and the test covers it by accepting such an offer successfully at that clock.
And `msRemaining` is reported rather than only a verdict, because **no safety margin is applied**:
a margin is rail-specific (how long a rail takes to accept a lock, how much slack a payee wants
before `claimByMs`), so it belongs to the caller. `claimByMs` is deliberately not a boundary here —
no transition guard reads it, and §3.1 calls it the payee's *safe* claim deadline.

## Why it is derived and not signed

This is the part that answers #41's decisions 3 through 5, and it is a deliberate refusal rather
than an omission. Absence of funding is not something a vanished party can be made to attest, and
a clock passing is not evidence that anyone intended to abort. So there is no signed terminal state
for the absence of funding: nothing here adds a frame, a field, a status, or contract state, and a
lapse authorises nothing on its own. Explicit, attributable abandonment is an agreement-state
question, which is tclk/2's to answer.

## Checks

`origin/main` at `5cc4ab9`.

```
```

109 library (was 104) + 40 mcp + 33 worker = 182 passed, 0 failed.

**Both ways.** A pure addition has a trivial revert, so the meaningful proof is the drift
tripwire: the reading takes its boundary from the same comparison the guard uses, and one test
pins them together over six clocks across it. One mutation at a time:

| mutation | tests failing |
|---|---|
| the `lock` guard's `>=` becomes `>` | **2** of 43 — an existing boundary test, and the agreement test |
| the reading's `>=` becomes `>`, guard untouched | **2** of 43 — both mine |

The second row is the argument for that test. On `main` a wrong boundary inside a new derived
reading is invisible to the entire suite; nothing else compares the two.

Fork CI can sit at `action_required` with no jobs, so those numbers are from a local run rather
than a green check I watched.

## Overlap

- **#32** is the other open PR in `src/machine.ts`, `SPEC.md` and `tests/tclk.test.ts`. Its hunks
  are the `cancel` and `receipt` cases, a `ContractState` field, SPEC §3.5 and `tests/tclk.test.ts`
  around line 372. Mine are a new function after `applyFrame`, the end of SPEC §4, and a new
  `describe` at the end of the test file. `git merge-tree --write-tree` is clean in both
  directions.
- Seven of my own are open and all touch `CHANGELOG.md`; each is anchored after a different
  existing bullet, and every pairing merges clean, checked the same way. This one is anchored
  after the `heartbeat` entry.
- **#12** stays the source of truth for the payee-opened secret-assignment seam; nothing here
  touches condition authority.
- **What a hostile counterparty gets: nothing.** A pure function over a state a caller already
  holds. It reads `state.offer` and `nowMs`, returns a fresh object, mutates nothing (pinned by a
  test), performs no I/O, and is not reachable from any frame.

Which side I treated as correct: **the code**. The guard was already right; SPEC §4 did not draw
the consequence, so §4 gains a paragraph. Fits the boundary #58 records — improves auditability,
preserves every existing byte.

## Provenance

Signed with the same `did:key` as #59, #60 and #66–#74.

```
head       0c38102eae33de22ab90ec8a111384a5a9cbbed1
digest     8292fb2848716c99bb6e159582e00bdcfe8d51bd39ce605cbdebc0bfe320c218
signature  TqXlGyFxW-l_kfipeNURVzOU155zdsPDJutJa0vR0BwvtBjolwrgXTaLE3ft1tK2L1-HrVmIh9TWVHTa_6_0Cw
```

Rebuild the digest: `sha256` each file's bytes at `head`, one `<sha256>  <path>` line per file in
the order `CHANGELOG.md`, `SPEC.md`, `src/machine.ts`, `src/index.ts`, `tests/tclk.test.ts`, then `sha256` that text.

```
397fb1a727c62e0aac8a31a5d8760a412461e7c5394552d9769cf339b8f56570  CHANGELOG.md
42391a8bd76a74d7b6d09bf6580fb95786a87c02fcb2ab783c889164e193c364  SPEC.md
3f3e13fe5f81428c31c396c606c41ea992bb59c878810378f6e320edfaf9d311  src/machine.ts
d752e7615e4ef264e0af0e8515e033aa28bdba5f15563cba833c69ae7a107f42  src/index.ts
7f68ef63f151c9bf5bf3a8740be26e4b60101d7ce43a6aa17b09613484808d5e  tests/tclk.test.ts
```

Signed payload `tclk-pr|flop-labs/tclk|<head>|<digest>`. Verify with the public key decoded out of
the DID — no private key in the loop, and a tampered payload must fail. Key note:
<https://technocore.chat/kv/agent/f15ddb2552fee06f> (the documented `/kv/did/` namespace is at its
cap, flop-labs/technocore-chat#85).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,18 @@ All notable changes to this project are documented here. Format follows
   and the normative `SPEC.md` table are checked for drift in CI.
 - A signed `heartbeat` frame for non-authoritative liveness while a contract is accepted
   or locked, without abusing terminal receipts or changing contract state.
+- `fundingOutlook(state, nowMs)`, a derived reading of whether a contract can still be funded.
+  An accepted contract's status sits at `accepted` indefinitely when a payer walks away, so an
+  abandonment and a party being briefly offline look the same; funding itself does not, because
+  `applyFrame` already refuses a `lock` once `nowMs >= offer.refundAfterMs` and that field is
+  committed by `contractId` before either party signs. The bound was therefore already agreed and
+  simply unexposed. The reading separates an accepted contract whose window is still open from one
+  whose window shut with no lock, from one explicitly cancelled, from one already funded, which is
+  the distinction funnel accounting needs (#41). It is derived and never signed: absence of
+  funding is not something a vanished party can be made to attest, and no coordination deadline
+  field or abort frame is added. No safety margin is applied — a margin is rail-specific, so the
+  time remaining is reported and the caller applies its own. Nothing in the state machine, the
+  wire format or the golden vectors changes.
 - A hosted deployment of the MCP server at `https://tclk.technocore.chat/mcp`, streamable
   HTTP, no account and no key. It is the no-custody Worker build: it binds neither
   `TECHNOCORE_SIGNING_KEY` nor `TCLK_PAYMENT_KEY` and refuses to serve if either is present,
--- a/SPEC.md
+++ b/SPEC.md
@@ -249,6 +249,20 @@ distinct from two valid sets having no overlap. For a historical custom tclk/1 i
 normalized, replay uses the original exact-string membership rule and never treats it as equal
 to a registered rail.
 
+**Abandonment is already observable, and `refundAfterMs` is the deadline.** A contract's *status*
+stays `accepted` indefinitely when a payer walks away, which is what makes an abandonment look
+like a party being briefly offline. Funding, though, does not: the `lock` guard above refuses a
+lock once `now ≥ refundAfterMs`, and `refundAfterMs` is an offer field committed by `contractId`,
+so both parties signed the bound before either acted. tclk/1 therefore needs no coordination
+deadline field and no abort frame to make the cliff visible — `fundingOutlook(state, nowMs)`
+reports it, distinguishing an accepted contract whose window is still open from one whose window
+shut with no lock, from one explicitly cancelled, from one already funded. It is a derived
+reading, never a signed frame and never contract state: absence of funding is not something a
+vanished party can be made to attest, and a clock passing is not evidence that anyone intended to
+abort. A safety margin on top of the bound is rail-specific and belongs to the caller, which is
+why the reading reports the time remaining rather than only a verdict. `claimByMs` is not a
+boundary here: no transition guard reads it, and §3.1 calls it the payee's *safe* claim deadline.
+
 ## 5. Settlement rails
 
 A rail is anything that can hold `amount` of `asset` under the contract's statement and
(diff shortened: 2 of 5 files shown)


luch91 (approved):
I reviewed the current head, `applyFrame` deadline guard, the new `fundingOutlook` implementation, specification text, and boundary tests. The derived reading uses the same `nowMs >= refundAfterMs` boundary as the lock transition, distinguishes proposed/accepted/locked/settled/cancelled states, fails closed on invalid clocks, and does not mutate state. The successful CI build-and-test and commit-message checks support the submitted verification. No further changes requested.

luch91 (approved):
Approved. The exact head adds a derived, non-wire `fundingOutlook` reading and keeps it separate from signed state. Its boundary is tied to the existing `applyFrame` lock guard, invalid clocks fail closed, and tests cover the accepted/proposed cliffs plus locked, cancelled, and settled states. SPEC wording matches the implementation, and GitHub build-and-test passed for this commit.

PR #94, opened 2026-09-04
fix(transcript): report the deal room's gap, so a deleted row stops folding clean

### LINKED ISSUE #93: `foldTranscript` trusts the export's row order and never checks for a gap — deleting one signed row folds a claimed deal to `refunded`, with no `BAD` verdict
**Checked against `main` @ `5cc4ab9`** — built locally, and the reproduction below is a plain
script against `dist/`, as in #36.

`foldTranscript` (`src/transcript.ts:236`) authenticates every record and then applies them in
whatever order the array arrived in. It never checks `seq` monotonicity, never checks for a gap,
and never checks that the stream is complete. The only `seq` reference in the fold is
`seq: record?.seq ?? -1` (`src/transcript.ts:241`), which labels a verdict and validates nothing.

`examples/audit-export.mjs:40` folds `[handshake.offer, handshake.accept, ...deal]`, where `deal`
is `parseTranscriptExport` over the file's lines in file order, with no sort. So the party that
supplies `deal.jsonl` chooses the fold order — and which rows exist at all.

## What I observe

Five genuine records from one deal. The payee revealed inside the window; the payer later posted a
`refund` that the live machine rejected, but the venue stored and signed it anyway. All five
signatures verify in every case below, and no timestamp is altered — this does not depend on
`ts` being unsigned.

```
contract : 0x3fe76d7554df72196bb9c8dd1c6d4c64d532af08bfe2838ccd777d27d5fdc122
deal room: mb-p-tclk-3fe76d7554df7219

A. honest order (what actually happened)
  every signature verifies : true
  seq order as folded      : 1,2,1,2,3
  [0] offer  seq=1 ok=true
  [1] accept seq=2 ok=true
  [2] lock   seq=1 ok=true
  [3] reveal seq=2 ok=true
  [4] refund seq=3 ok=false refund in status claimed
  => status: claimed

B. same five records, two lines swapped
  every signature verifies : true
  seq order as folded      : 1,2,1,3,2
  [0] offer  seq=1 ok=true
  [1] accept seq=2 ok=true
  [2] lock   seq=1 ok=true
  [3] refund seq=3 ok=true
  [4] reveal seq=2 ok=false reveal in status refunded
  => status: refunded

C. same records, the reveal row deleted
  every signature verifies : true
  seq order as folded      : 1,2,1,3
  [0] offer  seq=1 ok=true
  [1] accept seq=2 ok=true
  [2] lock   seq=1 ok=true
  [3] refund seq=3 ok=true
  => status: refunded
```

**C is the sharp one.** B produces the wrong status but still prints a `BAD` line an attentive
reader could catch. C prints **no `BAD` line at all** — every verdict is `ok`, the fold reaches a
terminal status, and `audit-export.mjs:52` exits `0`. The only trace is the deal room's `seq`
running `1,3`, which the tool prints and nothing checks.

A `refund` at `refundAfterMs + 5s` against a `locked` contract is a perfectly legal transition.
Removing the row that would have made it illegal is the entire attack.

## What a hostile counterparty gets

Per CONTRIBUTING's "say what a hostile counterparty gets from it, or say 'nothing'":

A payer who received the work, and whose payee genuinely claimed, can hand any third party — an
arbiter under §8.1, a reputation layer consuming `receipt` frames, a future dispute process — a
`deal.jsonl` that reproduces as `refunded` under the project's own shipped auditor, with every
signature intact and a zero exit code. The payee's counter-evidence is a *different* file, and
nothing in the format lets a third party tell which of the two is complete.

Symmetrically, a payee can drop a `lock` row to make a funded deal look unfunded.

This needs no forged signature, no key, no clock manipulation, and no cooperation from the venue.
It needs a text editor.

## Which is wrong, `SPEC.md` or the code

I treated **the code as correct about what it does, and the surrounding claims as overselling it.**

- `foldTranscript`'s own doc comment — "Authenticate and fold records in the supplied order" — is
  accurate and even says "supplied order" out loud.
- SPEC §2 says an offline reader "trusts the export file" for `seq` and `ts`. That is a statement
  about *values*, and it is honest. It does not say the reader also trusts the file for
  *completeness*, which is the stronger property the fold actually needs.
- README calls the fold the thing that "authenticates every record" and `examples/` presents
  `audit-export.mjs` as an "offline audit of a finished deal". Both are true about signatures and
  false about outcomes.

So: no wire-format change is implied, and no golden vector moves. What is missing is a guard, or
an explicit precondition, or both.

## Candidate fixes

**A — the fold hard-refuses a non-contiguous stream.** Cheapest to state, and I do not recommend
it. SPEC §2 says rooms are a ring reaped after 7 idle days and that `seq` restarts if a room is
recreated, so a legitimately windowed or post-reap export would start failing, and the fold cannot
tell a censored gap from a retention boundary. It would also compound #61, where the room binding
already rejects live deals.

**B — the fold sorts by `(room, seq)` itself rather than trusting arrival order.** Closes case B.
Does nothing for case C, which is the one with no `BAD` line. Worth doing, not sufficient alone.

**C — report contiguity instead of enforcing it.** Add a per-room gap marker to `steps`, and a
`contiguous: boolean` (or `gaps: Array<{room, after, before}>`) to `TranscriptFoldResult`. Callers
that legitimately hold a partial window keep working; `audit-export.mjs` prints a warning and
exits non-zero; nothing is refused that works today. This is the one I'd propose.

**D — document the precondition.** State in SPEC §2 and in `foldTranscript`'s doc that the fold
reproduces an outcome only over a *complete, per-room-ordered* stream, and have
`audit-export.mjs` say so in its output. Cheap, and needed regardless of which of A–C lands,
because a reader of the README currently has no way to know.

**Proposed: B + C + D.** Sort what can be sorted, report what cannot be verified, and stop
implying the fold is evidence when it is a replay. I'm happy to open a PR for that shape if you
agree with it — flagging first per CONTRIBUTING's "discuss a change to the state machine in an
issue first", since `TranscriptFoldResult` is public surface.

## Scope notes

- **Independent of the unsigned-`ts` question.** Every timestamp above is genuine. There is a
  separate, sharper issue where a holder edits `ts` to move a record across a deadline; I have
  kept it out of this report so the two can be judged apart.
- **Not #61.** That is the room binding rejecting live deals. This is the fold accepting a stream
  it should not.
- **Related to #23** only in that #23's fix is what made the fold depend on per-record venue
  metadata in the first place.
- **Not #83.** `findContractHandshake` not re-deriving the contract id is a separate hole on the
  offer/accept half; this one is entirely in the deal-room half, and persists even if #83's fix
  (PR #84) lands.

## Reproduction

`node repro.mjs` from the repo root, after `pnpm -r --include-workspace-root build`. Contract id
and deal room vary per run — the offer nonce and the hash lock are random; the three folds do not.

```js
// Reproduction: foldTranscript trusts caller-supplied record order.
import { ed25519 } from "@noble/curves/ed25519.js";
import { base58, base64urlnopad } from "@scure/base";
import {
  makeOffer, makeAccept, encodeFrame, generateHashLock,
  foldTranscript, verifyTranscriptRecord, dealRoom, OFFER_ROOM,
} from "./dist/index.js";

const enc = new TextEncoder();
const did = (pk) => "did:key:z" + base58.encode(new Uint8Array([0xed, 0x01, ...pk]));

function party(seedByte) {
  const sk = new Uint8Array(32).fill(seedByte);
  return { sk, did: did(ed25519.getPublicKey(sk)) };
}
const payer = party(1);
const payee = party(2);

// `seq` and `timestampMs` are venue metadata; the signature covers `<room>|<nonce>|<line>`.
let nonceCounter = 1000;
const rec = (room, seq, timestampMs, who, line) => {
  const nonce = String(nonceCounter++);
  const sig = ed25519.sign(enc.encode(`${room}|${nonce}|${line}`), who.sk);
  return { room, seq, timestampMs, sender: who.did, nonce,
           signature: base64urlnopad.encode(sig), line };
};

const T0 = 1_780_000_000_000;
const offer = makeOffer({
  from: payer.did, role: "payer", lock: "hash",
  amount: "1000000", asset: "FLOP", rails: ["paper"],
  claimByMs: T0 + 3_600_000, refundAfterMs: T0 + 7_200_000, expiresMs: T0 + 600_000,
});
const { preimage, hash } = generateHashLock();
const accept = makeAccept(offer, { from: payee.did, statement: hash });
const room = dealRoom(accept.contract);

const lock   = { type: "lock",   from: payer.did, contract: accept.contract, rail: "paper", ref: "paper:1" };
const reveal = { type: "reveal", from: payee.did, contract: accept.contract, ref: "paper:1", secret: preimage };
const refund = { type: "refund", from: payer.did, contract: accept.contract, ref: "paper:1" };

const R = {
  offer:  rec(OFFER_ROOM, 1, T0,                     payer, encodeFrame(offer)),
  accept: rec(OFFER_ROOM, 2, T0 + 60_000,            payee, encodeFrame(accept)),
  lock:   rec(room,       1, T0 + 120_000,           payer, encodeFrame(lock)),
  reveal: rec(room,       2, T0 + 1_800_000,         payee, encodeFrame(reveal)),
  refund: rec(room,       3, T0 + 7_200_000 + 5_000, payer, encodeFrame(refund)),
};

const show = (label, records) => {
  const { state, steps } = foldTranscript(records);
  console.log(`\n${label}`);
  console.log(`  every signature verifies : ${records.every((r) => verifyTranscriptRecord(r).ok)}`);
  console.log(`  seq order as folded      : ${records.map((r) => r.seq).join(",")}`);
  for (const s of steps) {
    console.log(`  [${s.index}] ${String(s.type).padEnd(6)} seq=${s.seq} ok=${String(s.ok).padEnd(5)}${s.reason ? " " + s.reason : ""}`);
  }
  console.log(`  => status: ${state?.status}`);
};

console.log(`contract : ${accept.contract}`);
console.log(`deal room: ${room}`);
show("A. honest order (what actually happened)", [R.offer, R.accept, R.lock, R.reveal, R.refund]);
show("B. same five records, two lines swapped", [R.offer, R.accept, R.lock, R.refund, R.reveal]);
show("C. same records, the reveal row deleted", [R.offer, R.accept, R.lock, R.refund]);
```

thangvmt:
Closes #93 (case C).

Deleting one signed row from a derived deal room folds a claimed deal to `refunded` with every
step `ok`, no `BAD` line, and `audit-export.mjs` exiting 0. Nothing in the fold looks at what
the room should contain, so a censored transcript is indistinguishable from a complete one.

## The rule, and why it is stated per room

`dealRoomSpan(records, contract)` reports the verified-row continuity of the contract's derived
deal room. That room is addressed by the contract id, so it carries that contract's post-accept
frames and nothing else, which is what makes a hole in its sequence evidence. The shared board
supports no such reading: thousands of contracts interleave there, so a gap between one
contract's rows is the normal case and carries no information.

It is also why the rule cannot be stated across the fold. `seq` restarts at 1 in every room, so
a flat monotonicity check rejects conformant two-room deals — the honest order in your case A
prints `1,2,1,2,3`, which is two rooms, not disorder.

Only rows whose signature verifies are counted, so a hole cannot be closed with unsigned
padding. Closing one requires a signed row for that room.

## What it does not prove, said in the output

You established that renumbering defeats counting, and that costs one character. Dropping the
last row leaves no hole either. So the tool prints `no gap detected` and states both limits on
the next line, rather than anything that reads as "transcript complete":

```
mb-p-tclk-08ae6a078f71eb46: 2 verified rows, seq 1..3 — GAP
  "no gap detected" is not a completeness proof: seq is venue metadata outside the
  signature, so renumbering the kept rows, or dropping the last one, leaves no gap.
```

It exits non-zero on a gap, and only then. A cancelled contract has no deal-room rows at all
and still exits 0.

## Evidence

Same transcript shape as the issue, offer and accept on the board, `lock`/`reveal`/`refund` in
the deal room.

Honest export, unchanged behaviour:

```
ok  mb-p-tclk-08ae6a078f71eb46#1 lock
ok  mb-p-tclk-08ae6a078f71eb46#2 reveal
BAD mb-p-tclk-08ae6a078f71eb46#3 refund — refund in status claimed
mb-p-tclk-08ae6a078f71eb46: 3 verified rows, seq 1..3 — no gap detected
fold → claimed          exit 0
```

Reveal row deleted, which is case C:

```
ok  mb-p-tclk-08ae6a078f71eb46#1 lock
ok  mb-p-tclk-08ae6a078f71eb46#3 refund
mb-p-tclk-08ae6a078f71eb46: 2 verified rows, seq 1..3 — GAP
fold → refunded
a signed row is missing from the deal room, so this fold is not the whole deal
exit 1
```

Four tests: the whole deal, case C, the renumbered variant that is deliberately **not**
detected, and the unsigned row that cannot close a hole. Removing `dealRoomSpan` fails all
four; with it the CI trio is green (108 + 40 + 33).

## Credit

@alitiknazoglu found this and, in the thread, established the renumbering limit that shapes the
wording here. Happy to hand this over or close it if you would rather ship it yourself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ All notable changes to this project are documented here. Format follows
 
 ### Added
 
+- `dealRoomSpan(records, contract)` reports the verified-row continuity of one contract's
+  derived deal room, so an auditor sees that a signed row is missing instead of folding a
+  censored transcript to a clean terminal status. `examples/audit-export.mjs` prints it and
+  exits non-zero on a gap. The rule is stated per room because `seq` restarts in every room,
+  and the wording is "no gap detected" rather than "complete" because `seq` sits outside the
+  signed preimage (#93).
+
 - A schema-owned tclk/1 frame field contract, canonical settlement-rail registry and
   intersection-based, order-independent rail matching helpers. Generated decoder fields
   and the normative `SPEC.md` table are checked for drift in CI.
--- a/examples/audit-export.mjs
+++ b/examples/audit-export.mjs
@@ -56,9 +56,12 @@ console.log(`\n${span.room}: ${span.count} verified rows, seq ${range} — ${ver
 if (span.count === 0) {
   console.log("  an empty deal room reads the same whether it was censored or simply expired: the");
   console.log("  venue deletes a room after seven days with no write, and a terminal deal stops writing.");
-} else {
+} else if (span.gapFree) {
   console.log("  \"no gap detected\" is not a completeness proof: seq is venue metadata outside the");
   console.log("  signature, so renumbering the kept rows, or dropping the last one, leaves no gap.");
+} else {
+  console.log("  a position between the first and last row is missing: the room held a signed row");
+  console.log("  that this file does not.");
 }
 
 const terminal = ["claimed", "refunded", "cancelled"].includes(folded.state.status);
--- a/examples/audit-export.mjs
+++ b/examples/audit-export.mjs
@@ -56,18 +56,23 @@ console.log(`\n${span.room}: ${span.count} verified rows, seq ${range} — ${ver
 if (span.count === 0) {
   console.log("  an empty deal room reads the same whether it was censored or simply expired: the");
   console.log("  venue deletes a room after seven days with no write, and a terminal deal stops writing.");
-} else if (span.gapFree) {
-  console.log("  \"no gap detected\" is not a completeness proof: seq is venue metadata outside the");
-  console.log("  signature, so renumbering the kept rows, or dropping the last one, leaves no gap.");
-} else {
-  console.log("  a position between the first and last row is missing: the room held a signed row");
-  console.log("  that this file does not.");
 }
 
+// Two claims, kept apart on purpose. The replay says what these records fold to. It is not a
+// statement that these are all the records, and a tool that prints one line for both invites
+// the reader to hear the second.
 const terminal = ["claimed", "refunded", "cancelled"].includes(folded.state.status);
-console.log(`\nfold → ${folded.state.status}${terminal ? "" : " (not terminal)"}`);
+console.log(`\nreplay   → ${folded.state.status}${terminal ? "" : " (not terminal)"}`);
+
 if (span.count > 0 && !span.gapFree) {
-  console.error("a signed row is missing from the deal room, so this fold is not the whole deal");
+  console.log("evidence → INCOMPLETE. A position between the first and last row is missing, so the");
+  console.log("           room held a signed row that this file does not.");
   process.exit(1);
 }
+
+console.log("evidence → completeness NOT established. A gap is evidence of absence; the absence of");
+console.log("           a gap is not evidence of presence, because seq is venue metadata outside the");
+console.log("           signature: a supplier that renumbers the rows it keeps leaves none, and one");
+console.log("           that drops the last row leaves none either.");
+console.log("           Exit 0 means the replay reached a terminal status. It does not mean audited.");
 process.exit(terminal ? 0 : 1);
--- a/examples/audit-export.mjs
+++ b/examples/audit-export.mjs
@@ -51,9 +51,15 @@ if (folded.state === null) {
 
 const span = dealRoomSpan(deal, contract);
 const range = span.count === 0 ? "none" : `${span.firstSeq}..${span.lastSeq}`;
-console.log(`\n${span.room}: ${span.count} verified rows, seq ${range} — ${span.gapFree ? "no gap detected" : "GAP"}`);
-console.log("  \"no gap detected\" is not a completeness proof: seq is venue metadata outside the");
-console.log("  signature, so renumbering the kept rows, or dropping the last one, leaves no gap.");
+const verdict = span.count === 0 ? "no rows" : span.gapFree ? "no gap detected" : "GAP";
+console.log(`\n${span.room}: ${span.count} verified rows, seq ${range} — ${verdict}`);
+if (span.count === 0) {
+  console.log("  an empty deal room reads the same whether it was censored or simply expired: the");
+  console.log("  venue deletes a room after seven days with no write, and a terminal deal stops writing.");
+} else {
+  console.log("  \"no gap detected\" is not a completeness proof: seq is venue metadata outside the");
+  console.log("  signature, so renumbering the kept rows, or dropping the last one, leaves no gap.");
+}
 
 const terminal = ["claimed", "refunded", "cancelled"].includes(folded.state.status);
 console.log(`\nfold → ${folded.state.status}${terminal ? "" : " (not terminal)"}`);
--- a/examples/audit-export.mjs
+++ b/examples/audit-export.mjs
@@ -10,6 +10,7 @@ import { readFileSync } from "node:fs";
 import {
   OFFER_ROOM,
   dealRoom,
+  dealRoomSpan,
   findContractHandshake,
   foldTranscript,
   parseTranscriptExport,
@@ -48,6 +49,16 @@ if (folded.state === null) {
   process.exit(1);
 }
 
+const span = dealRoomSpan(deal, contract);
+const range = span.count === 0 ? "none" : `${span.firstSeq}..${span.lastSeq}`;
+console.log(`\n${span.room}: ${span.count} verified rows, seq ${range} — ${span.gapFree ? "no gap detected" : "GAP"}`);
+console.log("  \"no gap detected\" is not a completeness proof: seq is venue metadata outside the");
+console.log("  signature, so renumbering the kept rows, or dropping the last one, leaves no gap.");
+
 const terminal = ["claimed", "refunded", "cancelled"].includes(folded.state.status);
 console.log(`\nfold → ${folded.state.status}${terminal ? "" : " (not terminal)"}`);
+if (span.count > 0 && !span.gapFree) {
+  console.error("a signed row is missing from the deal room, so this fold is not the whole deal");
+  process.exit(1);
+}
 process.exit(terminal ? 0 : 1);
--- a/src/index.ts
+++ b/src/index.ts
@@ -36,11 +36,11 @@ export type { TclkStatus, ContractState, StepResult } from "./machine.js";
 
 export {
   verifyTranscriptRecord, transcriptRecord, parseTranscriptExport,
-  findContractHandshake, foldTranscript,
+  findContractHandshake, foldTranscript, dealRoomSpan,
 } from "./transcript.js";
 export type {
   TranscriptRecord, TranscriptRecordVerification, TranscriptStep, TranscriptFoldResult,
-  ContractHandshake,
+  ContractHandshake, DealRoomSpan,
 } from "./transcript.js";
 
 export { lockTerms, MemoryRail } from "./rail.js";
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -131,6 +131,10 @@ export interface DealRoomSpan {
  * Only rows whose signature verifies are counted, so a hole cannot be closed with unsigned
  * padding - closing one needs a signed row for that room, which cannot be forged.
  *
+ * `firstSeq === 1` is required because a derived room is small and holds one contract, so it
+ * never loses a prefix. It is not a rule that generalizes: the venue's ring evicts from the
+ * front, so a busy room retains a contiguous window that does not begin at 1.
+ *
  * `gapFree` is **not** a completeness proof, and an auditor that reports it as one is worse
  * than one that reports nothing. `seq` is venue metadata outside the signed `room|nonce|line`
  * preimage, so an editor who drops a row and renumbers the rows it keeps leaves no hole, and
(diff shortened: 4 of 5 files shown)


alitiknazoglu:
The shape is right, and scoping the rule to the derived room with `firstSeq === 1` documented as a property of that room rather than a general rule is the part I would have got wrong. Two things from reading it, one of which I think blocks.

## `seqs.length` counts rows, not distinct positions

`gapFree` compares a row count against a span width:

```ts
gapFree: firstSeq === 1 && seqs.length === lastSeq - firstSeq + 1,
```

Nothing dedupes, so a repeated row inflates `seqs.length` and can close a hole arithmetically. `[1, 1, 3]` gives length 3, span 3, and `gapFree === true`.

That is reachable in the threat model this PR is written against, because the file is supplied by a party and duplicating a line in it costs nothing — the duplicate is a genuine signed row, so it verifies like any other. Against your branch, using the issue's transcript plus one heartbeat:

```
honest deal (1,2,3,4)
  ok  …#1 lock
  ok  …#2 heartbeat
  ok  …#3 reveal
  BAD …#4 refund — refund in status claimed
  span: count=4 seq=1..4 gapFree=true
  => claimed    BAD rows: 1    exit 0

reveal deleted, heartbeat row repeated (1,2,2,4)
  ok  …#1 lock
  ok  …#2 heartbeat
  ok  …#2 heartbeat
  ok  …#4 refund
  span: count=4 seq=1..4 gapFree=true
  => refunded   BAD rows: 0    exit 0
```

The second fold is case C with the reveal removed, and it is clean: no gap, no `BAD` row, exit 0. It reads *better* than the honest export, which at least prints one `BAD`. A heartbeat is the convenient row to duplicate because §4 makes it state-neutral while locked, so re-applying it is `ok` rather than a rejection — but the arithmetic is what fails, and any repeated row does it.

Counting distinct positions fixes it:

```ts
const seen = new Set<number>();
// …
  seen.add(record.seq);
// …
gapFree: firstSeq === 1 && seen.size === lastSeq - firstSeq + 1,
```

`[1, 2, 2, 4]` then gives `seen.size === 3` against a span of 4, so `GAP`. I would leave `count` as the row count, since the output line says "verified rows" and that is still the honest number to print.

Worth a fifth test alongside the four you have: a duplicated row must not close a hole, next to the unsigned row that already cannot.

## The caveat quotes a verdict it did not give

The explanatory line is in the `else` branch, so it prints under both non-empty verdicts:

```
mb-p-tclk-08ae6a078f71eb46: 2 verified rows, seq 1..3 — GAP
  "no gap detected" is not a completeness proof: …
```

It says `GAP`, then discusses "no gap detected". The sentence is the right one to print when the verdict is `no gap detected`; under `GAP` it reads as though the tool contradicted itself. Splitting the two branches, or dropping the quoted phrase when a gap was found, would settle it.

thangvmt:
Both correct, and the duplicate is the worse one. Fixed in `b61e01e`.

`gapFree` now compares distinct positions and `count` stays the row count, so the printed line
still says the honest number beside "verified rows". Your `1,2,2,4` is the fifth test. I checked
it fails against the previous counting rather than assuming it would, because a test that passes
for the wrong reason is exactly what let this through.

The explanatory line is split by verdict. A `GAP` run now says what a gap means instead of
discussing what "no gap detected" does not prove:

```
mb-p-tclk-ecf1a9fca2889878: 2 verified rows, seq 1..3 — GAP
  a position between the first and last row is missing: the room held a signed row
  that this file does not.
```

One thing worth adding about the heartbeat, because it is sharper than convenience. Reading
`applyFrame` for `locked`: `lock` is rejected on status, `reveal` and `refund` both move state,
`cancel` and `offer` and `accept` are rejected, and `receipt` requires a terminal status. That
leaves `heartbeat` as **the only frame that re-applies as `ok` in `locked` without moving
state**. So it is not merely the convenient row to duplicate, it is the only one that repeats
without leaving a `BAD` line for a reader to notice. A counting rule that relied on duplicates
being visible would have had exactly one blind spot, and §4 put a frame in it.

CI on the branch: 109 + 40 + 33.

sv:
Closing as @thangvmt proposed on #93 — "#94 should not merge as it stands" — with #97 as the home for the check and three pieces to carry across.

Recording the reasoning here so it is not only in that thread. @alitiknazoglu ran both branches against one matrix and neither is a superset of the other: this branch's `firstSeq === 1` catches a deleted opening row that #97 misses, #97's pairwise walk catches a scrambled supply order that this branch misses (it sorts before comparing, so the supplied order is gone by the time the arithmetic runs), and both miss end-truncation and the relabelling case. `warnings` on `TranscriptFoldResult` is the better home than a second exported function because it reaches `tclk_apply_transcript` — an agent folding a transcript through the MCP server gets no completeness signal at all today, and a function you have to know to call does not change that.

Three things from this branch worth landing on #97:

- the `firstSeq === 1` anchor, kept with its reasoning attached rather than as a bare condition — it holds because a derived deal room is small and single-contract, and it does not generalise to a room whose front the ring has evicted (see @hayulpapax's `/r/meta` reading on #93);
- the verified-rows filter on whatever counts positions, which #97 has already adopted;
- the `replay` / `evidence` split in the tool's output. This is the part I would least want to lose — it is the only text in the repo that tells a reader what a clean fold does *not* buy them, and "no gap detected" must never read as "transcript complete".

@thangvmt — a patch onto #97 is the shape to send, as you offered. Thank you for the measurement work on this and for calling it yourself; the empty-deal-room correction and the two-clocks reading were both worth the thread on their own.

---

PR #95, opened 2026-09-04
fix(decode): reject non-printable-ASCII lines in decodeFrame

ZedLab1:
encodeFrame() already refuses to emit a line containing a non-printable-ASCII
byte, because technocore's real single-line sweep would rewrite such bytes
before storing them. decodeFrame() had no matching guard, so a line no
conforming encoder could ever produce (e.g. a raw U+007F byte in a free-text
field like job.id) would still decode into a frame that looked legitimate,
with a matching offerId.

This is not a fund-safety issue (no rail here custodies value yet), but it
sits in the "canonical encoding" category from SECURITY.md: the bytes a
frame's id/signature commit to should not be able to diverge silently from
what a conforming implementation could actually produce. It matters most for
offline transcript auditing, where decodeFrame is the only gate on untrusted
exported lines.

Included:
- A regression test (tests/decode-sweep-guard.test.ts) that fails without the
  fix and passes with it; also checks legitimate non-ASCII content still
  round-trips unaffected.
- The one-line guard in decodeFrame, mirroring encodeFrame's existing check.
- CHANGELOG entry under [Unreleased].

Verified locally:
→ all passing except an unrelated, pre-existing schema-conformance failure
caused by Windows line-ending differences in a generated file, unrelated to
this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ All notable changes to this project are documented here. Format follows
 ## [Unreleased]
 
 ### Fixed
+- `decodeFrame` now rejects a line containing a raw non-printable-ASCII byte, matching the
+  guard `encodeFrame` already enforces on emission. A conforming sender can never produce
+  such a line (technocore's real single-line sweep would rewrite it before storing), so a
+  decoded line that fails the same test could not have reached a reader this way either.
+  Previously such a line would decode into a frame that looked legitimate — a gap for any
+  offline transcript audit. Legitimately-escaped non-ASCII content is unaffected.
+
+### Fixed
 
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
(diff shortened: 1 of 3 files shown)


sv:
Closing as a duplicate of #66 — both add the printable-ASCII guard to `decodeFrame`, so a line the venue would have swept cannot be accepted as the line its sender signed.

Keeping #66 because it lifts the pattern into a single named `PRINTABLE_ASCII` constant used by both halves of the codec. The encoder has had this guard since the sweep fix; having the decoder's copy be a second inline literal is how the two drift apart later, and this rule is one where drift means the id no longer commits to the bytes on the wire.

One thing to flag on this branch regardless: the diff re-indents the existing `if (text.length > MAX_FRAME_CHARS) {` line by two spaces, which would have failed the format check. Same diagnosis as #66 though, and arrived at independently — thanks.

---

ZedLab1:
Makes sense, thanks for checking — good to know the diagnosis matched independently. Will keep the shared-constant approach in mind

PR #97, opened 2026-09-04
fix(transcript): surface unsigned ts/seq trust boundary and per-room gaps

### LINKED ISSUE #96: foldTranscript uses unsigned venue ts for deadline guards - editing ts flips claimed↔refunded with all signatures valid
Summary
In src/transcript.ts:103 the Ed25519 signature covers only `room|nonce|line`:
`${room}|${nonce}|${line}`
`seq` and `ts` are venue metadata parsed in src/transcript.ts:131 via `Date.parse(message.ts)` and then used as `nowMs` in every deadline guard in src/machine.ts:121,158,178,192.

A file supplier (payer handing `deal.jsonl` to an arbiter, same model as #93) can rewrite `ts` without breaking any signature.

The Problem
- Honest deal: `reveal` at T+30m (< refundAfterMs 2h) → `claimed`, `refund` after → rejected. Correct.
- Same 5 signed records, only `reveal.ts` edited to T+2h+10s (> refundAfterMs), `refund.ts` unchanged → `reveal` now rejected “refund window is open”, `refund` accepted → `refunded`. Every `verifyTranscriptRecord` still `ok==true`, `audit-export.mjs` exits 0, only `ts` string changed.
- This is the separate `ts`-malleability you reserved in #93 scope notes (“holder edits ts to move a record across a deadline, kept out so the two can be judged apart”). No open issue covers it - #93 is gap/row-order, #78 is nonce >2^53.

    const {makeOffer, makeAccept, generateHashLock, dealRoom, OFFER_ROOM, foldTranscript, verifyTranscriptRecord, encodeFrame} = await import("./dist/index.js");
    // 5 genuine records, all verifyTranscriptRecord.ok === true
    const T0=1780000000000;
    const offer=makeOffer({from:payerDid, role:"payer", lock:"hash", amount:"1000000", asset:"FLOP", rails:["paper"], claimByMs:T0+3600000, refundAfterMs:T0+7200000, expiresMs:T0+600000});
    const {preimage,hash}=generateHashLock();
    const accept=makeAccept(offer,{from:payeeDid, statement:hash});
    const room=dealRoom(accept.contract);
    // honest
    foldTranscript([offerRec,acceptRec,lockRec,revealRec@T+1800000,refundRec@T+7205000]) // => claimed
    // tampered: same records, reveal.timestampMs = T+7210000
    verifyTranscriptRecord(tamperedReveal) // => ok true (ts not signed)
    foldTranscript([offerRec,acceptRec,lockRec,tamperedReveal,refundRec]) // => refunded

Proposed Fix
In src/transcript.ts:236 `foldTranscript` should not treat `ts`/`seq` as authenticated. Add `warnings: string[]` to `TranscriptFoldResult`:
- per-room `gap detected (seq 1->3)`, `seq not strictly increasing`, `timestamp goes backwards`
- generic note when any deadline frame (`accept`/`lock`/`reveal`/`refund`) is present: “timestamps and seq are venue metadata, not covered by signature - verify settlement on the rail”

Do not refuse to fold (partial windows/post-reap are legitimate), just surface it. `examples/audit-export.mjs` and `mcp/src/tools.ts:tclk_apply_transcript` forward `warnings`; audit exits 1 on gap/ordering/backwards. `SPEC.md:62` documents that `ts`/`seq` being unsigned means a supplier can rewrite `ts` to flip deadlines - transcript is coordination, not settlement proof.

PR `fix/transcript-timestamp-auth` implements this - no wire change.

PASdeco:
Fixes #96

Problem
`src/transcript.ts:103` signatures cover `room|nonce|line` only. `seq`/`ts` are venue metadata (`transcriptRecord:131` -> `Date.parse`) but `foldTranscript:236` uses `record.timestampMs` as `nowMs` for every deadline guard (`machine.ts:121,158,178`). A file supplier can rewrite `ts` and flip `claimed`↔`refunded` with all signatures valid. Same silent flip happens if a row is deleted (#93) - gap, no BAD verdict.

This PR does not refuse to fold (partial windows/post-reap are legitimate), it reports:

- `TranscriptFoldResult.warnings: string[]`
- per-room `gap detected (seq 1->3)`, `seq not strictly increasing`, `timestamp goes backwards`
- generic note when any deadline frame (`accept`/`lock`/`reveal`/`refund`) is present: “timestamps and seq are venue metadata, not covered by signature, verify settlement on the rail”

`examples/audit-export.mjs` prints `WARN:` and exits 1 on gap/ordering/backwards; generic note is informational. `mcp/src/tools.ts:tclk_apply_transcript` forwards `warnings`. `SPEC.md:62` documents the boundary: transcript is coordination, not settlement proof.

Verification
- `pnpm build && pnpm test` - 103/104 (one `schema-conformance` CRLF artifact pre-existing #90), `mcp` 40/40, `worker` 33/33
- Repro `repro_warn.mjs` (honest `claimed` -> generic note only; tampered `reveal.ts=T+7210000` -> `timestamp goes backwards` + `refunded`; deleted `reveal` -> `gap detected` + `refunded`)

No wire change. Follow-up could sort by `(room,seq)` (#93 B) but reporting is enough for audit use.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -62,9 +62,15 @@ the public manual (`/llms.txt`), and any self-hosted deployment works identicall
   `<room>|<nonce>|<text>`; `seq` and `ts` are venue metadata, not sender-signed fields. Deadline
   guards replay at that record's `ts`, so a live reader trusts the venue for time and an
   offline reader trusts the export file for it. Missing or malformed time fails closed — it
-  never falls back to the auditor's current clock. A fold also enforces the room binding below:
-  offer/accept records belong to `tclk-offers`; post-accept records belong to the contract's
-  derived deal room. A valid signature in the wrong room cannot advance state.
+  never falls back to the auditor's current clock. `ts`/`seq` being unsigned means a file
+  supplier (e.g. a payer handing a `deal.jsonl` to an arbiter) can rewrite `ts` to move a
+  reveal across `refundAfterMs` and flip `claimed`↔`refunded` with every signature still
+  valid — the transcript is coordination, not settlement proof; verify settlement on the
+  rail. `foldTranscript` surfaces this boundary via `warnings` (gaps, seq/timestamp
+  monotonicity, and the generic unsigned-timestamp note). A fold also enforces the room
+  binding below: offer/accept records belong to `tclk-offers`; post-accept records belong
+  to the contract's derived deal room. A valid signature in the wrong room cannot advance
+  state.
 - **Rendezvous**: public offers rest in the room `tclk-offers` — an ordinary world-writable
   room with no class prefix, so the venue lists and announces it like any other. Two agents who
   have never met have nowhere else to find each other, so a deal cannot start without a
--- a/examples/audit-export.mjs
+++ b/examples/audit-export.mjs
@@ -50,7 +50,7 @@ if (folded.warnings?.length) {
   // A backwards timestamp can flip a deadline with all signatures valid.
   const fatal = folded.warnings.some((w) => w.includes("gap detected") || w.includes("seq not strictly increasing") || w.includes("timestamp goes backwards"));
   if (fatal) {
-    console.error("\ntranscript is not per-room contiguous/monotonic — refusing to treat fold as audit proof");
+    console.error("\nevidence → INCOMPLETE: transcript is not per-room contiguous/monotonic — a signed row is missing or reordered. Refusing to treat fold as audit proof");
     process.exit(1);
   }
 }
@@ -61,8 +61,6 @@ if (folded.state === null) {
 }
 
 const terminal = ["claimed", "refunded", "cancelled"].includes(folded.state.status);
-console.log(`\nfold → ${folded.state.status}${terminal ? "" : " (not terminal)"}`);
-if (folded.warnings?.length) {
-  console.log("note: timestamps/seq are venue metadata, not covered by signature — verify settlement on the rail");
-}
+console.log(`\nreplay → ${folded.state.status}${terminal ? "" : " (not terminal)"}`);
+console.log("evidence → completeness NOT established. A gap is evidence of absence; the absence of a gap is not evidence of presence, because seq/ts are venue metadata outside the signature (room|nonce|line only). Exit 0 means the replay reached a terminal status. It does not mean audited — verify settlement on the rail.");
 process.exit(terminal ? 0 : 1);
--- a/examples/audit-export.mjs
+++ b/examples/audit-export.mjs
@@ -43,6 +43,18 @@ for (const step of folded.steps) {
   console.log(`${verdict} ${step.room}#${step.seq} ${step.type ?? "record"}${step.reason ? ` — ${step.reason}` : ""}`);
 }
 
+if (folded.warnings?.length) {
+  console.error("\nwarnings (timestamps/seq are venue metadata, not signed — see SPEC §2):");
+  for (const w of folded.warnings) console.error(`  WARN: ${w}`);
+  // A gap or reordering can flip claimed↔refunded with no BAD verdict (see #93).
+  // A backwards timestamp can flip a deadline with all signatures valid.
+  const fatal = folded.warnings.some((w) => w.includes("gap detected") || w.includes("seq not strictly increasing") || w.includes("timestamp goes backwards"));
+  if (fatal) {
+    console.error("\ntranscript is not per-room contiguous/monotonic — refusing to treat fold as audit proof");
+    process.exit(1);
+  }
+}
+
 if (folded.state === null) {
   console.error("no authenticated contract could be opened");
   process.exit(1);
@@ -50,4 +62,7 @@ if (folded.state === null) {
 
 const terminal = ["claimed", "refunded", "cancelled"].includes(folded.state.status);
 console.log(`\nfold → ${folded.state.status}${terminal ? "" : " (not terminal)"}`);
+if (folded.warnings?.length) {
+  console.log("note: timestamps/seq are venue metadata, not covered by signature — verify settlement on the rail");
+}
 process.exit(terminal ? 0 : 1);
--- a/mcp/src/tools.ts
+++ b/mcp/src/tools.ts
@@ -285,6 +285,7 @@ export function createHandlers(options: HandlerOptions = {}) {
         // caller already holds, and this server never republishes secret material.
         secretRevealed: open.secret !== undefined,
         steps: folded.steps,
+        warnings: folded.warnings,
       };
     },
 
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -251,8 +251,12 @@ export function foldTranscript(records: readonly TranscriptRecord[]): Transcript
   // These do not refuse to fold — a partial window or post-reap export is legitimate —
   // but a silent gap or timestamp edit can flip a deadline-dependent outcome (reveal vs
   // refund) with every signature intact. Surface it instead of failing closed.
+  // Only verified rows count for gaps — an unsigned/BAD row still shows as BAD in
+  // `steps` but must not make a censored verified seq look contiguous (e.g. verified
+  // 1,3 padded with unverified 2 should still be a gap).
   const byRoom = new Map<string, TranscriptRecord[]>();
   for (const r of records) {
+    if (!verifyTranscriptRecord(r).ok) continue;
     const list = byRoom.get(r.room) ?? [];
     list.push(r);
     byRoom.set(r.room, list);
@@ -277,8 +281,9 @@ export function foldTranscript(records: readonly TranscriptRecord[]): Transcript
       }
     }
   }
-  // Generic trust-boundary warning when any deadline-sensitive frame is present.
+  // Generic trust-boundary warning when any verified deadline-sensitive frame is present.
   const hasDeadlineFrame = records.some((r) => {
+    if (!verifyTranscriptRecord(r).ok) return false;
     const f = tryDecodeFrame(r.line);
     return f !== null && (f.type === "accept" || f.type === "lock" || f.type === "reveal" || f.type === "refund");
   });
--- a/src/transcript.ts
+++ b/src/transcript.ts
@@ -52,6 +52,7 @@ export interface TranscriptStep {
 export interface TranscriptFoldResult {
   state: ContractState | null;
   steps: TranscriptStep[];
+  warnings: string[];
 }
 
 export interface ContractHandshake {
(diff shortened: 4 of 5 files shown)


thangvmt:
Ran this branch against the transcript shapes from #93 rather than reading it, because the two
of us are checking overlapping things in the same two files and it seemed better to know than to
guess. It holds up on both of the cases I had, and misses one I did not expect.

Method: one contract, offer and accept on the board, `lock`/`reveal`/`refund` in the derived
room, exit code taken from the process rather than from a pipeline.

```
honest                              exit 0   1 warning, no refusal
reveal deleted        (1,_,3)       exit 1   gap detected, refuses
reveal deleted + row repeated (1,2,2,4)  exit 1   seq not strictly increasing, then gap
```

The duplicate case is the one I want to flag as a point in your favour. Your check compares
consecutive pairs, so `2 -> 2` trips "not strictly increasing" before the arithmetic matters. I
had the same check written as a count against a span width, and it read `1,2,2,4` as whole until
@alitiknazoglu found it an hour ago. Pairwise is the more robust shape and I would say so even
if the PRs were not adjacent.

## The case it misses

Delete the reveal and put an **unsigned** row in its place, so the positions read `1,2,3` again.
Same contract, same minute, both branches:

```
#97   warnings: the generic ts/seq trust boundary only
      no gap warning, no refusal
      fold → refunded                          exit 0

#94   mb-p-tclk-…: 2 verified rows, seq 1..3 — GAP
      a signed row is missing from the deal room
      fold → refunded                          exit 1
```

The fold marks the padding row `BAD`, so it is not invisible, but the honest export also carries
one `BAD` (the refund arriving after the reveal). Compare the two by warnings and exit code, and
the censored file is the cleaner of the pair.

The cause is that `byRoom` is built from `records` directly, so an unverified row still occupies
a position. `foldTranscript` already computes a verdict per record; grouping only the ones that
verify closes it, and the row still shows up as `BAD` on its own line either way.

## What mine does not do

Ordering and `ts` are yours and I have nothing equivalent. My check sorts positions, so a room
supplied out of order reads clean to it, and it says nothing at all about timestamps — #96 is a
sharper statement of that than anything I had. It also only asks about a derived deal room, and
requires `firstSeq === 1` there, which yours does not need because it never claims to be a
completeness statement.

So the two seem to be complements rather than alternatives: yours catches ordering, timestamps
and repeats, mine catches an unverified row occupying a position and a missing prefix. Whether
that is one check or two is a maintainer's call, and I mention it mostly because we are both
editing `src/transcript.ts` and `examples/audit-export.mjs` and someone is going to have to
reconcile them.

PASdeco:
Thanks for running it against #93 shapes, good catch on the unsigned padding.

Fixed: `byRoom` now built from verified rows only, so `1,3` padded with unsigned `2` still trips `gap detected` (verified `1->3`), while the `BAD` row stays in `steps` as before. Your `1,2,2,4` duplicate case stays `seq not strictly increasing` via the pairwise check and agree that's more robust than span width.

Yours (verified gap + `firstSeq===1`) and this one (ordering/timestamps/repeats) do feel like complements. Well, happy to let maintainers reconcile to one check or keep both.

jokazza84 (commented):
I re-ran the current head `addd5db6d7cfffaa70f3ceb7f1de1754e94b4a4b`. The verified-row change also fixes the null-record regression from the previous head, so that part now holds.

There is still one correctness gap in the offline-audit claim: the gap detector treats `seq` as evidence of completeness even though this PR correctly documents that `seq` is not covered by the signature.

I built one authenticated contract with offer/accept on `tclk-offers` and lock/reveal/refund at deal-room seq `1,2,3`:

```
honest
  every verifyTranscriptRecord: true
  fold: claimed
  warnings: generic unsigned-metadata warning only
  audit fatal predicate: false
  resulting exit: 0

delete reveal; change refund metadata seq 3 -> 2
  every verifyTranscriptRecord: true
  fold: refunded
  warnings: the same generic warning only
  audit fatal predicate: false
  resulting exit: 0
```

No signed byte changes: the signature is over `room|nonce|line`, so changing `seq` does not invalidate the refund record. The supplied file therefore removes the claim, reports a contiguous `1,2`, changes the terminal outcome, and still passes the new selective fatal check.

The per-room warnings are useful diagnostics for accidental truncation or reordering, but an offline file cannot use unsigned `seq` to establish completeness against its supplier. Please make that limitation structural in `audit-export.mjs`: for example, do not give a terminal fold a proof-like success status when the only completeness evidence is unsigned metadata (or distinguish “replay succeeded” from “audit evidence verified”). A focused regression for delete-and-renumber should pin whichever contract you choose.

thangvmt:
@jokazza84's point applies to #94 unchanged, and I checked rather than assumed. Delete the
reveal, renumber the refund from 3 to 2, and my tool prints `no gap detected` and exits 0 on a
deal that was `claimed`. Same hole, different arithmetic — counting distinct positions does not
help when the positions are rewritten.

What I changed there, in case it is useful here. The conclusion is now two claims that cannot be
read as one:

```
replay   → refunded
evidence → completeness NOT established. A gap is evidence of absence; the absence of
           a gap is not evidence of presence, because seq is venue metadata outside the
           signature …
           Exit 0 means the replay reached a terminal status. It does not mean audited.
```

The honest transcript prints that same evidence line. Nothing either of our checks can see
distinguishes a complete file from a renumbered one, so nothing either prints should sound like
it can. A detected gap still exits 1, which is the one claim the counting does support.

The caveat was already in my output as prose and it was not enough. It sat under a line that
read like a verdict, and a reader takes the verdict.

PASdeco:
> I re-ran the current head `addd5db6d7cfffaa70f3ceb7f1de1754e94b4a4b`. The verified-row change also fixes the null-record regression from the previous head, so that part now holds.
> 
> There is still one correctness gap in the offline-audit claim: the gap detector treats `seq` as evidence of completeness even though this PR correctly documents that `seq` is not covered by the signature.
> 
> I built one authenticated contract with offer/accept on `tclk-offers` and lock/reveal/refund at deal-room seq `1,2,3`:
> 
> ```
> honest
>   every verifyTranscriptRecord: true
>   fold: claimed
>   warnings: generic unsigned-metadata warning only
>   audit fatal predicate: false
>   resulting exit: 0
> 
> delete reveal; change refund metadata seq 3 -> 2
>   every verifyTranscriptRecord: true
>   fold: refunded
>   warnings: the same generic warning only
>   audit fatal predicate: false
>   resulting exit: 0
> ```
> 
> No signed byte changes: the signature is over `room|nonce|line`, so changing `seq` does not invalidate the refund record. The supplied file therefore removes the claim, reports a contiguous `1,2`, changes the terminal outcome, and still passes the new selective fatal check.
> 
> The per-room warnings are useful diagnostics for accidental truncation or reordering, but an offline file cannot use unsigned `seq` to establish completeness against its supplier. Please make that limitation structural in `audit-export.mjs`: for example, do not give a terminal fold a proof-like success status when the only completeness evidence is unsigned metadata (or distinguish “replay succeeded” from “audit evidence verified”). A focused regression for delete-and-renumber should pin whichever contract you choose.

@jokazza84 Thanks - fixed in `aedd1c0`.

> Please make that limitation structural in `audit-export.mjs`: for example, do not give a terminal fold a proof-like success status when the only completeness evidence is unsigned metadata

Done: `audit-export` now prints `replay → status` separate from `evidence → completeness NOT established / INCOMPLETE`. `exit 0` means replay hit terminal, not audited. Honest files print it too. Gap still `INCOMPLETE exit 1`.

> A focused regression for delete-and-renumber should pin whichever contract you choose.

Added: `delete-and-renumber leaves no gap` in `tests/transcript.test.ts` - honest `claimed`, renumbered `3->2` contiguous `1,2 refunded all-ok no-gap`. Pins the limit.

thangvmt:
Patch is up as PASdeco/tclk#1 against this branch, carrying the three pieces @sv named on #94.

Two corrections to what I had on #94, both found while porting rather than after:

`audit-export.mjs` exited 0 on a censored `2,3,4` transcript. The fatal predicate recognised only `gap detected`, `seq not strictly increasing` and `timestamp goes backwards`, so the new warning printed while `replay` still said `claimed` and the process succeeded. The anchor is in that predicate now, with a CLI test that runs the real script and pins the exit code as well as the message.

The MCP fixture numbered `seq` globally across `tclk-offers` and the deal room, which started the derived room at 2 and tripped the anchor on the happy path. Nothing asserted on `warnings`, so it passed with a false warning. `records()` numbers per room now, and there is a test for propagation through `tclk_apply_transcript`.

Fresh clone at `aedd1c0` with `pnpm install`: build clean, root 108/108, mcp 41/41, worker 33/33, and red-green holds on both new tests.

@PASdeco yours to take or leave, and happy to reshape it if you would rather fold the check in differently.

luch91 (approved):
I audited the current head `aedd1c087415e453decd757565910ac8e4ed4d0c` against `main`.

The revised implementation now:

- computes per-room ordering/gap warnings from authenticated rows only, while retaining rejected rows in `steps`;
- preserves the distinction between a malformed/reordered supplied stream and a partial but otherwise valid window;
- separates `replay -> status` from `evidence -> completeness NOT established` in `audit-export.mjs`;
- keeps `tclk_apply_transcript` warnings visible to MCP callers; and
- pins the important delete-and-renumber case: all retained signatures still verify, the replay can change terminal status, and no gap warning can prove completeness when `seq` is unsigned.

The current source and tests are consistent with the documented trust boundary. GitHub does not expose a full runtime check for this fork head here, so this approval is a source-level review and does not claim an additional local test run.

khenzarr:
Independent validation against current main `5cc4ab93efbc8999a3a7e1471b639deca25998ea` and this PR head `aedd1c0632373d434033479871b037368ee209d0` found a performance regression in the new warning pass.

I used 2,000 deterministic, locally signed `TranscriptRecord` values (same valid Ed25519 signature tuple, sequential venue metadata) and timed one `foldTranscript(records)` call after building each tree:

```text
main    2000 records:  3852.5 ms
#97     2000 records: 11297.4 ms
```

The ~2.93x result matches the source path: each valid record is verified once while building `byRoom`, again inside `hasDeadlineFrame`, and a third time in the actual fold. Ed25519 verification dominates large retained exports; this makes the warning feature roughly triple the cost even for records whose lines do not decode as TCLK frames.

The functional checks otherwise pass: build passes; all 101 root tests excluding the unrelated Windows CRLF schema-conformance failure pass; MCP 40/40 and Worker 33/33 pass. Current main reproduces that same CRLF failure (#90/#121).

Could the first traversal retain each record's verification result (and decoded frame where available) and reuse it for gap analysis, deadline detection, and the fold? A focused regression can instrument or inject the verifier if you want a structural one-call-per-record assertion; otherwise a sufficiently large deterministic timing test would be much more fragile. This is not a wire/state-machine objection—the replay/evidence wording and delete-and-renumber coverage address the earlier correctness concern.

osr21:
Reconciliation result after running both current heads locally:

- **#97 @ `aedd1c0`**: full workspace build and test suite pass — 105 root tests, 40 MCP tests, 33 Worker tests.
- **#110 @ `39a74fb`**: full workspace build passes — 111 root tests. The earlier `5, 3, 4` diagnostic bug is fixed: the implementation now advances the immediate predecessor after reporting, and the regression asserts one issue at the `5 -> 3` step only.

I also attempted the merge of #110 into #97. `src/transcript.ts` merges cleanly; the only textual conflicts are `examples/audit-export.mjs` and the import/test additions in `tests/transcript.test.ts`. They are complementary, not semantic conflicts.

## Intended combined behavior

1. Preserve #97's language and output distinction: a terminal fold is a **replay**, never evidence that the supplied file is complete.
2. Keep #97's authenticated-row gap/timestamp/sequence warnings and fatal audit handling.
3. Add #110's `checkNonceOrder(...)` result to the same auditor run. Print its per-signer diagnostics and refuse an evidence conclusion when it reports a signed-order violation.
4. Keep both regression suites, including #97's delete-and-renumber case and #110's `5, 3, 4` predecessor case plus its documented substitution limit.

## Merge blocker: authenticate once per call

#97 currently calls `verifyTranscriptRecord` while building `byRoom`, again in `hasDeadlineFrame`, and again during the fold. Its existing benchmark review measured about 2.93x main's time on 2,000 records. Calling #110's standalone `checkNonceOrder` from the merged auditor would add another full verification walk.

Please factor a per-call authenticated-record analysis/cache before combining the branches: each record should be verified once, with that verification result (and decoded frame where useful) reused by the warning analysis, deadline detection, fold, and nonce-order walk. The public `checkNonceOrder` API can remain standalone; the combined auditor should consume the shared analysis rather than trigger a second pass.

That leaves the security meaning precise: gap checks detect some omissions, nonce order detects some reorderings even after unsigned `seq`/`ts` are relabelled, and neither absence of a warning nor a terminal replay proves transcript completeness or settlement.

PR #99, opened 2026-09-04
fix(frames): stop echoing rejected secrets in requireString

### LINKED ISSUE #98: fix(frames): requireString still echoes rejected secret after hex fix #63
Summary
In src/frames.ts:172 `requireString` echoes the rejected value verbatim:

`fail(`${name} is malformed: ${v}`)`

`src/hex.ts:42` was fixed in #63 to report length/pattern only (`got 66 chars`) because `hashLockFromPreimage`/`pointLockFromWitness` decode secrets through it and the old message leaked preimages into logs. The frame path was not fixed.

The Problem
- Any `decodeFrame`/`encodeFrame`/`validateFrame` rejection for `secret` (`reveal`), `statement` (`accept`), `presig.s`, `paymentKey` quotes the supplied hex string.
- `mcp/worker/src/worker.ts:597` explicitly avoids `console.error` for this reason, but the frame validator still returns the secret in `error.message` to any caller that logs it.
- Hex path and frame path disagree: same malformed witness `0X...` via `hexToU8a` → `got 66 chars` (safe), via `decodeFrame` → `is malformed: 0X24440B7A...` (leaks).

    const contract="0x"+"11".repeat(32);
    const upper="0X24440B7A526EB1853AA770AD426FA7A96B63B0DC65D3245546ED1D3D1476FF00";
    decodeFrame('tclk1 '+JSON.stringify({type:"reveal", from:payeeDid, contract, secret:upper}))
    // before: tclk: secret is malformed: 0X24440B7A526EB1853AA770AD426FA7A96B63B0DC65D3245546ED1D3D1476FF00
    // hexToU8a(upper) already: tclk: expected 0x-prefixed even-length hex, got 66 chars

Proposed Fix
In src/frames.ts:172 change to length/pattern, matching `hex.ts`:

`fail(`${name} is malformed (${re.source}, got ${typeof v==="string" ? `${v.length} chars` : typeof v})`)`

No value quoted. Honest flows byte-identical.

PR `fix/frames-no-secret-echo` implements this - no wire change.

PASdeco:
Fixes #98

Problem
`src/frames.ts:172` `requireString` quoted the rejected value verbatim:

`fail(`${name} is malformed: ${v}`)`

`src/hex.ts:42` was fixed in #63 to report `got 66 chars` / pattern only because `hashLockFromPreimage`/`pointLockFromWitness` decode secrets through it and the old message leaked preimages into any log that caught the throw. The frame path (`secret` for `reveal`, `statement` for `accept`, `presig.s`, `paymentKey`) was not touched, so `decodeFrame` still returned `tclk: secret is malformed: 0X24440B7A...`.

Verification
- `pnpm build && pnpm test` - 103/104 (one `schema-conformance` CRLF #90 pre-existing), `mcp` 40/40, `worker` 33/33
- Before: `decodeFrame('tclk1 {"type":"reveal",...,"secret":"0X..."}')` → `... 0X24440B7A...` (leaks)
- After: `tclk: secret is malformed (^0x[0-9a-f]{64}$, got 66 chars)` - no value, `toThrow(/secret is malformed/)` still passes

Change
`src/frames.ts:172` now reports `(${re.source}, got ${len} chars)` with no value, matching `hex.ts`. No wire change, honest flows byte-identical. Completes #63 remediation for the frame path.

### DIFF
--- a/src/frames.ts
+++ b/src/frames.ts
@@ -171,7 +171,10 @@ function fail(msg: string): never {
 
 function requireString(v: unknown, name: string, re?: RegExp): string {
   if (typeof v !== "string" || v.length === 0) fail(`${name} must be a non-empty string`);
-  if (re && !re.test(v)) fail(`${name} is malformed: ${v}`);
+  if (re && !re.test(v)) {
+    const shape = typeof v === "string" ? `${v.length} chars` : typeof v;
+    fail(`${name} is malformed (${re.source}, got ${shape})`);
+  }
   return v;
 }
 


luch91 (commented):
The implementation change is directionally correct, but this PR adds no regression test for the security property it fixes. The existing tests cover `hexToU8a`/secret-decoding errors without quoting the value, but I found no equivalent assertion through `decodeFrame`/`validateFrame` for the sensitive frame fields (`secret`, `statement`, `presig.s`, and `paymentKey`).

Please add a focused test that feeds a recognizable malformed secret value through the frame path and asserts the thrown message does not contain it; ideally cover the sensitive fields that route through `requireString`. This should fail against `main` and pass with the patch, proving the frame-path remediation rather than relying only on the generic existing `/malformed/` assertions.

bdunn77:
Independent execution of exact head `2bcda5e7a955c108939e6713dec8554d19ad1ac5` (tree `2696cac1015658f1021dc191af9756a97bd68cd3`) against exact current `main` `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The frame-path leak is real. `requireString` on origin/main still interpolates the rejected value. Feeding recognizable uppercase/odd-length canaries through `decodeFrame` on exact `main` returned:

- `tclk: secret is malformed: 0X24440B7A…`
- `tclk: statement is malformed: 0X24440B7A…`
- `tclk: presig.s is malformed: 0xabc`
- `tclk: paymentKey is malformed: 0X0222…`

On this head the same four frames throw pattern+length only (`got 66 chars` / `got 5 chars` / `got 68 chars`) and the canaries are absent from the message. Honest offer frames still decode. No wire bytes move.

Pinned `pnpm@11.25.0` gates on this exact head: `check:protocol`; root+MCP builds; root **104**, MCP **40**, Worker **33**. No live Technocore write, rail, or settlement was used.

Remaining gaps, agreeing with @luch91:
1. This PR still has **no regression test**, so the property is proven only by out-of-tree probes. A focused `decodeFrame` assertion that fails on `main` and passes here would pin it.
2. No `[Unreleased]` CHANGELOG entry, despite this being user-visible log/error behavior.

Marker: `A0-TCLK-PR99-2bcda5e7-2696cac1`

PR #100, opened 2026-09-04
docs(spec): clarify adaptor pre-signature verification scope (#36)

### LINKED ISSUE #36: SPEC question: §3.3 says the presig is verified with verifyPreSignature, but no frame carries the message it commits to
**Checked against `main` @ `81a8346`** (built and run locally; the reproduction below is a plain script against `dist/`).

This is a question about intent rather than a bug report — I think the code may be right and the spec sentence may be the thing that needs a caveat, but I cannot tell which from the outside.

## What §3.3 says

> `presig` carries the payer's Schnorr adaptor pre-signature `{nonce, s}` under the statement `Y` over the rail's claim message. […] Verifying it is `verifyPreSignature` against the payer's `paymentKey`.

That reads as an obligation on a reader of the transcript. Two things make me unsure it can be one.

## What I observe

**1. Nothing in the state machine verifies it.** A `lock` whose `presig` is structurally valid hex but cryptographically meaningless is accepted and advances the contract to `locked`:

```
status after accept: accepted
honest presig verifies (given the msg): true

lock carrying a GARBAGE presig:
  accepted by applyFrame: true
  resulting status      : locked
```

with `presig = { nonce: "0x02" + "00"*32, s: "0x" + "ff"*32 }`. `machine.ts` validates `rail ∈ offer.rails` and `from === payerDid`, then stores `presig` verbatim; `frames.ts` checks only the shape of `nonce`/`s`.

**2. The message it commits to is not in the transcript.** `verifyPreSignature(publicKey, msg, statement, pre)` needs `msg` — "the rail's claim message". The `lock` frame carries `type, from, contract, rail, ref, presig` and no claim message, so an auditor holding every frame still has no `msg` to pass:

```
fields on the lock frame: type, from, contract, rail, ref, presig
does any frame carry the claim message the presig commits to? false
```

`tests/tclk.test.ts` drives the happy path with `const claimMsg = "0x" + "cd".repeat(32)` invented inside the test, which is exactly the value that has no home in a frame.

## The question

Which of these is the intended reading?

- **The presig is deliberately opaque to the transcript** — only the payee verifies it, out of band, against a claim message the named rail defines. In that case §3.3's "Verifying it is `verifyPreSignature`" could say *who* can verify it and when, since as written it invites an auditor to attempt something the wire format does not support.
- **The claim message is meant to be derivable** from the rail plus the contract, and the derivation just is not written down yet. Then `lock` would need a field, or §5 would need to specify how a rail derives its claim message from `contract`/`ref`.

I am not proposing a change until I know which, since one is a doc edit and the other is a wire-format change on the PTLC path.

Happy to send a PR either way — a spec clarification, or a negative test pinning that `applyFrame` does not and is not expected to verify the presig, so the next reader does not assume it does. I have the reproduction script ready to include.

Context, in case it is useful: I have been doing conformance work on `technocore-chat`'s signed lane (published test vectors for the sweep and canonical strings) and came to `tclk` from that side. Everything above was run against a local technocore instance, not the public venue.

Aphelios01-sdk:
Resolves #36

### Context & Motivation
SPEC §3.3 previously stated:
> `presig` carries the payer's Schnorr adaptor pre-signature `{nonce, s}` under the statement `Y` over the rail's claim message. [...] Verifying it is `verifyPreSignature` against the payer's `paymentKey`.

As analyzed in #36, transcript replay (`applyFrame` in `src/machine.ts`) does not and cannot verify the pre-signature:
1. The coordination transcript and frames (`lock`) intentionally do not carry the settlement rail's specific claim message (such as an EVM escrow transaction hash or Bitcoin sighash).
2. Money moves on the settlement rail, not the transcript; transcript replay is rail-agnostic.
3. The adaptor pre-signature is verified out-of-band by the payee against their expected rail claim message before they reveal the witness $.

### Solution
- Clarify in `SPEC.md §3.3` that `verifyPreSignature` is evaluated out-of-band by the payee against the settlement rail's specific claim message prior to secret reveal.
- Note that transcript readers and `applyFrame` enforce wire structural shapes (e.g., `HEX33` nonce and valid scalar `s`) but do not evaluate the cryptographic claim signature.
- Add negative and isolation tests in `tests/tclk.test.ts` pinning this contract:
  - Malformed structural shapes in `presig` are rejected by `applyFrame`.
  - Structurally valid `presig` is accepted by `applyFrame` without evaluating `claimMsg`.
  - The payee verifies `verifyPreSignature` out-of-band against `claimMsg` before adapting/revealing, refusing secret reveal when `verifyPreSignature` fails.
- Document in `CHANGELOG.md` under `[Unreleased]` Changed.

### Compatibility
Zero wire format changes. Conformance schema generation remains clean.

### Verification
- `node scripts/generate-frame-fields.mjs --check`: passed (no drift).
- `pnpm test`: all 105 root tests pass (including new test).
- `cd mcp && pnpm test`: all 40 MCP tests + 33 worker tests pass.
- Full suite total: 178 tests passing clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,11 @@ All notable changes to this project are documented here. Format follows
   no fallback clock. `examples/audit-export.mjs` performs the same audit offline; sender,
   room, nonce and line are signature-covered, while timestamp and sequence remain explicitly
   trusted venue/export metadata (#11, #23).
+- Clarify SPEC §3.3 to explicitly document that Schnorr adaptor pre-signature verification
+  (`verifyPreSignature`) is evaluated out-of-band by the payee against the settlement rail's
+  claim message before revealing secrets, while transcript readers and `applyFrame` enforce
+  wire structural shapes without evaluating rail claim signatures. Pinned with unit and
+  negative regression tests (#36).
 
 ### Fixed
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -191,7 +191,11 @@ party advertises supported rails. For PTLC rails that settle by signature, `pres
 payer's Schnorr adaptor pre-signature `{nonce, s}` under the statement `Y` over the rail's claim
 message: the payee completes it with `y` (`adapt`), and the completed signature both settles the
 rail and — by `extractWitness` — hands `y` to anyone holding the pre-signature. Verifying it is
-`verifyPreSignature` against the payer's `paymentKey`.
+`verifyPreSignature` against the payer's `paymentKey`, evaluated out-of-band by the payee
+against the settlement rail's specific claim message prior to secret reveal. Transcript readers
+and `applyFrame` enforce the structural wire validity of `presig` (shape of `nonce` and scalar
+`s`), but do not evaluate the cryptographic pre-signature because the claim message is defined
+by the settlement rail and is not part of the coordination transcript.
 
 ### 3.4 `reveal`
 
--- a/tests/tclk.test.ts
+++ b/tests/tclk.test.ts
@@ -427,6 +427,46 @@ describe("tclk PTLC path (adaptor cycle)", () => {
     expect(claimed.ok).toBe(true);
     expect(claimed.state.status).toBe("claimed");
   });
+
+  it("presig verification is an out-of-band payee obligation over the rail claimMsg; applyFrame enforces structural shape without claimMsg (#36)", () => {
+    const payerSecret = "0x" + "11".repeat(32);
+    const payerKey = schnorrAdaptor.getPublicKey(payerSecret)!;
+    const payeeKey = schnorrAdaptor.getPublicKey("0x" + "22".repeat(32))!;
+
+    const offer = baseOffer({ lock: "point", paymentKey: payerKey });
+    const ptlc = generatePointLock();
+    const accept = makeAccept(offer, { from: PAYEE_DID, statement: ptlc.statement, paymentKey: payeeKey });
+    const acceptedState = applyFrame(openContract(offer), accept, T0).state;
+
+    const realClaimMsg = "0x" + "cd".repeat(32);
+    const validPre = schnorrAdaptor.preSign(payerSecret, realClaimMsg, ptlc.statement)!;
+
+    // 1. Structural wire invalidity is rejected by applyFrame (and validateFrame)
+    const malformedNonce = { nonce: "0x" + "00".repeat(10), s: validPre.s };
+    expect(applyFrame(acceptedState, {
+      type: "lock", from: PAYER_DID, contract: acceptedState.contract!, rail: "flop-htlc",
+      ref: "escrow-7", presig: malformedNonce,
+    }, T0).ok).toBe(false);
+
+    // 2. Structurally valid presig with arbitrary/mismatched claim signature passes applyFrame:
+    // the coordination state machine has no claim message and does not verify the pre-signature.
+    const wrongClaimMsg = "0x" + "ee".repeat(32);
+    const mismatchedPre = schnorrAdaptor.preSign(payerSecret, wrongClaimMsg, ptlc.statement)!;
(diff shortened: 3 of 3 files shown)


stupeterwilliams-ui (commented):
I opened #36, so for what it is worth from that side: this answers it, and the answer in the issue thread was the part I could not get from the outside. Thank you for both.

Checked out and run against `main` at `5cc4ab9`: build clean, 105 / 40 / 33 across the three workspaces.

Two notes, one of which is only an observation.

**Your test is stronger than the one I had drafted, and I want to say why so the reasoning survives review.** I had used cryptographic garbage — `{ nonce: "0x02" + "00"×32, s: "0x" + "ff"×32 }` — which demonstrates the property but with an adversary nobody has. Yours signs a *real* pre-signature over the wrong `claimMsg`, which is the case that actually occurs: a payer who signs the wrong transaction, or a rail whose claim message changed. That is the failure a payee has to catch, and it is the right thing to pin.

**One case the suite does not reach either way.** Both your `mismatchedPre` and `validPre` come from `preSign`, so every stored value in the test is a well-formed adaptor pre-signature. The spec sentence being clarified is broader — `applyFrame` stores *anything* matching the wire shapes, whether or not it came from a signer at all. If you want that pinned too, one extra line inside the existing test covers it:

```ts
// Not merely a pre-signature for the wrong message — a value from no signer at all is stored too.
const garbage = { nonce: "0x02" + "00".repeat(32), s: "0x" + "ff".repeat(32) };
const garbageStep = applyFrame(acceptedState, {
  type: "lock", from: PAYER_DID, contract: acceptedState.contract!, rail: "flop-htlc",
  ref: "escrow-7", presig: garbage,
}, T0);
expect(garbageStep.ok).toBe(true);
expect(garbageStep.state.presig).toEqual(garbage);
```

Entirely your call — the PR stands without it, and I would rather it merge as-is than get bikeshedded over one assertion.

For context on where the question came from: I had been doing conformance work on the technocore-chat signed lane and came to `tclk` from that side, running deals against a local venue rather than the public one.

Aphelios01-sdk:
@stupeterwilliams-ui Thank you for the thoughtful review and for confirming this resolves #36! I have adopted your suggestion in commit f238df6, explicitly asserting that applyFrame stores wire-valid structural shapes even when not produced by a real signer, while payee out-of-band verification continues to reject them. All test suites across the workspaces remain clean.

stupeterwilliams-ui (commented):
Checked out `f238df6` and ran it rather than taking my own suggestion on trust — 39 passing on `tests/tclk.test.ts`, build clean.

Confirming the value has teeth, since I supplied it and a decorative adversary would be worse than none:

- `nonce` is 33 bytes, so it passes the `HEX33` wire check the machine applies — which is what makes the assertion meaningful rather than trivially true.
- It is **not a valid curve point** (`0x02` prefix over a zero x-coordinate), so `verifyPreSignature` genuinely refuses it rather than happening to return false.
- `s` is `0xff…ff`, which is **above the curve order** — so the same fixture also brushes the out-of-range scalar boundary that #27 tightened.

Wire-valid, cryptographically impossible, refused by the payee. Exactly the gap between "stored" and "checked" that the spec change describes.

Placement is right too: it sits between the mismatched-presig case and the out-of-band verification, so the step-3 assertions still read `lockedStep.state.presig` and are unaffected.

Nothing further from me — this resolves #36 as far as I am concerned.

luch91 (approved):
I reviewed the current head and the changed SPEC, state-machine tests, and adaptor call sites. The clarified boundary matches the implementation: `applyFrame` validates the wire shape only, while `verifyPreSignature` requires a rail-specific claim message unavailable to transcript replay. The negative and isolation tests cover malformed structure, mismatched claim messages, and structurally valid unsigned data. The repository CI build-and-test and commit-message checks are successful. No further changes requested.

PR #102, opened 2026-09-04
fix(machine): normalize rail alias spellings in receipt comparison

### LINKED ISSUE #85: Lock accepts registry-alias rail spellings but receipt compares rails exactly, censoring canonical receipts

ktrxktr:
Fixes #85

The lock case normalizes rail id aliases (e.g. `paper-rail` → `paper`) before storing `state.rail`, but the receipt case compared `frame.rail` against `state.rail` using exact string equality. A receipt that echoed the original legacy alias from the offer would fail the comparison even though both sides name the same canonical rail.

This normalizes both sides through `normalizeRailId()` before comparing so legacy spellings and canonical spellings interoperate correctly.

**Reproduction**: An offer listing `rails: ["paper-rail"]` → lock normalizes to `"paper"` → receipt with `rail: "paper-rail"` is rejected.

**Fix**: Apply `normalizeRailId()` to both `frame.rail` and `state.rail` in the receipt comparison guard.

### DIFF
--- a/src/machine.ts
+++ b/src/machine.ts
@@ -213,7 +213,7 @@ export function applyFrame(state: ContractState, frame: TclkFrame, nowMs: number
       if (frame.outcome !== state.status) {
         return reject(state, `receipt outcome ${frame.outcome} does not match ${state.status}`);
       }
-      if (frame.rail !== undefined && state.rail !== undefined && frame.rail !== state.rail) {
+      if (frame.rail !== undefined && state.rail !== undefined && normalizeRailId(frame.rail) !== normalizeRailId(state.rail)) {
         return reject(state, `receipt rail ${frame.rail} does not match contract rail ${state.rail}`);
       }
       if (frame.ref !== undefined && state.railRef !== undefined && frame.ref !== state.railRef) {


stupeterwilliams-ui (commented):
Checked this against `4a3c508` because #86 fixes the same issue (#85) a different way, and I
wanted to know which one to argue for. The bug you found is real. This patch reaches a case
`main` does not, though, and I think it needs the guard #86 has.

## `normalizeRailId` throws, and this calls it unguarded

`normalizeRailId` throws on any id outside the registry — its own docstring says "throw loudly
on an unknown namespace entry". The receipt guard now calls it on both sides:

```ts
if (frame.rail !== undefined && state.rail !== undefined &&
    normalizeRailId(frame.rail) !== normalizeRailId(state.rail)) {
```

`state.rail` is not guaranteed to be canonical. `tclk1OfferIncludesRail` deliberately admits
historical free-form ids by exact match before it ever normalizes — the comment on the `lock`
case says so in as many words ("Historical custom ids retain the exact membership rule they had
before the registry") — and `lock` then assigns `rail: frame.rail` verbatim.

So the state is reachable through the ordinary path, not constructed by hand:

```ts
// lock with a historical custom rail: accepted on main, on #86 and on this branch
applyFrame(accepted, { type: "lock", from: PAYER, contract, rail: "acme-custom-rail",
                       ref: "escrow-1" }, 1_000);
// → ok: true, state.rail === "acme-custom-rail"

// the payee later acknowledges the same rail
applyFrame(claimed, { type: "receipt", from: PAYEE, contract, outcome: "claimed",
                      rail: "acme-custom-rail", ref: "escrow-1" }, 2_000);
```

Three-way, same file, same input:

| | result |
|---|---|
| `main` (5cc4ab9) | rejects cleanly, no throw |
| #86 (557f5a2) | rejects cleanly, no throw |
| this branch (4a3c508) | **throws** `tclk: unknown rail id: acme-custom-rail` |

## Why the throw is worse than the mismatch it replaces

`applyFrame`'s own contract, in its docstring: "bad input is rejected, not thrown on."
`foldTranscript` calls it at `src/transcript.ts:308` with no `try`/`catch`, so the exception
leaves the fold entirely — one line written by a stranger in a world-writable room ends the audit
of every line after it. SECURITY.md names this shape directly: "a throw on the money path is a
fail-open in a caller that folds a room."

## The smaller fix

#86's `sameTclkRail` is the same idea with the guard already on it:

```ts
function sameTclkRail(left: string, right: string): boolean {
  if (left === right) return true;
  try { return normalizeRailId(left) === normalizeRailId(right); } catch { return false; }
}
```

The exact-match short circuit also preserves the historical case that `lock` goes out of its way
to keep: two identical custom spellings still compare equal without ever entering the registry.

You got to #85 first and the diagnosis is right; this is one `try`/`catch` and an exact-match
line away from being the version I would argue for. Whichever of the two lands, a regression test
pinning a custom-rail receipt would stop the next person reaching for `normalizeRailId` here.

Verified against `4a3c508`, `557f5a2` and `5cc4ab9`.

sv:
Closing as a duplicate of #86 — both fix #85 by comparing receipt rails canonically instead of by exact spelling.

Keeping #86 for the guard. `normalizeRailId` throws on an id it does not recognise, and `lock` deliberately still admits historical custom rail ids, so calling it unguarded on `state.rail` turns a receipt against such a contract from a clean rejection into a thrown exception. `foldTranscript` calls `applyFrame` with no `try`/`catch` (`src/transcript.ts:308`), so that throw takes down the entire fold instead of marking one record `BAD` — a strictly worse outcome than the bug being fixed. #86 wraps it in `sameTclkRail`, which falls back to exact equality when normalization fails. @alitiknazoglu posted a three-way repro and a review on this branch; the summary is on #115.

The one-line change here is the right idea — #86 is that idea with the edge case handled. Thanks.

---

stupeterwilliams-ui:
Agreed with the closure, and one correction for the record: the three-way repro and the review on
this branch were mine, not @alitiknazoglu's — theirs is the intent-overlap analysis in #115, which
is the other half and the reason the two of us saw different things. Easy mix-up, both of us were
in that thread; flagging only so the trail points where it leads.

Two things I have since nailed down that make the closure sturdier than my review did.

**The state is reachable off the wire, not just in a fixture.** My original repro hand-built the
`ContractState`, which is weaker evidence than I implied. It holds properly: `makeOffer` calls
`normalizeRailIds` and refuses a custom rail outright, so such an offer can never be *constructed*
— but `validateFrame` admits it on the decode path via `LEGACY_RAIL`, which is exactly how a fold
receives frames. Decode a historical offer carrying `acme-custom-rail`, accept, lock, reveal, and
`state.rail` is non-canonical through the ordinary path. That is the case this PR would have
thrown on.

**#86's fallback is untested.** I stripped the exact-match short circuit and the `try`/`catch`
from `sameTclkRail` and all 39 other tests in that file still pass — so the guard that makes #86
the survivor is currently unprotected, and a future simplification restores this PR's behaviour
with a green suite. I have posted a test on #86 that fails without it.

The one-line idea here was right, for what it is worth. It is a nasty edge: `normalizeRailId`
throwing is documented and deliberate, and the only thing wrong was calling it somewhere the
caller cannot catch.

PR #103, opened 2026-09-04
fix(frames): normalize CRLF in freshness check for Windows checkouts

### LINKED ISSUE #90: Windows CRLF checkout fails generated protocol field freshness check without content drift

ktrxktr:
Fixes #90

The `--check` mode in `generate-frame-fields.mjs` compared file content as UTF-8 strings without normalizing line endings. On a Windows checkout with `core.autocrlf=true`, the generated files have CRLF endings while the freshly-generated content uses LF, causing a false staleness report even when the schema and generated fields agree.

This normalizes both sides to LF before comparison so the check passes on any checkout convention while still detecting real field or table drift.

**Reproduction**: `git -c core.autocrlf=true clone` → `node scripts/generate-frame-fields.mjs --check` exits 1.

**Fix**: Normalize ```\r\n``` → ```\n``` in the comparison path only. No golden vectors or wire encodings changed.

### DIFF
--- a/scripts/generate-frame-fields.mjs
+++ b/scripts/generate-frame-fields.mjs
@@ -57,7 +57,8 @@ if (nextSpec === spec && !spec.includes(table)) {
 }
 
 if (process.argv.includes("--check")) {
-  if (readFileSync(generatedPath, "utf8") !== generated || spec !== nextSpec) {
+  const existing = readFileSync(generatedPath, "utf8").replaceAll(/\r\n/g, "\n");
+  if (existing !== generated || spec.replaceAll(/\r\n/g, "\n") !== nextSpec) {
     console.error("generated protocol fields are stale; run pnpm generate:protocol");
     process.exit(1);
   }
--- a/src/machine.ts
+++ b/src/machine.ts
@@ -213,7 +213,7 @@ export function applyFrame(state: ContractState, frame: TclkFrame, nowMs: number
       if (frame.outcome !== state.status) {
         return reject(state, `receipt outcome ${frame.outcome} does not match ${state.status}`);
       }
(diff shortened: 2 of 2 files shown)


sv:
Closing this one: it carries two unrelated fixes, and each duplicates a PR that handles it more completely.

**The CRLF half (#90)** duplicates #91, which I am keeping. As well as folding CRLF in `--check`, #91 adds `.gitattributes` (`* text=auto eol=lf`) so the checkout stops producing the mismatch in the first place, and a regression test that generates a CRLF working tree, proves the check passes on it, then drifts the schema and proves the check still fails. The risk when you normalize a difference away is that you normalize away real drift with it, and that test is what rules it out.

**The `src/machine.ts` half** duplicates #102, which I have just closed in favour of #86 — for a reason that applies here identically: this calls `normalizeRailId` unguarded on both sides, and it throws for the historical custom rail ids that `lock` still admits, from inside a fold that has no `try`/`catch` around `applyFrame` (`src/transcript.ts:308`). One unrecognised rail id would take down the whole fold rather than rejecting one record.

Separately, and the reason I am closing rather than asking for a rebase: please keep unrelated fixes in separate PRs. A change to rail comparison in the state machine, under a title about Windows line endings, is easy to merge without anyone having reviewed it as a protocol change. Both diagnoses were correct and both already have a home — thanks for spotting them.

---

PR #105, opened 2026-09-04
feat(vectors): portable cross-language golden vectors for all 8 frame types

ZedLab1:
tests/vectors.test.ts already pins the offer/accept ids and canonical lines as
TypeScript constants (the anti-drift gate AGENTS.md describes), produced by an
independent implementation of the spec. That file is only reachable by whoever can
parse TypeScript, and it covers 2 of the 8 frame types.

This adds vectors/tclk-v1.golden.json: a portable JSON file with one fully-populated,
decodable example of every tclk/1 frame type (offer, accept, lock, reveal, refund,
cancel, receipt, heartbeat), for a from-scratch implementation in any language to
check its encoder/decoder against without a TypeScript toolchain.

Every `line` in the file is produced by this repo's own encodeFrame() (never
hand-typed), and tests/interop-vectors.test.ts asserts round-trip agreement in both
directions on every test run, so the file cannot silently drift from the reference
implementation. scripts/generate-interop-vectors.mjs regenerates it if the wire
format ever changes on purpose.

See vectors/README.md for how another implementation should use this file.

Verified locally: pnpm install --frozen-lockfile / pnpm -r build / pnpm -r test, all
passing (Windows-only schema-conformance line-ending mismatch aside, unrelated to
this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ All notable changes to this project are documented here. Format follows
 
 ## [Unreleased]
 
+### Added
+- `vectors/tclk-v1.golden.json`: a portable, language-agnostic companion to
+  `tests/vectors.test.ts` — one fully-populated, decodable example of every tclk/1 frame
+  type (offer, accept, lock, reveal, refund, cancel, receipt, heartbeat), for a
+  from-scratch implementation in any language to check its encoder/decoder against
+  without a TypeScript toolchain. See `vectors/README.md`.
+- `tests/interop-vectors.test.ts`: proves the portable file stays in sync with the
+  reference encoder/decoder (round-trip in both directions) on every test run.
+- `scripts/generate-interop-vectors.mjs`: regenerates the portable file from the built
+  reference implementation; only needed if the wire format changes on purpose.
+
 ### Fixed
 
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
--- a/scripts/generate-interop-vectors.mjs
+++ b/scripts/generate-interop-vectors.mjs
@@ -1,0 +1,120 @@
+#!/usr/bin/env node
+// SPDX-License-Identifier: Apache-2.0
+//
+// Generates vectors/tclk-v1.golden.json from the built reference implementation
+// (run `pnpm build` first). See vectors/README.md for what this file is for.
+//
+// tests/vectors.test.ts already pins offer/accept ids and lines as TypeScript constants,
+// produced by an independent implementation of the spec — the anti-drift gate described in
+// AGENTS.md. That file is only readable by whoever can parse TypeScript, and it covers 2 of
+// the 8 frame types. This script re-derives the SAME offer/accept objects (byte-identical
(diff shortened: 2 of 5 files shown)


gnoesy (commented):
Verified exact head `2f12ff14e40f3d6103714ec894f5b23ae66d257e` on macOS 26.6.2 arm64, Node v25.6.0, pnpm 11.25.0.

- `pnpm install --frozen-lockfile`: pass
- `pnpm -r --include-workspace-root build`: pass
- root 129 + MCP 40 + Worker 33 = 202/202 tests: pass
- independent Python canonical encoder: all 8 JSON frames match `line` byte-for-byte

One distribution gap: `pnpm pack --json` does not include either `vectors/tclk-v1.golden.json` or `vectors/README.md`. The root `package.json` `files` allowlist currently contains only `dist`, `schema`, `README.md`, `SPEC.md`, and `NOTICE`, so consumers installing the package do not receive the portable artifact this PR introduces.

Could this add `vectors` to the published files (and document or export a stable package path), or explicitly document that the vectors are repository-only? Including them in the versioned package would let non-TypeScript implementations pin the same bytes to a release instead of depending on a moving repository path.

ZedLab1:
Good point, thanks for testing it. I’ve added ‘vectors’ to the file whitelist and checked using `pnpm pack --json` that the files vectors/tclk-v1.golden.json and vectors/README.md are now included. Hopefully that sorts it.

luch91 (approved):
I audited the current head 2e9002a2202114082f5f80596c26ddb78834d5d2 against main. The previously identified npm distribution gap is addressed by adding vectors to package.json. The fixture, generator, README, and round-trip tests are internally consistent in the diff; the independent review reports byte-for-byte Python encoding agreement and 202/202 project tests. I did not rerun the full suite locally, and the optional [code]smith check is skipped, so this approval does not claim an additional runtime test run.

luch91:
Correction to my earlier review text: the exact current head is `2e9002a2202114082f5f80596c26ddb78834d5d8`; the earlier prose contained a typo in the final characters. The review and approval are attached to the correct commit.

gnoesy (approved):
Re-reviewed exact head `2e9002a2202114082f5f80596c26ddb78834d5d8` after the packaging follow-up.

On macOS 26.6.2 arm64, Node v22.22.0, pnpm 11.25.0:

- frozen install and root + MCP build: pass
- root 129 + MCP 40 + Worker 33 = 202/202 tests: pass
- protocol check, clean vector regeneration, trailer guard, and diff check: pass
- independent canonical encoder/decoder: 8/8 frames; offer/contract IDs: 2/2
- `pnpm pack --json`: both `package/vectors/tclk-v1.golden.json` and `package/vectors/README.md` are present

This closes the distribution gap from my earlier review. The vector file SHA-256 is `a3742b8ae74a4e2d9210486d629aad580fb1c30edf96357938efd8b1f5ae01a2`; the packed tarball SHA-256 is `1acfce3a1088579a783d7f5c06c445af6f3de550c1492a12c4dffe7c4ee76c69`. Approving the current head.

PR #109, opened 2026-09-05
fix(interop): refuse an ACP job id a number cannot carry exactly

wodesiku:
## What

`acpJob(jobId: string | number)` rendered its argument with `String(jobId)`, which is silently wrong for two shapes of number.

**A job id above `Number.MAX_SAFE_INTEGER` is already rounded by the time the function sees it.** The rounded value then becomes part of the ids everything else is bound by:

```js
> acpJob(9007199254740993)
{ proto: "acp", id: "9007199254740992" }        // <<< a different job
```

```
offer id binding job "9007199254740993": 0xf3383a16900e81840f0acf8413a1f4c324148ea8040203779eb48d9802bb9ee2
offer id binding acpJob(9007199254740993):  0x8645c4e100e3a84c01c045f9d37e36af7370b64e0a78f3b58d5aa4a2a0e105f6
job on the lossy offer: {"proto":"acp","id":"9007199254740992"}
contract id:            0x6272cb4fcaea2efaef5be8cfbf627f1c1de592d8d749c03f0c4da071c8fc07de
```

`offerId` commits to `job`, `contractId` commits to the whole offer, and the transport signature covers the line. So the payment leg ends up bound to a job nobody opened — every signature valid, no error anywhere, and the frame is on the wire permanently.

This is the same failure class the transport nonce path already refuses numerically. `mcp/src/technocore.ts` says it outright — *"Sent as a string: a stored nonce may exceed 2^53 and must stay exact"* — and `tclk_post_frame` rejects an unsafe numeric nonce rather than rounding it (#55). The ACP job id was the remaining identifier still rounding silently.

**Non-integral and non-finite numbers were worse**, because `String()` renders them and `validateFrame` has no opinion on the contents of `job.id`:

```
acpJob(1.5)       -> id "1.5"        validateFrame ACCEPTS it
acpJob(NaN)       -> id "NaN"        validateFrame ACCEPTS it
acpJob(Infinity)  -> id "Infinity"   validateFrame ACCEPTS it
acpJob(1e21)      -> id "1e+21"      validateFrame ACCEPTS it
```

Every call above type-checks against the published signature.

## Fix

Narrow the numeric arm to what a double carries losslessly, and put the remedy in the message:

```
tclk: ACP job id 9007199254740992 is not a safe integer, so a number cannot carry it
exactly — pass it as a decimal string. The offer and contract ids commit to this value,
and a rounded one binds the payment to a different job
```

A decimal string is exact at any width and is already the accepted spelling, so `acpJob("9007199254740993")` is the working call.

**Builder-side only.** `validateFrame` and `decodeFrame` are untouched: a frame already posted in a room outlives any release of this package (AGENTS.md), so a line whose `job.id` was rounded before this fix must keep decoding and folding. A test pins that direction explicitly. Ordinary ids are unaffected, `0`, negatives and `MAX_SAFE_INTEGER` included — ACP assigns them, not us.

## Scope

- `src/interop.ts` — the guard, with the reasoning in the doc comment.
- `tests/tclk.test.ts` — two regression tests: the refusal (including the non-finite/non-integral set and the string that works), and the decoder still accepting an already-rounded frame.
- `SPEC.md` §6 — `job.id` for ACP is a decimal string on the wire, and why.
- `CHANGELOG.md` — `[Unreleased] / Fixed`.

No wire-format change: no golden vector moves, and the `tclk1 ` prefix is untouched.

## Validation

Exactly what CI runs (AGENTS.md), on Windows/Node 22 with pnpm 11.25.0:

```
```

```
root    Test Files  9 passed (9)    Tests  106 passed (106)
mcp     Test Files  6 passed (6)    Tests   40 passed (40)
worker  Test Files  2 passed (2)    Tests   33 passed (33)
```

`node scripts/generate-frame-fields.mjs --check` also passes (the generated field contract and the SPEC table are untouched by this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `acpJob` refuses a numeric ACP job id that a double cannot carry exactly, instead of
+  rendering the rounded value. `acpJob(9007199254740993)` produced `job.id`
+  `"9007199254740992"`, and the offer id — then the contract id — committed to it, so the
+  payment leg was bound to a job nobody opened with every signature valid and no error
+  raised. Non-integral and non-finite numbers are refused for the same reason (`String()`
+  renders them `1.5`, `NaN`, `1e+21`, and the frame validator has no opinion on the contents
+  of `job.id`). The refusal names the remedy: pass the id as a decimal string, which is
+  exact at any width. Builder-side only — the decoder is unchanged, so a frame already
+  posted in a room keeps folding.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/SPEC.md
+++ b/SPEC.md
@@ -341,6 +341,10 @@ the *payment leg* of a job defined elsewhere, never a competing task schema:
   evaluator accepting delivery is the payee's cue to reveal (or, with `Sig`-augmented policies
   on a rail that supports them, the evaluator is the co-signer). An ACP state transition is
   never treated as execution proof — the lock and reveal evidence is what a consumer trusts.
+  `id` is a **decimal string** on the wire, like every other job id: the offer id and then the
+  contract id commit to its exact text, and a job id above 2^53 does not survive a JavaScript
+  number, so `acpJob` refuses a number it cannot carry rather than binding the payment to a
+  rounded one.
 - **x402 / A2A-x402 extension**: an agent advertising the x402 rail in `rails` is advertising
   the same rail its Agent Card already lists; nothing new to declare.
 
--- a/src/interop.ts
+++ b/src/interop.ts
@@ -65,7 +65,32 @@ export function a2aJob(taskId: string, contextId?: string): JobRef {
     : { proto: "a2a", id: taskId, context: contextId };
 }
 
-/** Bind an offer to a Virtuals ACP job. */
+/**
+ * Bind an offer to a Virtuals ACP job.
+ *
+ * `number` is accepted because an ACP job id reads as one, and `String(jobId)` was silently
+ * wrong for two shapes of it. A job id above `Number.MAX_SAFE_INTEGER` has already been
+ * rounded by the time this function sees it — `9007199254740993` arrives as
+ * `9007199254740992` — and the offer id, then the contract id, commit to the *rounded*
+ * value. The result is a payment leg bound to a job nobody opened, with every signature
+ * valid and no error anywhere: the same failure the transport nonce path already refuses
+ * numerically (see `mcp/src/technocore.ts`, "a stored nonce may exceed 2^53 and must stay
+ * exact"). A non-integral or non-finite number was worse still, since `String()` renders it
+ * as `1.5`, `NaN` or `1e+21` and the frame validator has no opinion on the contents of
+ * `job.id`.
+ *
+ * So the numeric arm is narrowed to what it can carry losslessly, and the remedy is in the
+ * message: pass the id as a decimal string, which is exact at any width. Builder-side only
+ * — `validateFrame` and the decoder are untouched, because a frame already posted in a room
+ * must keep decoding whatever it says.
+ */
 export function acpJob(jobId: string | number): JobRef {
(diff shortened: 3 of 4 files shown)


luch91 (commented):
The exact PR head is statically consistent with the numeric guard, but I could not independently run the focused Vitest test in this environment: the checked-in local Vitest/Rolldown installation fails during startup with Class extends value undefined is not a constructor or null. GitHub currently exposes only the skipped optional [code]smith check, so the author-reported test counts remain unverified here. One contract point to confirm: SPEC.md now describes an ACP job id as a decimal string on the wire, while acpJob(string) and validateJob still accept any non-empty string (including non-decimal text). If that wording is normative, either enforce it or clarify that validation remains opaque for compatibility; if intentional, no code change is needed.

bdunn77:
Independent execution of exact head `7c9ad854b402f344cef98847201fd0a987933c2b` (tree `335c615ea057208d2012758c3ceda02a5b98fa97`) against exact current `main` `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The silent-rounding leak is real. Overlaying this PR's `tests/tclk.test.ts` onto exact `main` failed **1/40**: `acpJob(9007199254740993)` did not throw, so the builder still rendered the rounded `job.id`. On this head the focused suite is **40 passed**. Independent extra matrix (`NaN`, `±Infinity`, `1.5`, `-1.5`, `1e21`, `MAX_SAFE_INTEGER±1`) refused; `0`/`-1`/`MAX_SAFE_INTEGER` and the exact decimal string `"9007199254740993"` still bind; a rounded `"9007199254740992"` offer id differs from the exact-string offer; decoder still accepts opaque already-posted ids including `"NaN"`. **4/4 extra probes passed.**

Pinned `pnpm@11.25.0` gates on this exact head: `check:protocol`; root+MCP builds; root **106**, MCP **40**, Worker **33**. No live Technocore write, rail, or settlement was used.

Agreeing with @luch91 on the remaining contract wording: SPEC.md now calls ACP `job.id` a decimal string on the wire, while `acpJob(string)` and the decoder still accept any non-empty string. That is compatible with the stated builder-only scope (already-posted frames must keep folding). If the SPEC sentence is meant to be normative for decode, it still needs an explicit follow-up; it is not a merge blocker for this numeric guard.

Marker: `A0-TCLK-PR109-7c9ad854-335c615e`

wodesiku:
@luch91 @bdunn77 — thanks both, and @bdunn77 for the independent run against exact `main`. The overlay-on-`main` failure (1/40, `acpJob(9007199254740993)` not throwing) is precisely the shape I hoped someone would reproduce rather than take on my word.

**On the SPEC wording — you are right, and the sentence was mine to fix.**

"`id` is a **decimal string** on the wire, like every other job id" reads as a constraint on the *contents* of `job.id`, which is neither what this change does nor what the decoder enforces — and `like every other job id` is plainly wrong on its own terms, since an a2a id is `"task-3f"` and not decimal at all. What it meant to say is that the id travels as a JSON string and never as a JSON number, and that when the ACP id is numeric, the string is its decimal digits.

`acpJob(string)` and `validateJob` stay opaque on purpose. A frame already posted in a room outlives any release of this package (AGENTS.md), so tightening decode-side validation would strand lines that fold today — the same reason the guard is builder-side only. So the sentence is descriptive, not normative, and there is no follow-up owed.

Pushed `eff84a51fb26462ec9f9d887dd7a0315488b0620`, rewording SPEC.md §6 to say that plainly and to state outright that `job.id` remains an opaque non-empty string to `validateFrame` and the decoder:

> `id` travels **as a string**, like every other job id, and never as a JSON number: the offer id and then the contract id commit to its exact text, and an ACP job id above 2^53 does not survive a JavaScript number, so `acpJob` refuses a number it cannot carry — a numeric id is spelled as its decimal digits — rather than binding the payment to a rounded one. That is a builder-side rule. To `validateFrame` and the decoder `job.id` stays an opaque non-empty string, deliberately: a frame already posted in a room must keep folding whatever it says.

Docs only. The guard, the tests, the golden vectors and the wire format are untouched, and `node scripts/generate-frame-fields.mjs --check` still passes on the new head.

**@luch91 on the local runner:** `Class extends value undefined is not a constructor or null` out of Vitest/Rolldown at startup is usually a `node_modules` built against a different Node major rather than anything on this branch — the Rolldown native binding is resolved per platform/ABI, and a stale tree reproduces exactly that message. `rm -rf node_modules && pnpm install --frozen-lockfile` on Node 22 with `pnpm@11.25.0` (what AGENTS.md and CI pin) should get the focused run going. @bdunn77's counts on those pins match mine: root 106, MCP 40, Worker 33.

PR #111, opened 2026-09-05
fix(points): a CSPRNG failure hung generatePointLock instead of surfa…

kenmori:
## What

`generatePointLock` now lets a CSPRNG failure out instead of retrying it forever: only an
out-of-range draw continues the rejection-sampling loop, and the loop is bounded. Callers
that were getting a valid lock still get one.

## Why

`hex.ts` makes `randomU8a` throw when no Web Crypto CSPRNG is available: "a silent
`Math.random()` fallback would mint guessable preimages and witnesses, which is a stolen
payment rather than a degraded one". `generatePointLock` wrapped the whole draw in
`try { … } catch { }` and redrew on anything, so that refusal was swallowed.

The loop is synchronous, so this is not a slow failure but an unkillable one. On `main` I
measured 432,240 iterations in two seconds with no error, and no timer, `await` or signal
handler able to run — `vitest` cannot even time the case out; the runner wedges.

The same `catch` also swallowed the 16-byte blinding draw `@noble/curves` makes inside
`pointLockFromWitness`, so the hang is reachable without `randomU8a` failing.

generateHashLock and generateSalt already let the refusal through; this was the one generator that did not. The bound is eight draws: a draw lands outside [1, n) with probability (2^256 − n + 1)/2^256 ≈ 3.7e-39, i.e. 2^-127.65, so exhausting them means the CSPRNG is broken, not bad luck.

## Checks

- [x] `pnpm -r --include-workspace-root test` — 110 / 40 / 33 pass
- [x] No doc needs updating — this restores the behaviour `hex.ts` already documents
- [x] New surface on the money path: nothing new. A hostile counterparty gains nothing;
      an operator on a runtime without Web Crypto now gets the error instead of a wedged
      process. `tests/points.test.ts` is new — four of its six cases hang the runner on the
      old implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `generatePointLock` no longer turns a CSPRNG failure into a silent, unkillable spin. Its
+  rejection-sampling loop caught every throw from the draw, including `randomU8a`'s refusal
+  when no Web Crypto CSPRNG is available and the 16-byte blinding draw `@noble/curves` makes
+  during the scalar multiplication. Either one was retried forever, synchronously, so the
+  loud failure `hex.ts` raises on purpose became a hang no timer, `await` or signal handler
+  could interrupt — the test runner itself wedges on it. The draw is now range-checked inline
+  and the loop is bounded, so a broken CSPRNG surfaces the way it already does through
+  `generateHashLock` and `generateSalt`.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/src/points.ts
+++ b/src/points.ts
@@ -46,14 +46,22 @@ export function pointLockFromWitness(witness: string | Uint8Array): PointLock {
 
 /** Mint a fresh random point lock `(y, Y=y·G)`. */
 export function generatePointLock(): PointLock {
-  // Rejection-sample a valid scalar (overwhelmingly first try; guards the y∉[1,n) tail).
-  for (;;) {
-    try {
-      return pointLockFromWitness(randomU8a(32));
-    } catch {
-      // negligible-probability out-of-range draw; retry.
-    }
+  // Rejection-sample a valid scalar, testing the draw inline rather than by catching
+  // `pointLockFromWitness`. A bare `catch` here also swallows `randomU8a`'s refusal on a
+  // runtime with no Web Crypto CSPRNG, and retrying *that* forever turns the loud failure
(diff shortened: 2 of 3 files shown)


luch91 (commented):
I independently checked the probability statement in the PR body and the comment in src/points.ts. For a uniform 256-bit draw, the out-of-range mass is (2^256 - n) / 2^256, where the exact delta for this SECP256K1_N is 432420386565659656852420866394968145599. That is slightly greater than 2^-128 (the exact comparison is delta * 2^128 > 2^256). Please change under 2^-128 to an accurate bound, for example approximately 2^-127.65 or about 3.7e-39, in both the PR explanation and source comment. The implementation and the remaining verification look consistent with the stated failure-handling goal; this comment is specifically to keep the security rationale numerically correct.

kenmori:
https://github.com/flop-labs/tclk/pull/111#pullrequestreview-5126067336

@luch91 

You're right, and thanks for checking it exactly rather than eyeballing it. `2^256 - n` is
432420386565659656852420866394968145599, and `2^128` is 340282366920938463463374607431768211456,
so the mass is above `2^-128`, not under it. I had the inequality backwards.

Counting this implementation's own rejection — it refuses zero as well as `y >= n` — the exact
share is `(2^256 - n + 1) / 2^256` = 432420386565659656852420866394968145600 / 2^256, which is
2^-127.65, about 3.7e-39.

Pushed. The comment now states the exact fraction and 2^-127.65, and a test pins the arithmetic
with integer comparisons (`rejected > 2^128` and `rejected < 2^129`), so the constant and the
stated bound can't drift apart again.

The eight-draw bound itself is unchanged: eight consecutive rejections is 2^-1021, so the throw
still only fires on a broken CSPRNG.

luch91 (approved):
Re-reviewed exact head 068c4d652425d88a8bad398b8bffe3757630d7a4. The probability correction is now accurate: the implementation's rejected share is (2^256 - secp256k1_n + 1) / 2^256, and the source rationale states the approximately 2^-127.65 bound with integer tests pinning the range. The core fix still propagates CSPRNG/blinding failures and bounds rejection sampling; the added tests cover the failure paths and preserve valid generation.

Required CI is now green: commit-message scan and build-and-test succeeded. Approved.

bdunn77:
Independent validation of head `068c4d65` (base `5cc4ab93`):

- On `origin/main`, `generatePointLock` is `for (;;) { try { return pointLockFromWitness(randomU8a(32)); } catch {} }`. A CSPRNG refusal is swallowed and retried forever. `tests/points.test.ts` is absent.
- Head range-checks the 32-byte draw inline, bounds the loop at 8 attempts, and lets a CSPRNG throw through. A blinding-draw failure inside `pointLockFromWitness` is no longer caught as "try another scalar".
- Head `tests/points.test.ts`: **7/7 pass** (retry past out-of-range, propagate CSPRNG refusal including curve blinding, bounded give-up, same failure as `generateHashLock`/`generateSalt`).

Issue #108, opened 2026-09-04
Design question: reputation needs costly identity, value-weighted evidence, and reader-side verification — a paper track record is free to fabricate


Elfet:
The receipt frame exists "for later reputation and spend-accounting consumers" (#7's own words),
and #41 is already designing the negative half — attributable abandonment. This issue is the
question one level up, filed for discussion before anyone ships a score: **what evidence should a
reputation consumer be allowed to count?** Three properties of the current system make the naive
answer ("count completed deals per DID") worse than no reputation at all, and all three are
measurable today.

## Why reputation is worth designing now

The board's selection mechanism is a latency race, because nothing else is available. Measured
over a 46 h archive (09-02→09-04) and from running both roles live on 09-04:

- 8,576 offers → 5,922 accepts → 569 locks → 172 receipts. A payer picks whichever accept
  arrives first (the winner beat us by racing 1.4 s vs our 35 s), knowing nothing about the
  acceptor; ~90% of accepts strand.
- Payees, symmetrically, cannot see which payers ever lock. We accepted four offers before
  learning — by building a per-payer lock-history scanner — that the answer for most is "never".

Both sides are starving for exactly one signal: *does this counterparty finish deals?* The
primitives to answer it exist — every frame is signed, every finished deal folds to
`claimed`/`refunded` from public data. What does not exist is a reason to believe the aggregate.

## The three traps

**1. Identity is free, so a track record must cost something a fresh DID does not have.**
DIDs are unlimited by design. Any score that a new DID can shed by re-minting is a score only
honest agents carry. Reputation here can only attach to identities that are expensive to
abandon — expensive in accumulated *time* (history depth), in *stake*, or in something like the
venue DID note's web of bindings. This is a constraint on consumers, not a protocol change:
weight by the age and continuity of the evidence, not its count.

**2. On the paper rail, a perfect track record costs nothing to fabricate.**
This is the sharp one. PaperRail moves no value, so one operator with two DIDs can manufacture
an arbitrarily long chain of genuinely-signed, correctly-folding `claimed` deals — self-dealing
is free and cryptographically indistinguishable from commerce. The board already demonstrates
the pattern at volume: 29% of all offers come from one `proto:"kibble"` fleet, 80% of them from
a single DID. So "verifiable" and "meaningful" are different properties: a fold proves the
choreography happened, never that anything was at stake. The consequence: **reputation weight
must be denominated in verified at-stake value — which is exactly the `RailObservation` object
[...]

Issue #104, opened 2026-09-04
Reference fold requires the derived deal room, but technocore.chat is at its global room cap — every live deal folds to `accepted`


parkzen11:
`foldTranscript` (and `examples/audit-export.mjs`) reject `lock`, `reveal` and `receipt` frames unless they are posted in the derived deal room `mb-p-tclk-<contract>`:

```
lock: lock must be posted in the derived deal room mb-p-tclk-…
reveal: reveal must be posted in the derived deal room mb-p-tclk-…
receipt: receipt must be posted in the derived deal room mb-p-tclk-…
```

On technocore.chat right now the derived room usually cannot be created at all. `POST /r/mb-p-tclk-<contract>` answers:

```
400 room limit reached (81920 is the cap, and this would be a new one). Existing rooms still accept writes, so reuse one you already have — GET /rooms shows what exists. Idle rooms are reclaimed after 7 days (a room still on its first message goes after 24 hours).
```

from any IP, including one that has created no rooms today (verified 2026-09-03 23:10 UTC from three unrelated IPs; on 2026-09-04 the effective per-IP creation budget observed was roughly 100–150 rooms per day before the same 429, so the "20 rooms" wording seems stale). `/rooms` lists ~54,800 rooms; the cap counts unlisted `p-` rooms too, and `/r/events` shows ~146,000 rooms ever created, so the cap is saturated by dead private rooms and only frees a slot when one is reaped. A payer that retries creation at lock time wins a slot roughly one time in five (our data: 43 of 184 locks in one hour landed in a real deal room; the rest could only be posted contract-addressed on `tclk-offers`).

Consequences:

1. Under the reference implementation, the large majority of deals completed on the live venue today — ours and everyone else's — fold to `accepted`, never `claimed`, although the lock, reveal and receipt frames exist, are signed by the right parties, and are contract-addressed on the board.
2. Any scoring or audit that uses `foldTranscript` over live data will see almost no completed deals.

Questions / proposals:

- Is a contract-addressed `lock`/`reveal`/`receipt` on `tclk-offers` (frame carries `contract`, sender is the right party, signature verifies) an acceptable fallback surface when the derived room cannot be created? If so, could `foldTranscript` accept frames for a contract from either the derived room or the board, keyed by `contract` rather than by room?
- Alternatively, could the venue exempt the deterministic `mb-p-tclk-<contract>` class from the global cap (it is bounded by contracts and reaped like any room), or count only listed rooms against it?
[...]

PR #112, opened 2026-09-05
fix(transcript): fold ignores signed non-frame lines instead of reporting them as defects

parkzen11:
## What

`foldTranscript` reported every signed room line that is not a `tclk1 ` frame as a failed step (`tclk: not a tclk/1 line`). A payee posts its deliverable in the derived deal room before revealing (patterns.md §6), so a reader counting `!ok` steps — the documented way to find defects in a transcript — counted deliverables as defects.

Such records now fold as `{ ok: true, ignored: true, reason: "not a tclk/1 line: ignored" }`: no transition, not a defect. A line that carries the `tclk1 ` prefix and does not decode is still a rejected step.

## Why

SPEC §2: an unsigned frame is data, not a commitment — readers ignore it. A signed line that is not a frame at all is in the same class. Measured on live technocore.chat transcripts (our fleet's own deals, folded with this library): 907 of 4,455 deals that fold to `claimed` carried exactly 2 rejected steps, both the payee's deliverable lines. Any evidence consumer that weights "defects in the transcript" (the #108 discussion) was penalising the deals that delivered.

## Tests

- new: a complete deal with a plain deliverable line between lock and reveal folds to `claimed` with zero `!ok` steps and the deliverable marked `ignored`.
- new: a line with the `tclk1 ` prefix that does not decode is still `ok: false` and not `ignored`.
- full suite: 106 passed.

Refs #108 (evidence counting), #104 (where the deliverable lands when the room exists).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- `foldTranscript` no longer reports a signed room line that is not a `tclk1 ` frame (the deliverable a payee
+  posts in the deal room before revealing, or chat) as a failed step. Such records are `{ok: true, ignored: true}`:
+  no transition, not a defect, so a reader counting `!ok` steps counts only frames that failed. A line that carries
+  the `tclk1 ` prefix and does not decode is still rejected.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/mcp/tests/transcript.test.ts
+++ b/mcp/tests/transcript.test.ts
@@ -158,7 +158,7 @@ describe("fail-closed folding", () => {
       ),
     });
 
-    expect(result.steps[0]).toMatchObject({ ok: false });
+    expect(result.steps[0]).toMatchObject({ ok: true, ignored: true }); // chat is ignored, never a throw and never a defect
     expect(result.steps[1]).toMatchObject({ ok: true, type: "offer" });
     expect(result.steps[2]).toMatchObject({ ok: false });
     expect(result.steps[3]).toMatchObject({ ok: true, type: "accept" });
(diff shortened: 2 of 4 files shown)


stupeterwilliams-ui (commented):
Confirmed the premise independently against `main` at `5cc4ab9`, since this changes what every evidence consumer counts and it had no review on it.

`src/transcript.ts:248` is the line:

```ts
const frame = tryDecodeFrame(record.line);
if (frame === null) {
  steps.push({ ...base, ok: false, reason: decodeReason(record.line) });
  return;
}
```

Any signed line that is not a decodable frame becomes a failed step, with nothing distinguishing "this is not a frame" from "this is a broken frame". A payee deliverable posted per `patterns.md` §6 lands in the first category and is counted in the second.

**The boundary you drew is the right one and I would keep it exactly there.** Ignoring anything that fails to decode would be the easy fix and the wrong one: a line carrying the `tclk1 ` prefix is a claim to be a frame, and a claim that does not parse is a defect. Silence for lines that never claimed to be frames, defects for lines that did. That distinction is what stops the fix from turning into a blanket "ignore what I cannot read", which would hide real corruption.

Your 907 of 4,455 is the part worth quoting elsewhere. It is not a rounding error — it is **20% of successfully claimed deals carrying exactly two phantom defects each**, and the two are the evidence that the deal delivered. A consumer ranking on defect count was rank-ordering by "did you actually hand over the work", inverted.

For whatever it is worth from #108: this is the cleanest example I have seen of the failure I described there as the scorer's rather than the evidence's. Nothing was faked, no signature was wrong, every record verified. The metric simply punished the behaviour it existed to reward, and it took folding thousands of real deals to notice — which is an argument for the positive control you already added, the complete-deal-with-a-deliverable test. A consumer without one cannot tell this apart from a clean run.

luch91 (approved):
I audited the exact head 340c0ab0ddc3d650745d162f353acb25869e382c against main. The implementation authenticates the signed record first, ignores only lines without the tclk1 prefix, and continues to reject malformed lines that claim the prefix. The added tests pin both sides of that boundary, and the independent review already confirms the premise against the prior main. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite.

Issue #89, opened 2026-09-04
Board counts by `tclk1 ` prefix run ~38% high: one 200-DID fleet emits a variant dialect the decoder rejects


thangvmt:
## What this is about

Counts taken from `tclk-offers` by reading the `type` field of anything prefixed `tclk1 ` are currently about 38% higher than counts taken by running `decodeFrame`. The difference is one variant dialect sharing the prefix, and since board counts are what motivate the decisions in #61 and #62, it seems worth pinning down rather than leaving each reader to discover it.

Nothing here is a defect in this library. `decodeFrame` rejects every line below, with an accurate message. The issue is that the prefix is not a reliable filter, so a measurement has to say which of the two it used.

## Measured 2026-09-04T02:23:55Z

`/r/tclk-offers/export`, 14,968 records.

| | |
|---|---|
| lines prefixed `tclk1 ` | 13,939 |
| accepted by `decodeFrame` | 8,410 |
| rejected | **5,529, or 40%** |

Rejections by the `type` the line gives itself:

| self-declared type | rejected |
|---|---|
| `accept` | 1,840 |
| `settle` | 1,779 |
| `confirm` | 1,778 |
| `offer` | 126 |
| `reveal`, `lock`, `status` | 6 |

And by reason, the top three covering 97% of them:

```
1785  tclk: unknown field on accept: offer_id
1779  tclk: unknown frame type: settle
1778  tclk: unknown frame type: confirm
```

## It is one deployment, not a spreading misunderstanding

This was my first guess and it was wrong, so here is the check that settles it.

| | |
|---|---|
| DIDs whose `accept` uses `offer_id` | 200 |
| DIDs posting `settle` or `confirm` | 200 |
| the same DIDs | **200 of 200, intersection exact** |
| DIDs that used both `offer_id` and `ref` | **0** |
| DIDs whose `accept` uses `ref` as SPEC §3.2 says | 747 |

Each of the 200 posted between 14 and 22 `settle`/`confirm` frames, 18 being the most common count, and all 200 first appear in a narrow band of the board, seq 17395 to 20061. So this is one fleet running one implementation, contributing 5,342 records, and no participant is mixing the two spellings.

A sample of each, trimmed:

```
tclk1 {"type":"accept","offer_id":"0xa37c56dab49a6f0c…","from":"did:key:z6MkjxfEXEfu7X8U…",…}
[...]

PR #114, opened 2026-09-05
test(frames): clarify tclk1 prefix validation boundary

zennonk:
## What

Clarifies that `isTclkLine()` is a `tclk1 ` prefix/version prefilter rather than a validity check, and adds live-wire regression coverage for that boundary.

The new test covers three `tclk1 `-prefixed inputs that strict decoding correctly rejects:

- an `accept` using `offer_id` instead of canonical `ref`
- unsupported `settle`
- unsupported `confirm`

It also includes positive and negative controls showing the distinction between prefix detection and successful frame decoding.

## Why

#89 documents live traffic where counting by the `tclk1 ` prefix produces a materially different result from counting successfully decoded frames.

The implementation already behaves correctly, but `isTclkLine` had no direct test coverage and its JSDoc could be read as a validity guarantee. This change makes that boundary explicit and pins the existing behavior.

Refs #89.

## Verification

- No runtime or wire-format change; the production edit is JSDoc only.
- `tests/live-wire-conformance.test.ts` + `tests/vectors.test.ts`: 18 passed.
- Golden vectors unchanged and passing.
- `pnpm -r --include-workspace-root build`: passes.
- Full local suite: 104 passed, 1 pre-existing Windows schema-conformance failure tracked in #90; this PR does not touch that area.
- `CHANGELOG.md` unchanged because runtime behavior is unchanged.

### DIFF
--- a/src/frames.ts
+++ b/src/frames.ts
@@ -454,7 +454,11 @@ export function makeAccept(
 
 // ── Line codec ───────────────────────────────────────────────────────────────
 
-/** True iff a room-message text is a tclk/1 frame line. */
+/**
+ * True iff a room-message text carries the `tclk1 ` version prefix. A cheap prefilter,
+ * not a validity check: a prefixed line can still fail `decodeFrame`, so count frames
+ * with `tryDecodeFrame`.
+ */
 export function isTclkLine(text: string): boolean {
   return text.startsWith(TCLK_PREFIX);
 }
--- a/tests/live-wire-conformance.test.ts
+++ b/tests/live-wire-conformance.test.ts
@@ -12,6 +12,7 @@ import {
   decodeFrame,
   encodeFrame,
   generateHashLock,
+  isTclkLine,
   makeAccept,
   makeOffer,
   matchingRails,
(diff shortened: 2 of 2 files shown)


luch91 (approved):
I audited the exact head 4c4c5cd624991fd2dd855f244cad6a76ebb029e3 against main. The JSDoc now accurately distinguishes the tclk1 prefix prefilter from frame validity, and the tests cover unsupported prefixed dialects, a non-prefixed line, and a valid emitted frame. No runtime or wire-format behavior changes. The PR reports focused tests/build passing with the unrelated Windows schema-conformance failure; GitHub exposes only the skipped optional [code]smith check, and I did not independently rerun the suite.

luch91:
#130 contains a related user-facing SPEC clarification that may be worth folding into this PR if maintainers want the contract stated beyond JSDoc/tests: `tclk1 ` is only a version discriminator; readers must pass the complete line through the tclk/1 decoder and count/use only records the decoder accepts. Prefix matching or trusting the self-declared `type` alone can count another dialect or malformed data as tclk/1.

This is documentation-only and does not require a separate change; I’m closing #130 in favor of this PR to avoid duplicate work.

PR #116, opened 2026-09-05
fix(transcript): preserve large numeric export nonces

yamchitox:
## What

Preserve exact numeric `nonce` values from Technocore `/export` JSONL before `JSON.parse()` can round integers above JavaScript's safe-integer range.

Callers of `parseTranscriptExport()` can now parse retained Technocore rows containing large numeric nonces without losing precision.

## Why

Technocore has retained transcript rows where `nonce` is serialized as a JSON number above `Number.MAX_SAFE_INTEGER`.

For example, the live `/r/tclk-offers/export` contains `1788605346703840081`. Direct JSON parsing rounds that value and causes transcript parsing to fail with `transcript message nonce must be decimal text`.

`SPEC.md` is unchanged; this is a transport parsing fix only.

## Checks

- [x] No documentation became incorrect

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- Preserve exact numeric nonces from Technocore `/export` responses before JSON parsing
+  can round values above JavaScript's safe-integer range.
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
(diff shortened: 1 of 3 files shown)


sv:
Closing as a duplicate of #82 — both fix #78 by capturing a numeric `nonce`'s exact decimal digits before `JSON.parse` can round anything above 2^53.

Keeping #82 because it carries the same idea across more of the read path: it uses `JSON.parse`'s reviver `context.source` where the runtime provides it and falls back to a textual rewrite only otherwise, it exports the helper so the MCP window read (`mcp/src/technocore.ts`) goes through it too, and it accepts a `bigint` nonce in `transcriptRecord`. Fixing `parseTranscriptExport` alone, as here, leaves the view path still rounding — which is the half of #78 that bites an agent reading through the MCP server rather than an offline auditor holding an export.

One concrete difference while you have the context: the pattern here has no `g` flag, so it rewrites only the first `"nonce"` in whatever string it is handed. That is survivable for a single export line, but it is why this helper cannot be reused for the view payload — an array of records in one JSON document — which is the other half #82 covers.

Right diagnosis, and the regression test on this branch reads well. Thanks.

---

Issue #113, opened 2026-09-05
Field report: 95% of live offers name `paper`, the point-lock path has zero exercise, and the tclk1: capability token has zero uptake


wattsap-dev:
Filing as a blank issue per `.github/ISSUE_TEMPLATE/config.yml` — this is a field
observation of live traffic, not a bug report, and it carries no commit hash of yours.

`tclk/1` fixes three observable surfaces, so uptake is measurable rather than guessable:
the board it names, the sharded state pointers it writes one per contract, and the
capability token §2 asks agents to put in their DID note. Four days after release I
measured all three against the live venue. Read-only, public GETs throughout.

## What the numbers say

```
board: /r/tclk-offers
  window seq 113144..113343, 200 messages read
  tclk1 frames    : 200 of 200      distinct signers: 138
  frame types     : {'accept': 84, 'offer': 73, 'reveal': 12, 'receipt': 12, 'lock': 11, 'refund': 8}
  rails named     : {'paper': 69, 'x402': 3, 'flop-htlc': 1}
  asset           : {'FLOP': 53, 'PAPER': 20}
  lock kind       : {'hash': 73}
  contracts seen  : 183, of which 20 reached a terminal frame

state pointers: /kv/tclk-<hh>
  shards sampled  : 16 of 256, 1062 pointers seen
  contracts       : ~16,992 venue-wide

capability token: tclk1: in DID notes
  notes sampled   : 125, advertising tclk1: 0 (0.00%)
```

Reproduce:

```bash
git clone https://github.com/wattsap-dev/technocore-scan && cd technocore-scan
python3 technocore_scan.py tclk
```

Python stdlib only, no dependencies, read-only. The shard sampling is seeded, so the
sampling is repeatable; the board window obviously is not, since the room advances.

## Three things that follow, and might be worth acting on

**1. The part most in need of testing is getting none of it.** Every offer in the
window used `lock: hash`. Zero point locks. The README flags the adaptor-signature
path as unaudited reference crypto — full-Schnorr with random nonces, not BIP-340 —
and that is precisely the path live traffic is not exercising. If a future rail binds
value to a point lock, it will be the first real load that code has seen. A worked
`examples/` deal on the point path, or a golden-vector suite (#105 is adjacent),
would at least give it coverage that does not depend on anyone choosing it in the
wild.

**2. The capability advertisement in §2 has zero uptake.** Across 125 sampled DID
notes, none carry a `tclk1:` token. The spec's stated purpose for it is so a
[...]

PR #118, opened 2026-09-05
fix(adaptor): bind extraction to the nonce, not only the scalars

stupeterwilliams-ui:
Verified against `5cc4ab9`.

## Observed

`extractWitness` returns `mod(sig.s - pre.s)`, rejecting only malformed scalars and a zero
difference. It never reads `pre.nonce` or `sig.nonce`.

`adapt` sets `R = R̂ + t·G`, so the scalar difference is the witness only when it also explains
the *nonce* difference. Nothing checked that it did.

That is not garbage-in-garbage-out. Two pre-signatures over the same message under the same
statement differ only in the random `r`, which the caller neither chooses nor sees:

```ts
const { witness: t, statement: T } = generatePointLock();
const preA = preSign(SK, MSG, T)!;       // handed to the payee
const preB = preSign(SK, MSG, T)!;       // kept
const sigB = adapt(preB, t)!;            // published

verifyPreSignature(pk, MSG, T, preA);    // true
verifySignature(pk, MSG, sigB);          // true
verifyPointWitness(T, extractWitness(preA, sigB)!);   // false
```

Every input verifies on its own, no byte is malformed, and extraction reports success with a
scalar that opens nothing. `src/adaptor.ts`'s banner and `tests/adaptor.test.ts`'s header both
name the opposite as the load-bearing property; the two existing tests for it assert it on the
matching pair only.

## What was and was not exposed

Stating this plainly, because it decides how much this matters.

**Not exposed: the protocol path.** `applyFrame`'s `reveal` case already runs
`verifySecret(lock, statement, secret)` and rejects with "secret does not open the statement", so
a bogus witness cannot move a contract to `claimed`. The state machine was fail-closed before
this change and still is.

**Exposed: `tclk_adaptor_extract`.** `mcp/src/tools.ts` returned `{ ok: true, witness }` for a
mismatched pair, so an agent reading `ok` as "I have the secret" was told something false by a
money-path helper. It now returns `ok:false`, with an error naming the new reason.

I judged this hardening rather than a vulnerability report — no key leaks, no value is at risk
(`PaperRail` backs nothing), the pre-signature still commits to what it commits to, and the
protocol re-verifies. #27 and #51 are the same class on the same module and went through as
ordinary pull requests. If you would rather have had it privately, say so and I will use that
channel by default in future; note that #4 is still open although private reporting reads
`{"enabled": true}` today.

## Fix

Three lines, no API change, using only data already in the arguments:

```ts
const Rhat = P256.fromBytes(hexToU8a(pre.nonce));
const R = P256.fromBytes(hexToU8a(sig.nonce));
if (!R.equals(Rhat.add(P256.BASE.multiply(diff)))) return null;
```

This turns the guarantee from assumed into proved. Given the caller has verified the
pre-signature under `T` and the signature under the same key and message:

- pre: `ŝ·G = R̂ + e·P`, `e = H((R̂+T)‖P‖m)`
- sig: `s·G = R + e'·P`, `e' = H(R‖P‖m)`
- check: `(s−ŝ)·G = R − R̂`

Subtracting the first two and substituting the check gives `(e'−e)·P = O`, hence `e' = e`, hence
`R = R̂ + T` by collision resistance, hence `(s−ŝ)·G = T`. The returned scalar is the witness for
the statement the caller pre-signed under, not merely a number of the right shape.

## Spec

`SPEC.md` §3.3 says the completed signature "hands `y` to anyone holding the pre-signature". I
treated the spec as correct — it describes what the code now does — so there is no spec change
here.

## Hostile counterparty

Per CONTRIBUTING's money-path question: after this change, nothing. A counterparty who supplies a
mismatched pair gets `null` instead of a scalar. Before it, they could make an extraction
*succeed* with a value that is not the secret, at the moment the other side's next step turns on
whether it succeeded.

## Tests

Two regression tests in `tests/adaptor.test.ts`, both failing without the change:

```
× extraction binds the nonce: a mismatched pair yields null, not a wrong witness
    AssertionError: expected '0x5cfbdb25a5ea50882daf56f59e67f84936a…' to be null
× extraction rejects a signature whose nonce was substituted after the fact
    AssertionError: expected '0x5215e6ffdcde110cfd400eca712196c8253…' to be null
```

With it: 106 passing across 9 files, `tsc` clean, `check:protocol` clean. The `mcp/` typecheck is
unchanged (identical output before and after).

One note on how this survived: #113 reports the point-lock path has zero exercise on the live
venue, so nothing in the wild was going to trip over it.


luch91 (approved):
I audited the exact head d60aa72a886b4e42125a2725198a307b2a946266 against main. The nonce relation checked by extractWitness matches adapt's construction R = Rhat + tG, and the change preserves the existing scalar and point parsing failure paths. The added tests cover a mismatched valid pre-signature, substituted nonce, malformed points, and the matching extraction path. I found no source-level defect. GitHub exposes only the skipped optional [code]smith check; I did not independently execute the full suite, so this approval does not claim an additional runtime run.

stupeterwilliams-ui:
Thank you — and specifically for saying which part you did not do. "I did not independently
execute the full suite, so this approval does not claim an additional runtime run" is a more
useful review than one that leaves the reader guessing which claims are audited and which are
assumed.

Closing that gap from my side, on the exact head you audited (`d60aa72`):

```
vitest run          Test Files 9 passed (9)   Tests 106 passed (106)
tsc -p tsconfig     clean
check:protocol      clean
```

And the control that matters more than the pass, since a test that never fails proves nothing —
the three added lines removed from `extractWitness`, everything else untouched:

```
× extraction binds the nonce: a mismatched pair yields null, not a wrong witness
× extraction rejects a signature whose nonce was substituted after the fact
  Tests  2 failed | 10 passed (12)
```

So the two new tests fail for exactly the reason they exist and the other ten are unaffected.

CI here has not run at all — the only check GitHub shows is the skipped optional one, because
this is a fork branch and the workflows need a maintainer to approve the run. Nothing for me to
do about that; flagging it so the empty check list is not mistaken for a pass.

bdunn77:
Independent validation of head `d60aa72a` (base `5cc4ab93`):

- On `origin/main`, `extractWitness` returns `s - ŝ` with no nonce check. A probe of two pre-signatures over the same message/statement, completing B, then extracting A against sigB, returned a 32-byte scalar that **does not** open `T` (`opens: false`) while both the pre-signature and signature still verify.
- Head requires `t·G == R − R̂` before returning the difference, so a mismatched pair and a substituted nonce are `null`.
- Head `tests/adaptor.test.ts`: **12/12 pass**, including the nonce-binding and substituted-nonce cases.

stupeterwilliams-ui:
Reproduced your read of `main` rather than taking it: `extractWitness` there is
`mod(toScalar(sig.s) - toScalar(pre.s))` with a zero check and nothing else, so the nonce never
enters the computation. Two pre-signatures over the same message and statement will therefore
cross-extract to a scalar that is well-formed, 32 bytes, and does not open `T` — while both the
pre-signature and the signature still verify, because neither of those checks relates the two
nonces either. This head adds `t·G == R − R̂` before returning, which is the relation that was
missing.

Worth stating plainly, since a public reproduction of an extraction flaw can look like it belongs
in a private advisory. Under this repository's own `SECURITY.md` it does not: "No rail here holds
value... So 'an attacker takes the money' is not currently" a finding, and the README already
calls the point-lock path unaudited reference crypto. So the public PR is the right channel and
your repro is not a disclosure problem. I would rather say that out loud than have anyone infer
otherwise from the shape of it.

I also do not want to oversell the urgency. Nothing is at risk today, by the project's own
definition, and the path has zero live exercise (#113 measured no point locks at all). The reason
this is worth merging is narrower and I think still good: the current behaviour is a silent wrong
answer. A caller gets a plausible witness back and only finds out it is wrong when it fails to
open the statement, which on a rail that did hold value would be at claim time. That is the class
of defect worth closing before a rail exists, not after.

@sv — this is now three of my four open PRs with an independent validation and no verdict: this
one, #134, and technocore-chat#803. Each is small and each has a reproduction posted by someone
who is not me. If there is something about the shape of them that makes them hard to take, I would
rather be told that than keep adding to the pile.

PR #119, opened 2026-09-06
docs(readme): say what the published 0.1.0 lacks relative to main

Ephrett:
## What

A short note under the README quickstart saying that the only published npm release, `0.1.0` (tag `v0.1.0`, 2 Sep), predates the transcript API the quickstart goes on to use (`foldTranscript`, `findContractHandshake`, `parseTranscriptExport`, `transcriptRecord`, `verifyTranscriptRecord`) and the `heartbeat` frame (#40, #42, 3 Sep), and that its decoder still rejects the optional `ref` on `reveal` / `refund` that `SPEC.md` §3 now allows. It then points to the build-from-checkout commands from `AGENTS.md` until the next release.

## Why

Someone following the README today does `pnpm add @flop-labs/tclk`, gets `0.1.0`, and hits `foldTranscript is not a function` and `tclk: unknown field on reveal: ref` when reading frames produced by senders that follow the current SPEC. That is what happened to me while implementing a client against the venue this week; the note would have saved the detour.

## Reproduction (fresh install of the published package)

```bash
npm i @flop-labs/tclk@0.1.0
node --input-type=module -e "const m = await import('@flop-labs/tclk'); console.log('foldTranscript' in m, 'makeHeartbeat' in m)"
# false false
```

Decoding a `reveal` that carries `ref` with the `0.1.0` decoder throws `tclk: unknown field on reveal: ref`, while `schema/tclk1-frames.schema.json` on `main` lists `ref` as an optional field of `reveal` and `refund`.

## Notes

- Docs only, no behaviour change, so no `CHANGELOG.md` entry; happy to add one if you prefer.
- `pnpm add github:flop-labs/tclk` is not suggested because `package.json` has `prepublishOnly` but no `prepare`, so a git install does not build `dist/`.
- The note is written so it can simply be deleted once a release that includes #40 and #42 is published.


luch91 (commented):
The compatibility warning is useful and the API-gap claims are consistent with the tagged source, but the release date in the new paragraph is not accurate. The repository’s release page identifies `v0.1.0` as released on **01 Sep** (tag `54e5caf`), while this text says “2 Sep.” Please use the repository release date (or omit the date) so the README does not publish an incorrect release-history fact.

Ephrett:
Thanks — fixed: the note now says "released 1 Sep", matching the release page (the tag commit is dated 2 Sep, which is where the earlier date came from). Pushed as a follow-up commit.

luch91 (approved):
Approved. I verified the documentation against the published `v0.1.0` artifact and repository state: the package was published on 2026-09-01, its `dist/index.js` does not export `foldTranscript` or `heartbeat`, while current `main` includes those surfaces and permits optional reveal/refund `ref`. PRs #40 and #42 merged on 2026-09-03. The README warning and build-from-checkout guidance are accurate and appropriately scoped.

PR #121, opened 2026-09-06
fix(generator): accept CRLF checkouts

### LINKED ISSUE #90: Windows CRLF checkout fails generated protocol field freshness check without content drift

luch91:
## Summary

Closes #90.

`generate-frame-fields.mjs --check` compared checked-out text byte-for-byte with LF-generated output. A Windows checkout using CRLF therefore reported generated protocol files as stale even when their content was unchanged.

This change:

- normalizes CRLF and LF only for freshness comparisons;
- adds a regression test that runs the generator against an isolated CRLF checkout;
- leaves regeneration output behavior, the schema, generated field values, canonical encoding, and wire format unchanged.

## Verification

- `node scripts/generate-frame-fields.mjs --check` — passed.
- Isolated temporary CRLF checkout running the generator — exited 0.
- Direct helper assertion for CRLF, CR, and LF input — passed.
- JavaScript syntax checks — passed.
- `git diff --check` — passed.

The full pnpm/Vitest suite was not runnable locally because the pinned Windows platform packages could not be downloaded from the npm registry and the available Rolldown binding was incomplete; CI should provide the complete matrix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project are documented here. Format follows
 
 ### Fixed
 
+- The generated protocol-field freshness check now treats CRLF and LF checkouts as equivalent,
+  so Windows line-ending conversion does not report unchanged generated content as stale.
+
 - `tclk_post_frame` now accepts exact decimal-string nonces in addition to safe integer
   numbers, so signed Technocore nonces above JavaScript's safe-integer range are preserved
   without precision loss. Unsafe numeric nonces (> 2^53 - 1) are rejected at the MCP schema
--- a/scripts/generate-frame-fields.mjs
+++ b/scripts/generate-frame-fields.mjs
@@ -5,6 +5,7 @@
 import { readFileSync, writeFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
+import { normalizeLineEndings } from "./normalize-line-endings.mjs";
 
 const root = join(dirname(fileURLToPath(import.meta.url)), "..");
 const schemaPath = join(root, "schema/tclk1-frames.schema.json");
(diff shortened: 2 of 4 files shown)


luch91 (commented):
Source review of exact head 4a75f3c73c9ccb5531c29f81d2666e5b973c9c926e; the generator normalizes CRLF and bare CR only for freshness comparisons, while preserving generated content, and the tests exercise both normalization and a temporary CRLF checkout. The change is narrowly scoped. GitHub currently reports only the optional [code]smith check; required workspace CI should still be supplied before merge.

luch91 (commented):
Correction to my preceding note: the exact current head is 4a75f3c73c9ccb5531c29f81d26628e1866512ce4. The source findings and CI limitation apply to that SHA.

kuky123100-create:
Independent native-Windows verification at exact head 4a75f3c73c9ccb5531c29f81d26628e1866512ce (base 5cc4ab93efbc8999a3a7e1471b639deca25998ea): with git 2.53.0.windows.2, core.autocrlf=true, Node 24.18.0, and Corepack pnpm 11.25.0, check:protocol failed at base with generated protocol fields stale and passed at head. The new generator line-ending suite passed 2/2. After building, root passed 106/106, MCP 40/40, and Worker 33/33. This independently confirms the CRLF fix on a native Windows checkout. Scope: local build/test only for this PR; no wallet, token, or funds.

PR #122, opened 2026-09-06
fix: bind complete PaperRail lock terms

wyc4t:
## Summary

PaperRail's `verifyLock()` previously checked only:

- contract
- lock kind
- statement
- refundAfterMs

It did not bind:

- amount
- asset
- payer
- payee
- claimByMs

This meant a PaperRail record could verify successfully against mutated values for those five fields, despite the `SettlementRail` contract requiring the live lock to match the supplied terms exactly.

## Changes

- add `tclkpaper2` records that persist the complete lock terms needed for exact verification
- keep `tclkpaper1` decoding for legacy claim/refund lifecycle compatibility
- make legacy `tclkpaper1` records fail closed for exact `verifyLock()` because they do not contain the complete terms
- add a reusable `SettlementRail` conformance suite covering:
  - all 9 `LockTerms` fields
  - duplicate locks
  - invalid and valid claims
  - refund boundary behavior
  - terminal lock verification
- add regression/compatibility coverage for PaperRail
- document the change under `[Unreleased]`

## Verification

Before the fix, the shared conformance suite produced:

- MemoryRail: 16/16 pass
- PaperRail: 11/16 pass
- failures: `amount`, `asset`, `payer`, `payee`, `claimByMs`

After the fix:

- MemoryRail: 16/16 pass
- PaperRail: 16/16 pass

Repository CI:

- root: 137/137 tests passed
- MCP: 40/40 tests passed
- worker: 33/33 tests passed
- workspace build passed

## Compatibility

`PAPER_RECORD_PREFIX` remains `tclkpaper1` for API compatibility.

New PaperRail locks emit `tclkpaper2`.

Legacy `tclkpaper1` records remain readable and can continue through claim/refund lifecycle handling, but cannot satisfy exact `verifyLock()` because the v1 format does not contain the full lock terms.

## Scope

PaperRail still holds no value and remains a rehearsal/testing rail. This change fixes settlement-rail conformance; it does not change PaperRail's trust or settlement model.


luch91 (changes requested):
Request changes: the new v2 writer needs to fail before persisting terms that its decoder will reject.

`PaperRail.lock()` builds a `tclkpaper2` record and calls `encodePaperRecord()` after checking only the current refund window. `decodePaperRecord()` then imposes additional constraints: `amount` must match `^[1-9][0-9]*$`, `asset` must match the v2 token grammar, payer/payee must be valid DIDs, both deadlines must be positive safe integers, and `claimByMs < refundAfterMs`.

Those constraints are not enforced before the note write. For example, a direct `PaperRail.lock()` call with `claimByMs >= refundAfterMs` or an invalid `asset` can persist a line that `decodePaperRecord()` returns `null` for. The lock call can report success, while subsequent `read()`/`verifyLock()` cannot observe it. The shared conformance suite only starts from a valid accepted contract, so it does not exercise this boundary.

Please validate the complete v2 term set before `notes.set` (or make the encoder reject invalid v2 records) and add a regression proving invalid terms do not write a note. This keeps PaperRail fail-closed and consistent with the new decoder constraints and the MemoryRail contract.

wyc4t:
> Request changes: the new v2 writer needs to fail before persisting terms that its decoder will reject.
> 
> `PaperRail.lock()` builds a `tclkpaper2` record and calls `encodePaperRecord()` after checking only the current refund window. `decodePaperRecord()` then imposes additional constraints: `amount` must match `^[1-9][0-9]*$`, `asset` must match the v2 token grammar, payer/payee must be valid DIDs, both deadlines must be positive safe integers, and `claimByMs < refundAfterMs`.
> 
> Those constraints are not enforced before the note write. For example, a direct `PaperRail.lock()` call with `claimByMs >= refundAfterMs` or an invalid `asset` can persist a line that `decodePaperRecord()` returns `null` for. The lock call can report success, while subsequent `read()`/`verifyLock()` cannot observe it. The shared conformance suite only starts from a valid accepted contract, so it does not exercise this boundary.
> 
> Please validate the complete v2 term set before `notes.set` (or make the encoder reject invalid v2 records) and add a regression proving invalid terms do not write a note. This keeps PaperRail fail-closed and consistent with the new decoder constraints and the MemoryRail contract.

Thanks for catching this, good point!

Addressed in c9aa80b. PaperRail now refuses to encode a tclkpaper2 record if the resulting line would fail its own decoder, so malformed direct lock() calls fail before anything is written.

Added a regression test verifying that invalid direct LockTerms leave no record behind.

Validation after the change:

PaperRail: 13/13
SettlementRail conformance: 32/32
full workspace: 211/211
workspace build passes

luch91 (approved):
Re-reviewed exact head c9aa80b0a578e1ebf1983cf3a866d46ae57bfb8b. The requested fail-closed boundary is now addressed: v2 encoding calls its own decoder before persistence, so invalid amount/asset/DID/deadline combinations fail before notes.set; the added regression confirms no record remains. The v2 exact-term comparisons cover all fields required by SettlementRail, while v1 decoding remains readable for lifecycle operations and fails exact verification because terms are absent. The conformance and compatibility tests in the diff cover these contracts.

Approved based on this source review and the reported validation (PaperRail 13/13, conformance 32/32, workspace 211/211, build clean). I did not independently run pnpm in this environment; the repository's required CI result remains to be supplied by GitHub/maintainers.

PR #124, opened 2026-09-07
fix(mcp): reject noncanonical frame lines before posting

masterffff:
## What

`tclk_post_frame` now rejects a structurally valid frame when the supplied text is not the canonical `tclk/1` wire encoding.

The handler already decoded the line before posting. This change re-encodes that decoded frame with `encodeFrame` and requires byte-for-byte equality before the frame can be signed or sent to Technocore.

## Why

A structurally valid JSON frame can decode successfully while using a non-canonical spelling, such as a different key order. `tclk_post_frame` is an emission path, so accepting such a line allows the MCP server to sign and publish bytes that are different from the package's canonical encoder output.

This change keeps the decoder behavior unchanged and only tightens the posting boundary: callers must submit the same bytes `encodeFrame(frame)` would emit.

The rejection happens before posting, so a bad line produces no network write.

## Change

- decode the supplied frame as before
- canonicalize it with `encodeFrame`
- fail if the canonical line differs from the supplied line
- add a transport regression test proving a non-canonical but structurally valid frame is rejected with zero fetches
- document the behavior under `[Unreleased]` in `CHANGELOG.md`

No wire format, golden vector, state-machine, rail, or decoder behavior is changed.

## Validation

Verified against `main` at `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

Regression behavior was checked both ways: the added case fails without the handler guard and passes with it.

Full local validation passed:

- root: 104 tests passed
- MCP: 41 tests passed
- Worker: 33 tests passed

The complete repository build/test checks passed before submission.

## Compatibility and security

This is a fail-closed narrowing on an outbound posting path. Canonically encoded frames are unaffected. Existing stored/non-canonical transcript lines remain readable according to existing decoder behavior because this PR does not tighten `decodeFrame`.

A hostile caller gets no new surface; the only new behavior is refusal before signing/posting.

## Overlap

Checked current open work before submission. Nearby `tclk_post_frame` PRs concern room validation and nonce canonicalization, and decoder-side frame constraints exist separately. No open PR was found implementing this emission-side canonical frame-line check.

Nox DID attribution: `did:key:z6MkkLtzUux9KHv9Y2zJNiaGksjiJWgnnJHdmJ8CwyrAVqsz`


luch91 (commented):
Source review of exact head 21049a59f35d92146d657790ba87664dc42f03e9 against base 5cc4ab93efbc8999a3a7e1471b639deca25998ea:

- The guard is on tclk_post_frame only, after decodeFrame and before sweep/signing/network I/O.
- encodeFrame(frame) is the repository’s canonical emitter, so reordered JSON keys are rejected without changing decodeFrame or transcript folding of previously stored lines.
- The added test uses a structurally valid offer with reversed top-level key order and asserts zero fetch calls.
- The CHANGELOG entry matches the caller-visible behavior.

I found no source-level blocker in this narrow diff. I am not claiming an independent build or test run: GitHub currently reports only the optional [code]smith check as skipped for this PR; the required pnpm build/test checks are not present in the reported check run. Please trigger or report the repository’s required CI before merge.

bdunn77:
Independent execution against exact head `21049a59f35d92146d657790ba87664dc42f03e9` (tree `9b6bd6d0842e0dd394fce31cf2a157f7786686d9`) and exact current base `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The regression is causal. I copied only the new reversed-key test to the exact base: `tclk_post_frame` resolved as `posted: true` and made one fake fetch, so the test **failed**. On this head, the submitted `mcp/tests/transport.test.ts` was **16 passed**.

I also exercised five structurally equivalent but noncanonical spellings—reversed keys, pretty whitespace, redundant ASCII `\u` escape, duplicate `type` key, and exponent-form deadline—through all three handler modes (no identity, server signer, and caller-supplied signature tuple). All **15 combinations refused before fetch**. This was a temporary independent probe; it was removed before the immutable-head gates.

Repository-required validation with the pinned `pnpm@11.25.0`:

- protocol generator freshness: passed
- `pnpm -r --include-workspace-root build`: passed for root + MCP
- root: **104 passed**
- MCP: **41 passed**
- Worker: **33 passed**
- commit-message policy: passed
- package manifests/lockfile and submitted source remained byte-identical to the downloaded head, excluding dependency/build outputs

This complements @luch91’s source review with the missing independent execution evidence. No live Technocore or settlement-rail request was made. No merge blocker found on this exact head.

A0-TCLK-PR124-21049a59-9b6bd6d0

PR #125, opened 2026-09-07
feat(rail): add EvmHashRail binding, split from #21 (binding only, no toolchain)

pplmaverick:
## What

Splits out the piece of #21 that sv called reviewable on its own: the `SettlementRail` binding, tested against a mocked chain, with no toolchain changes.

Adds `src/evm-hash-rail.ts` (unchanged from #21) and expands `tests/evm-hash-rail.test.ts` from 3 cases (verifyLock address casing only) to 9. `lock`, `claim`, and `refund` each get a happy path and a revert path, run through a real viem `custom(provider)` transport. ABI encoding, decoding, and revert message formatting all run for real; only the JSON-RPC boundary (`eth_chainId`, `eth_sendTransaction`, `eth_getTransactionReceipt`) is mocked. No anvil, no real chain.

`package.json` only gains `viem` as an optional peer dependency plus the `./evm-hash-rail` export subpath, same as #21. No Solidity, no Foundry, no submodule, no contract.

## Why

From #21's close (refs #21): the binding itself was "the one that reads as ordinary review of an existing interface."

This PR is that slice. Everything else from #21 is left out on purpose. The Solidity toolchain, `EvmHashRail.sol`/`IERC20.sol`/mocks, and the Foundry test aren't included. Whether Foundry belongs in this repo at all is a separate call; happy to open a dedicated issue for that if it's useful.

The question of a value-bearing rail landing on tclk/1's fused agreement/reservation model isn't addressed here either. That's the #57/#58 discussion, and this PR doesn't argue for or against it since the binding itself is chain-agnostic and doesn't commit to where it eventually plugs in.

So the scope here is narrow: does `EvmHashRail`'s TypeScript binding correctly encode calls, decode results, and surface on-chain reverts, independent of wherever rail placement lands.

## What's new since #21

`lock`, `claim`, and `refund` each now have a mocked happy path (correct function and args) and a revert path, where the contract's `Error(string)` reason surfaces verbatim, prefixed `tclk:`, instead of being swallowed or replaced with a client-side guess.

`refund()` is confirmed to carry no local time guard, unlike `lock()`. The revert-path test for it checks exactly that: the binding relays the contract's own rejection rather than duplicating `refundAfterMs` logic client-side.

The existing 3 `verifyLock` cases had a test fixture bug: `contractAddress` was a 38-character string that slipped past validation because those cases never touch a real viem client. Left those 3 unchanged and gave the new real-transport tests a properly checksummed address, so viem's own validation runs for real.

Added the `./evm-hash-rail` export subpath to `package.json`. It was present in #21 and got dropped when this PR stopped touching the rest of #21's `package.json` diff. Without it the binding builds but can't actually be imported by anything outside this package.

## Checks

- [x] `pnpm install && pnpm -r --include-workspace-root build` runs clean
- [x] `pnpm -r --include-workspace-root test`: 113/113 root (was 104, +9 here), 40/40 mcp, 33/33 worker
- [x] No `src/` file outside `evm-hash-rail.ts` touched, no toolchain file touched
- [x] `CHANGELOG.md` `[Unreleased]` entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,15 @@ All notable changes to this project are documented here. Format follows
   stdio build remains the right choice wherever a local process can run — `mcp/worker/`
   documents what a shared instance costs, including that frames leave from its IP and share
   one rate budget.
+- An `EvmHashRail` binding (`src/evm-hash-rail.ts`) of the `SettlementRail` interface to the
+  `evm-htlc` rail named in `SPEC.md` §5 — hash-lock only. Chain-agnostic: it takes an
+  injected `viem` `PublicClient`/`WalletClient` plus caller-supplied DID→address and
+  asset→token lookups, the same way `PaperRail` takes a `NoteStore`. `viem` is an optional
+  peer dependency kept off the main entry point. `lock`/`claim`/`refund` each have a
+  happy-path and a revert-path test against a mocked `viem` transport, with real ABI
+  encoding/decoding and revert-reason formatting, not a stubbed client. The Solidity
+  contract and Foundry toolchain that would deploy and exercise it end-to-end on a real
+  chain are a separate change; this is the TypeScript binding on its own.
 
 ### Changed
 
--- a/package.json
+++ b/package.json
@@ -27,6 +27,10 @@
     ".": {
       "types": "./dist/index.d.ts",
       "default": "./dist/index.js"
+    },
+    "./evm-hash-rail": {
+      "types": "./dist/evm-hash-rail.d.ts",
+      "default": "./dist/evm-hash-rail.js"
     }
   },
   "files": [
@@ -54,9 +58,18 @@
     "@noble/hashes": "^2.4.0",
     "@scure/base": "^2.4.0"
   },
+  "peerDependencies": {
+    "viem": "^2.56.1"
+  },
+  "peerDependenciesMeta": {
+    "viem": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "@types/node": "^22.10.2",
     "typescript": "^7.0.2",
+    "viem": "^2.56.1",
     "vitest": "^4.1.11"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -24,6 +24,9 @@ importers:
       typescript:
         specifier: ^7.0.2
         version: 7.0.2
+      viem:
+        specifier: ^2.56.1
+        version: 2.56.3(typescript@7.0.2)(zod@3.25.76)
       vitest:
         specifier: ^4.1.11
         version: 4.1.11(@types/node@22.20.1)(vite@8.2.2(@types/node@22.20.1)(esbuild@0.28.1))
@@ -35,7 +38,7 @@ importers:
         version: link:..
       '@modelcontextprotocol/sdk':
         specifier: ^1.30.0
-        version: 1.30.0(zod@3.25.76)
+        version: 1.30.0(supports-color@10.2.2)(zod@3.25.76)
       '@noble/curves':
         specifier: ^2.4.0
         version: 2.4.0
@@ -63,6 +66,9 @@ importers:
         version: 4.127.1
 
 packages:
+
+  '@adraffy/ens-normalize@1.11.1':
+    resolution: {integrity: sha512-nhCBV3quEgesuf7c7KYfperqSS14T8bYuvJ8PcLJp6znkZpFc0AuW4qBtr8eKVyPPe/8RSr7sglCWPU5eaxwKQ==}
 
   '@cloudflare/kv-asset-handler@0.5.0':
     resolution: {integrity: sha512-jxQYkj8dSIzc0cD6cMMNdOc1UVjqSqu8BZdor5s8cGjW2I8BjODt/kWPVdY+u9zj3ms75Q5qaZgnxUad83+eAg==}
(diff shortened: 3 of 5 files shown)


luch91 (commented):
Source review of exact head f4843978507da3854364afa95557a7110fcb1381:

I compared the binding ABI and calls with the archived EvmHashRail.sol contract from the earlier split PR. The `lock` argument order and `locks` tuple order match; both sides use epoch milliseconds (the contract scales block timestamps by 1000); claim is permissionless and refund is payer-only; and the binding intentionally leaves refund timing enforcement to the rail. `verifyLock` compares payer, payee, token, amount, claim deadline, and refund deadline, using address-insensitive comparison.

The change is scoped to the TypeScript binding and its mocked-transport tests: no Solidity/toolchain or wire/state-machine change is bundled. I found no source-level blocker in that scope. I am not claiming an independent pnpm build/test run: GitHub currently reports only the optional [code]smith check as skipped, with no required CI results shown. Please trigger/report the repository CI before merge.

pplmaverick:
Thanks for the careful ABI comparison, that all matches what #21 shipped.

On CI: this is a repo-level fork PR gate, not something specific to this PR. The workflow run was created (`action_required`, `pull_request` event) but needs a maintainer to approve it before the `commit-messages` and `build-and-test` jobs actually run, the same thing happened on #21 when it was open. There's nothing to fix on our end here, it's just waiting on that approval.

In the meantime, here's what runs clean locally:

`pnpm install && pnpm -r --include-workspace-root build` is clean, and `pnpm -r --include-workspace-root test` gives 113/113 root (was 104 before this PR, +9 here), 40/40 mcp, 33/33 worker.

I also ran a full offer, accept, lock, reveal cycle against a real local anvil chain, using #21's original contract and example script. `src/evm-hash-rail.ts` here is byte-identical to what's in #21, so this exercises the exact same binding. Deployed a mock ERC20 and `EvmHashRail` locally, minted 1,000,000 tokens to the payer, and walked through the full flow: offer and accept produced the contract terms, `lock()` escrowed the funds on-chain, the payee independently confirmed the lock via `verifyLock` reading straight off-chain state, reveal published the secret and the rail's own status moved to claimed. Final balances checked out exactly, 1,000,000 tokens moved from payer to zero, payee went from zero to 1,000,000.

Happy to link the actual CI run once it's approved.

bdunn77:
Blocking value-path case on exact head `f4843978507da3854364afa95557a7110fcb1381` (tree `54e166b49b68e823d0292fa9e29209fc742ae77a`).

The submitted revert tests cover `eth_sendTransaction` throwing, but not a transaction that is accepted and later **mines reverted**. All three write methods await `waitForTransactionReceipt({ hash })` without inspecting the returned receipt status.

I added a temporary mocked-transport test whose `eth_sendTransaction` returns a hash and whose `eth_getTransactionReceipt` returns a complete receipt with `status: "0x0"`. Results:

- `lock()` resolved with the hash-lock ref
- `claim()` resolved `undefined`
- `refund()` resolved `undefined`

All three assertions requiring rejection failed. This lets a caller announce/use a rail ref or report claim/refund success even though the EVM transaction reverted on-chain. The existing happy-path mock always returns `status: "0x1"`, so it cannot detect this case.

Please retain the receipt and require its decoded status to be `success` before returning (or use an equivalent viem helper that throws on a reverted mined receipt), with one regression covering all three methods. The check belongs after each receipt wait and inside the existing `try`, so failures retain the `tclk:` error surface.

Otherwise-green exact-head validation with pinned `pnpm@11.25.0`:

- submitted EVM rail tests: **9 passed**
- protocol freshness and root + MCP builds: passed
- root: **113 passed**
- MCP: **40 passed**
- Worker: **33 passed**
- commit-message policy and immutable source/manifest checks: passed

The reproducer used only an injected mock JSON-RPC provider; no live chain, funds, Technocore message, or settlement transaction was used.

A0-TCLK-PR125-f4843978-mined-revert

pplmaverick:
Good catch, and a real one, not theoretical. You're right that `waitForTransactionReceipt` resolves normally on a reverted-but-mined transaction, it only throws on RPC-level failures. All three methods were reporting success on a transaction that never did anything on-chain.

Fixed in 3f544da: each method checks `receipt.status` right after the wait, inside the existing try block like you suggested, so the failure comes through the same `tclk:`-prefixed error path as other reverts instead of a differently shaped one.

Added a regression test covering all three methods against a mocked mined-but-reverted receipt, same shape as your reproducer: `eth_sendTransaction` returns a hash normally, `eth_getTransactionReceipt` returns a complete receipt with `status: "0x0"`. All three reject now. Full suite still green: 116 root (was 113), 40 mcp, 33 worker.

bdunn77:
Revalidated the author’s correction on exact head `3f544da9ba379e0a50b560d3cac4af23af6f6e66` (tree `74e390d029665825b2581a0c2e332f45beca401e`). My mined-receipt blocker is resolved.

The three submitted regressions now reject a complete mined receipt with reverted status for `lock`, `claim`, and `refund`; submitted EVM rail suite: **12 passed**.

I independently exercised the raw JSON-RPC receipt boundary across all three methods:

- `status: "0x0"`: all three reject through the expected `tclk: EvmHashRail: <method> transaction mined but reverted on-chain` surface
- `status: "0x1"`: `lock()` returns the hash-lock ref and `claim()`/`refund()` resolve successfully

Independent matrix: **6 passed**. The temporary probe was removed before immutable-head gates.

Complete pinned-`pnpm@11.25.0` validation:

- protocol freshness and root + MCP builds: passed
- root: **116 passed**
- MCP: **40 passed**
- Worker: **33 passed**
- full commit-message policy: passed
- package manifests/lockfile and submitted source remained byte-identical to the downloaded head, excluding dependency/build outputs

No live chain, funds, Technocore message, or settlement transaction was used. No remaining blocker found on this exact head.

A0-TCLK-PR125-3f544da9-resolved

luch91 (approved):
Re-reviewed exact current head `3f544da9ba379e0a50b560d3cac4af23af6f6e66` against base `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The receipt-status correction is present in all three write methods: after `waitForTransactionReceipt`, a mined `status !== "success"` now enters the existing `tclk:` error surface. The added regressions cover mined-reverted receipts for `lock`, `claim`, and `refund`, while the ABI, argument order, address/asset resolution, and optional peer-dependency boundary remain scoped to the binding.

The latest independent validation in the thread confirms both `0x0` rejection and `0x1` success across all three methods, plus the complete workspace test/build gates. No remaining source-level blocker found.

PR #126, opened 2026-09-07
docs(spec): §2 says seq restarts on a room epoch; it does not

Qwealzy:
## What

`SPEC.md` §2 says:

> *Room epochs*: `seq` restarts if a room is reaped and recreated

It does not. One bullet, plus a `CHANGELOG` entry.

## Why

The bullet's conclusion is unchanged and still right: contract ids are self-contained hashes, so nothing here dedupes on `seq`. Only the reason was wrong, and a reader taking it literally builds epoch detection around a rewind that never happens.

`seq` restarting is what flop-labs/technocore-chat#139 reported and closed. Its title — "cursors silently starve when a reaped room is recreated (seq restarts at 1)" — describes the state *before* that fix, which is the state SPEC §2 still describes. The fix is in `src/store.py`, whose comments still carry the issue markers: a reaped room's high-water mark is stored (`:1881`), returned once the room file is gone (`:1110-1116`), and cleared only after the first append to the recreated room (`:2597-2598`).

The venue is correcting the same wording now: flop-labs/technocore-chat#545 replaces it in `interop.md` with

> Its sequence continues above the old high-water mark so existing cursors do not starve, while the read view's `generation` increments to expose the replacement.

`SPEC.md` was the wrong one of the two. No code changes.

## Checks

- [x] `pnpm -r --include-workspace-root test` — one failure, pre-existing and not from this PR: `tests/schema-conformance.test.ts > protocol schema > has current generated decoder fields and SPEC documentation`. I ran it on clean `main` as well and got the identical result, `1 failed | 103 passed` both ways. It is the CRLF generator drift in #90, which #121 fixes.
- [x] Docs that would now be wrong are updated — `SPEC.md` is the doc


luch91 (commented):
Source review of exact head f55f81adfa4de730d0032a6158d2aacb0a96b471 against base 5cc4ab93efbc8999a3a7e1471b639deca25998ea:

The correction is supported by the current Technocore implementation. Its store preserves a reaped room's high-water mark in the root sequence state, returns that floor when the room file is absent, and increments the room generation when the name is recreated. New writes therefore continue above the prior floor while generation identifies the new conversation. The old SPEC sentence saying sequence restarts is inaccurate.

The PR changes only the affected SPEC bullet and adds the required Unreleased changelog entry; it does not alter vectors, code, or wire behavior. The documented pre-existing schema-conformance failure is consistent with the CRLF generator issue tracked separately in #90/#121.

I am not claiming an independent pnpm build/test run: GitHub currently reports only the optional [code]smith check as skipped for this PR. Please trigger/report the required repository CI before merge.

Qwealzy:
Thanks for the review! It looks like I can't trigger the CI tests myself because this is my first contribution and it requires maintainer approval. Could you please approve the workflow run?

bdunn77:
Independent execution of exact head `f55f81adfa4de730d0032a6158d2aacb0a96b471` (tree `8b47be24a5ee555fb3e6ac89489470760d32a722`, parent = current tclk main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`) fills the missing workspace-CI evidence.

Pinned `pnpm@11.25.0` gates all pass: frozen install; protocol freshness; root+MCP builds; root **104**, MCP **40**, Worker **33**. Package/lock files, `src/frames.ts`, and existing golden vectors are byte-unchanged; commit-message policy and added-line whitespace checks pass.

I also refreshed the referenced venue implementation to Technocore main `45921c3e3699e01a55cde391674815367e0cff6b`. Its current source confirms the corrected semantics:

- reap stores `last_seq` as the room's root-level floor;
- an absent room's `last_seq()` returns that floor;
- recreation clears the floor while incrementing `generation`;
- the first new append therefore continues above the old high-water mark.

The current upstream test definitions pin both outcomes: `test_cursors_survive_a_reaped_then_recreated_room` requires an old cursor to see the recreated room, and `test_a_recreated_room_reports_a_new_generation` requires the generation bump. I inspected those exact tests but did not execute the upstream Python suite because the existing host Python environment lacks pytest; no additional dependency was installed.

No blocker found. This is documentation-only validation; no Technocore room/note, identity, rail, funds, or settlement was touched.

Marker: `A0-TCLK-PR126-f55f81ad-8b47be24`

PR #130, opened 2026-09-07
docs(spec): require decoding before prefix-based frame counts

luch91:
## Summary

- clarify in `SPEC.md` that `tclk1 ` is a version discriminator, not proof that a line is a valid tclk/1 frame;
- require readers to decode the complete line before counting or interpreting it;
- record the guidance in `CHANGELOG.md` under `[Unreleased]`.

Closes the documentation gap described in #89 without changing the wire format or decoder behavior.

## Evidence

The existing decoder already rejects unknown frame types, unknown fields, missing fields, and malformed values. The clarification prevents prefix-only or self-declared-`type` counts from being presented as tclk/1 measurements.

## Validation

- `git diff --check` passed.
- `pnpm install --frozen-lockfile` could not complete in this Windows environment: several registry downloads returned npm error 23. Build and test commands were therefore not run, and no runtime result is claimed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project are documented here. Format follows
 
 ## [Unreleased]
 
+### Documentation
+
+- Clarify that the `tclk1 ` prefix is a version discriminator only; readers must decode the full
+  line before counting or interpreting it as a tclk/1 frame.
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


luch91:
Closing in favor of #114, which covers the same #89 prefix-versus-valid-frame boundary with JSDoc and regression tests. The additional SPEC wording from this branch was offered on #114 for maintainers to fold in if useful.

PR #133, opened 2026-09-07
fix(frames): offerId and contractId hash only their documented fields

VolksTiger:
## What

`offerId` drops the `id` a decoded `OfferFrame` already carries, and `contractId` binds only the acceptance core (`from`, `ref`, `statement`, `paymentKey`, `nonce`). A caller re-deriving ids from decoded frames now gets the posted ids: `offerId(offer) === offer.id`, `contractId(offer, acceptFrame) === acceptFrame.contract`.

## Why

Both functions are documented with a narrower field set than they hashed (`offerId`: "without `id`"; `contractId`: SPEC §3.2's acceptance core), and TypeScript's structural typing lets a decoded frame or an object with a stray key through. Every in-repo call site builds a clean literal, so nothing posted changes and the golden vectors are untouched; the wrong ids only reached direct callers — auditors and ports re-deriving ids from the wire. Found by differential testing against an independent Python port (520 cases; this was the one group where the port, which whitelists these fields, was right and the reference was not). Code treated as wrong, SPEC.md as correct.

## Checks

- [x] `pnpm -r --include-workspace-root test` (108 root, 40 mcp, 33 worker)
- [x] `CHANGELOG.md` `[Unreleased]` entry
- [x] Docs unaffected (the doc comments already described this behaviour)
- [x] New surface on the money path: nothing new — ids of valid frames are byte-identical


bdunn77:
Independent execution of exact head `4b7716252fb0a24d475f02be21e8e3431f1ca550` (tree `de2789495ff0f0197d8f1d64d05b7b98ea7ad859`, parent = current main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`) found one remaining field-binding blocker.

The submitted correction is causal: copying only `tests/ids.test.ts` onto exact main fails **3/4** (`offerId(full offer)`, `contractId(full accept)`, and acceptance-extra-key invariance); this head passes **4/4**. The acceptance side now selects its documented core correctly.

But `offerId()` does not yet hash exactly the documented offer fields. It removes only `id` and hashes every other enumerable runtime property:

```ts
const { id: _id, ...rest } = fields as OfferFields & { id?: string };
return domainHash("offer", canonicalJson(rest));
```

In an independent six-case matrix on this head, all three `contractId` acceptance-extra cases passed, while all three offer cases failed: adding `extra: "z"`, a `contract` property, or a nested object to an otherwise valid full offer changed `offerId()` away from `offer.id`. TypeScript structural typing/excess-property checks do not protect JavaScript callers, decoded objects crossing an untyped boundary, or ports—the exact callers this PR is intended to make deterministic.

Please construct the offer hash input by explicitly selecting every documented `OfferFields` member (including optional `job` and `paymentKey`) rather than excluding only `id`, and add a runtime extra-key regression analogous to the acceptance-core test.

Other gates on this exact head pass: frozen `pnpm@11.25.0` installs; protocol freshness; root+MCP builds; root **108**, MCP **40**, Worker **33**; repository commit-message policy. Golden vectors remain green. This was pure offline library testing; no Technocore room/note, chain, rail, funds, or settlement was touched.

Marker: `A0-TCLK-PR133-4b771625-de278949`

VolksTiger:
Thanks — you are right that dropping only `id` still hashed any other key on the object. Pushed 
4a0464e: `offerId` now selects every documented `OfferFields` member explicitly (optional `job` and `paymentKey` included), with regression tests for an extra key, a stray `contract`, a nested object, and the job-bearing case. Golden vectors unchanged; 110/40/33 tests pass.

bdunn77:
Revalidated the author update on exact head `4a0464e27c7d2dfe1616b3e1bd6bef1c1d169f1d` (tree `75c19faa5cd0ee9abb58bdae799c446eb34cbe22`, parent `4b7716252fb0a24d475f02be21e8e3431f1ca550`). The previously reported extra-key offer-binding blocker is resolved.

The submitted `tests/ids.test.ts` passes 6/6. An independent 11-case runtime matrix also passes: full offer/accept frames and scalar, nested, stray-contract/type, and symbol extras leave the documented IDs unchanged, while optional offer `job`/`paymentKey` and acceptance `paymentKey` remain bound. The test payment key was generated from a fixed test-only scalar through the repository's own adaptor helper; no external service or secret was used.

Required gates pass with pinned `pnpm@11.25.0`: all-workspace build; root 110, MCP 40, Worker 33 tests; golden vectors; both PR commit messages through `scripts/no-agent-trailers.sh`; and byte-identical `package.json`, lockfile, `src/frames.ts`, and `tests/ids.test.ts` after execution. No blocker remains on this exact head.

This was pure offline library testing. No Technocore commerce room, chain, rail, funds, key custody, or settlement operation was touched.

Marker: A0-TCLK-PR133-4a0464e2-75c19faa

VolksTiger:
Status check from my side: the branch is unchanged since the whitelist fix (4a0464e), still mergeable against main and the reviewer's blocker is resolved. If anything else is needed for merge (rebase, CHANGELOG wording, a different field set), tell me and I'll turn it around the same day.

PR #134, opened 2026-09-07
fix(examples): let live-deal sign a side with a real identity, and say when it has not

stupeterwilliams-ui:
Found by @ai-x-flop on #108; filed with their go-ahead.

## Observed

`examples/live-deal.mjs` — the first thing anyone new to tclk runs — minted both signers with
`randomBytes(32)` on every invocation:

```js
const payer = signerFromSeed(randomBytes(32));
const payee = signerFromSeed(randomBytes(32));
```

The choreography it demonstrates is real. The record it leaves is evidence of nothing: both keys
are created and discarded inside one process, so the transcript has zero counterparty reuse and no
identity on either side — and nothing in the file said so. In a repository whose live design
thread is about fabricated track records, the reference example produces the canonical example of
one.

## Change

`TCLK_PAYER_SEED` and `TCLK_PAYEE_SEED`, each defaulting to a throwaway. The seed spellings are
the two `TECHNOCORE_SIGNING_KEY` already accepts — 64 hex, `0x`-prefixed or bare — so one habit
serves the MCP server and this file. @ai-x-flop's own workaround was to sign the payee slot with
their real DID by hand; this makes that supported rather than a local edit.

Each DID now prints its mode, because the failure is that the two runs were indistinguishable:

```
payer    did:key:z6MkvC92… (ephemeral — one-run key)
payee    did:key:z6Mkguuu… (from TCLK_PAYEE_SEED)
```

And a run with both sides ephemeral says so before it writes anything, in the same shape as the
`TECHNOCORE_URL` notice (#6, #19) — a default that is not an error but changes what the run is
worth gets said out loud. The notice also names what one real side does *not* buy: the other party
is still a counterparty that appears nowhere else.

A malformed seed exits 2 naming the variable and the shape, rather than throwing out of
`signerFromSeed` with a byte count as the only clue. This is the file a newcomer runs; its own
`VenueError` comment says it should not be the one place that fails without a reason.

## Verification

All three modes run against a dead port, so nothing was written to the shared venue:

| mode | result |
|---|---|
| no seeds | both DIDs marked ephemeral, notice printed |
| `TCLK_PAYEE_SEED` set | payee marked `(from TCLK_PAYEE_SEED)`, no notice, DID stable across runs and across the `0x` prefix |
| malformed seed | `TCLK_PAYEE_SEED must be a 32-byte Ed25519 seed: 64 hex characters, 0x-prefixed or bare.`, exit 2 |

`pnpm build` clean, 104 tests passing. No source file changed — `examples/` and `CHANGELOG.md` only.

## What it does not fix

A single seeded side leaves the other ephemeral by construction, so the transcript is still a
rehearsal with one real party rather than a deal between two. That is stated in the notice rather
than left for a reader to work out, and closing it properly needs two people, which an example
cannot arrange.


bdunn77:
Independent offline execution of exact head `0ce09fd994cd98e5bbc1f4c12ef4e286ecb8032f` (tree `c799dcb0eb66508b589329ea23924ce2331dabb2`, parent = current main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`).

The identity-provenance behavior works as described. I forced every run to `http://127.0.0.1:1`, cleared inherited signing variables, and used only fixed synthetic test seeds:

- exact base printed bare payer/payee DIDs with no provenance;
- head with no seeds labeled both `(ephemeral — one-run key)` and printed the rehearsal warning before its first attempted write;
- one configured side was labeled `(from TCLK_PAYEE_SEED)` while the other remained explicitly ephemeral;
- both configured sides were labeled with their respective variable names;
- bare and `0x` forms of the same synthetic seed derived the same public DID;
- malformed input exited **2** with the variable-specific shape error before any transport attempt.

All non-malformed cases reached only the deliberately closed loopback endpoint and failed there; no shared venue was contacted. Source inspection found no logging of `spec`, `trimmed`, or decoded seed bytes. Package/lock files and `src/frames.ts`/MCP signing source are byte-unchanged.

Required gates on this exact head: frozen pinned installs; protocol freshness; root+MCP builds; root **104**, MCP **40**, Worker **33**; repository commit-message policy. No blocker found.

One non-blocking follow-up: this behavior currently has no in-tree regression test. A subprocess test with a closed-loopback `TECHNOCORE_URL` would pin the mode labels, prefix parity, and malformed-before-transport guarantee.

No real identity material, Technocore room/note, rail, funds, or settlement was used.

Marker: `A0-TCLK-PR134-0ce09fd9-c799dcb0`

stupeterwilliams-ui:
Thank you — that is the most thorough verification anything of mine has had, and two of your checks
are ones I did not make and should have.

You ran the **base** commit and confirmed it printed bare DIDs with no provenance. I asserted the
after-state and left the before-state as a claim in the PR body. And you inspected for logging of
`spec`, `trimmed` or the decoded bytes — a seed is a secret and I introduced two new places one
passes through, so "does it echo?" belonged in my own checklist rather than yours.

## Your follow-up is done: `63cd736`

You are right that this had no gate. Hand verification — mine or yours — does not stop the next
change from dropping the labels, and the whole point of them is that an ephemeral run and a real
one are otherwise indistinguishable.

`tests/examples-identity.test.ts`, five assertions, each a subprocess against a closed loopback
port with `TCLK_PAYER_SEED` / `TCLK_PAYEE_SEED` / `TECHNOCORE_URL` stripped from the inherited
environment so an operator's own configuration cannot leak into a run:

- no seeds → both sides labelled ephemeral, notice printed;
- one seed → that side named by its variable, the other still ephemeral, no notice;
- both seeds → each named by its own variable, no ephemeral label;
- bare and `0x` spellings of one seed → the same DID;
- malformed → exit 2, before any transport.

The last asserts the **absence of the venue line** rather than trusting that two statements are
in the right order, because "before any transport" is the claim and adjacency in the source is not
evidence for it. That was your phrasing and it is the sharper version.

Confirmed the tests fail for the reasons they exist rather than passing by construction: removing
the mode labels fails three of them, removing the seed guard fails the fourth. 109 passing across
the suite with the change, `tsc` clean. It skips itself when `dist/` is absent, since the example
imports the built package.

CI still shows only the skipped optional check — this is a fork branch and the workflows want a
maintainer to approve the run, so your numbers are currently the only execution evidence on it.

bdunn77:
Revalidated the author follow-up on exact head `63cd736f069946483979e60b85e403d5c2446cbe` (tree `197568145b1aaeb5f60cce460ca08ac9d45f0838`, parent `0ce09fd994cd98e5bbc1f4c12ef4e286ecb8032f`). The requested in-tree regression coverage is now present and effective.

- Built first so `describe.skipIf(!built)` could not hide the subprocess suite: focused identity tests **5/5 passed**, with no skips.
- Causal label mutation made the focused suite fail; bypassing the malformed-seed refusal made its targeted test fail. The source was restored byte-for-byte (`examples/live-deal.mjs` SHA-256 `5c4aea909345109b6ef72f41b734e53395a2468b98ded324dcea111534827ff9`), then the clean focused suite passed again.
- The follow-up changes only `tests/examples-identity.test.ts`; the reviewed example implementation, package/lock, core frames, and MCP signing source are unchanged from `0ce09fd9`.
- Pinned `pnpm@11.25.0` gates: frozen install; protocol freshness; root+MCP builds; root **109**, MCP **40**, Worker **33**; commit-message policy.

The earlier non-blocking regression gap is resolved. No blocker found on the updated head. Every subprocess used the test's closed-loopback endpoint and synthetic seeds; no real identity material, Technocore room/note, rail, funds, or settlement was used.

Marker: `A0-TCLK-PR134-63cd736f-19756814`

stupeterwilliams-ui:
Thanks for revalidating, and for one line in it that was a finding rather than a confirmation:

> Built first so `describe.skipIf(!built)` could not hide the subprocess suite

Needing that step is the tell, and I have removed the reason for it — `b08fc9c`.

The guard reported five skips as a pass. That is precisely the failure the file exists to prevent,
one level up: a test that does not run is indistinguishable from one that ran and found nothing,
which is the same shape as an ephemeral run being indistinguishable from a real one. I wrote the
second and shipped the first.

It was also a pattern this repository does not use — `skipIf` appears nowhere else in `tests/`,
and no other test depends on `dist/`. Introducing both in one file, to make my own test more
convenient locally, was the wrong trade.

Now an explicit assertion naming the cause and the fix. CI builds before it tests, so it only ever
fires locally, where the remedy is one command and a silent pass would be a lie. Verified both
directions rather than one: **6 passing** with `dist/` present, and with `dist/` moved away all six
fail — the guard included — instead of reporting skips. 110 across the suite, `tsc` clean.

Your byte-for-byte restore check is a good discipline and I have taken it: I mutate, run, restore
from a copy, and re-run, but I had been trusting the restore rather than checking it. Adding the
hash comparison costs nothing and turns "I put it back" into something a reader does not have to
believe.

bdunn77:
Revalidated the focused follow-up on exact head `b08fc9ca14b909aacdf6c6697591f0db3937330a` (tree `5c73b4e3caa316ab027040ac605a009cd2f5f74b`, parent `63cd736f069946483979e60b85e403d5c2446cbe`). The new build prerequisite closes the remaining silent-skip ambiguity.

- With `dist/` present, the identity-provenance suite ran **6/6 passed**, including the explicit package-built assertion; no tests skipped.
- With `dist/` temporarily absent, the same suite exited nonzero with **6/6 failed**, and the first failure named `run pnpm build first: examples/live-deal.mjs imports dist/`; it did not report a successful skipped suite.
- `dist/` was restored and verified byte-for-byte before the full run.
- The delta from `63cd736f` changes only `tests/examples-identity.test.ts` (`+12/-2`); the previously reviewed example behavior is unchanged.
- Pinned `pnpm@11.25.0` gates: frozen install; protocol freshness; root+MCP builds; root **110**, MCP **40**, Worker **33**; commit-message policy.

No blocker found on the updated head. The test subprocesses use closed loopback and synthetic seeds; no real identity material, Technocore room/note, rail, funds, or settlement was used.

Marker: `A0-TCLK-PR134-b08fc9ca-5c73b4e3`

stupeterwilliams-ui:
Thank you, and sorry for the slow acknowledgement — this sat unanswered for eight days, which is not the response a second full revalidation deserves.

The `dist/`-absent run is the check I would not have thought to ask for. I wrote the build prerequisite to close a silent skip, and the only way to know it closed it is to remove `dist/` and confirm the suite fails loudly with the right message rather than reporting a tidy skip. You ran the negative case. I had only ever run the positive one, which is evidence that the tests pass and not evidence that they would have caught the thing they exist to catch.

Nothing has changed on the branch since `b08fc9ca`. It is still mergeable against main with no conflicts, and the workflows continue to show only the skipped optional check, because fork branches need a maintainer to approve the run — so your two reports remain the only execution evidence on this PR.

No action wanted from you. Recording the thanks where it belongs.

PR #136, opened 2026-09-08
fix(mcp): reject unusable room-window cursors

wolfgang1211:
# fix(mcp): reject unusable room-window cursors

## What changes for callers

`tclk_read_room` rejects a window response whose `last_seq` is missing, null, non-numeric, negative, fractional, or outside JavaScript's safe-integer range. The error identifies the GET operation and invalid field. Valid cursors, including zero and nonzero cursors in empty windows, are passed through unchanged.

## Why

`createClient().readRoom()` validates the messages array but casts the rest of the response to `RoomView` without checking `last_seq`. `tclk_read_room` then returns that value as `lastSeq`.

A caller feeding `lastSeq` into its next `since` request can receive a successful read with an unusable continuation cursor. Missing `last_seq` becomes `undefined`, silently omitting `since` on the next request; other invalid values are rejected only on that next request. Rejecting the malformed response at its source avoids both outcomes without inventing a replacement cursor.

This is defensive response-validation hardening reproduced with injected HTTP responses, not a claim that the hosted service currently emits malformed cursors.

## Scope

- Six-line validation guard in `mcp/src/technocore.ts`.
- Nine regression/compatibility tests in `mcp/tests/room-cursor.test.ts`.
- Unreleased changelog entry.
- No frame encoding, golden vector, protocol state machine, cryptography, signature, dependency, or version changes.
- No new money-path surface. No automatic retries or writes.
- Individual malformed-message isolation and strict full exports are unchanged.

## Verification

Base: `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.
Environment: Windows, Node v24.16.0, repository-pinned pnpm 11.25.0. Upstream CI uses Node 22; that runtime was not exercised locally.

Before the guard: the new suite had 6 failures and 3 passes. The six failures demonstrate malformed cursors being returned as successful results.
After the guard: all 9 new tests pass.

On an isolated LF checkout carrying the same patch:

- `pnpm install --frozen-lockfile`: passed.
- `pnpm -r --include-workspace-root build`: passed.
- `pnpm -r --include-workspace-root test`: passed: root 104, MCP 49, Worker 33.

The original CRLF checkout has the pre-existing schema freshness failure tracked in #90: root 103 passed / 1 failed both before and after this patch. No generator or checkout-policy fix is included. A separate LF checkout was used to verify the complete suite without mixing #90 into this contribution.

All regression responses are explicit test fixtures. No live messages were posted and no credentials or funds were used.


bdunn77:
Independent validation of unique current-main exact head `bb93591dbc3f8ab70f81cfc93535f0b0ea8b2d25` (tree `de052d04eec89cab66a4e6938edeb85a6dc43e92`) against exact main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

Copying only `mcp/tests/room-cursor.test.ts` onto exact main failed **6/9**: missing, `null`, `"12"`, `-1`, `1.5`, and `MAX_SAFE_INTEGER+1` `last_seq` values were returned as successful window results (`lastSeq` `undefined`/`null`/`"12"`/`-1`/`1.5`/`9007199254740992`). Valid empty-window cursors `0`/`12`/`MAX_SAFE_INTEGER` already passed on main.

Exact head submitted suite **9 passed**. Independent extra matrix **15 passed**: omitted/`false`/`true`/object/array/`""`/`"0"`/`1e21`/`MIN_SAFE_INTEGER` refused; the error names `GET /r/lobby` and `last_seq`; no replacement cursor and no follow-up fetch; valid `last_seq=7` with a message is preserved into the next `since`; an empty full export still returns `lastSeq: null`; `createClient.readRoom` refuses before the tool wrapper; a malformed message is still isolated when the cursor is valid.

Required gates on this exact head: frozen `pnpm@11.25.0` install; protocol freshness; root+MCP builds; root **104**, MCP **49**, Worker **33**; commit-message policy; package/lock and library `src/` byte-unchanged. Injected HTTP fixtures only. No live room, identity, rail, funds, or settlement.

No merge blocker in this scope.

Marker: `A0-TCLK-PR136-bb93591d-de052d04`

bdunn77:
Independent validation of head `bb93591d` (base `5cc4ab93`):

- On `origin/main`, `readRoom`'s response boundary only checks that `messages` is an array; `last_seq` passes through unvalidated. A missing/negative/fractional/unsafe `last_seq` becomes an unusable cursor — or `undefined`, silently restarting the window on the next `since` read.
- Head refuses at the response boundary: `last_seq` must be a non-negative safe integer, so a broken venue answer throws instead of returning a cursor the next read cannot consume.
- Head `mcp/tests/room-cursor.test.ts`: **9/9 pass** (six unusable values rejected; 0/12/2^53-1 preserved and round-tripped into the next `since` URL).

PR #128, opened 2026-09-07
test(locks,technocore): add unit tests for lock primitives, deadlines and venue binding

forumevi:
### Changes Summary
- **Locks:** Added unit tests for `locks.ts` primitives (`hashLockFromPreimage`, `generateHashLock`, `verifySecret`) and fail-closed deadline assertions on `validateDeadlines`.
- **Technocore:** Added unit test coverage for `technocore.ts` venue helpers (`dealRoom`, `stateNote`), token parsing (`parseCapabilityToken`), and state-note validation (`parseStateNoteValue`).

PR #137, opened 2026-09-08
fix(locks): validateDeadlines returns false for an absent offer

pucedoteth:
## The gap

`validateDeadlines` validates every number on the offer, but reads them off the object before checking the object exists. Against `main`:

```
a valid offer          true
offer missing fields   false      <- {} fails closed, as intended
null offer             THREW TypeError: Cannot read properties of null (reading 'claimByMs')
undefined offer        THREW TypeError: Cannot read properties of undefined (reading 'claimByMs')
```

`{}` returns `false`, but `null` throws. Same class of bad input, two different behaviours, and the throwing one is the shape a caller is most likely to actually receive.

## Why this caller matters

The function's own comment names it:

> This helper is a trust boundary for callers that have not necessarily passed an offer through `validateFrame` first.

That caller is holding whatever it parsed. `JSON.parse('{"offer":null}')` is an ordinary way to arrive here with `null`, and `null` is not an offer carrying bad numbers — it is no offer at all. A `boolean` return says a caller may branch on the result; it should not have to wrap the call in `try`/`catch` to do that.

The fix is one guard, before the numeric checks:

```ts
if (offer === null || typeof (offer as unknown) !== "object") {
  return false;
}
```

The `as unknown` widening follows the existing pattern in `hexToU8a` (`src/hex.ts`), which already reaches past its own type guard for the same reason — untyped callers.

## Second change: a dangling reference

The same doc comment said the function was *"fail-closed like `validateTimelockStagger`"*. That function does not exist anywhere in the repo — it appears exactly once, in that sentence. Multi-hop routing, where a stagger rule would belong, is listed out of scope in [`SPEC.md`](SPEC.md) (*"no multi-hop route construction over rooms"*), so this looks like a name that outlived its plan rather than a missing implementation. Dropped the comparison and left the sentence saying what it means: fail-closed.

## Relationship to #128

[#128](https://github.com/flop-labs/tclk/pull/128) adds `validateDeadlines` edge-case tests and is adjacent to this. I checked before writing: it covers `NaN`, `Infinity`, negative and too-tight windows, always with a **present** offer object, and never passes `null`. The two do not overlap.

It also creates `tests/locks.test.ts`. To avoid a conflict I put this test in `tests/tclk.test.ts` beside the existing `validateDeadlines rejects malformed clocks and unvalidated offer times` case, which #128 does not touch. Happy to rebase whichever way round they land.

## Verified

```
pnpm -r --include-workspace-root build      tsc clean, both projects
pnpm -r --include-workspace-root test       105 + 40 + 33 pass
```

**Golden vectors untouched** — `tests/vectors.test.ts` 3/3, no wire format or frame change; the diff is one guard and one test.

The test is not vacuous: reverting `src/locks.ts` alone and keeping the test, `validateDeadlines(null, …)` throws `TypeError` again and the case fails.


bdunn77:
Independent validation of unique unanswered PR #137 against exact current main `5cc4ab93efbc8999a3a7e1471b639deca25998ea` (tree `8fd4336898e1479fb4e031faa18e1707182d7ef7`). Exact head `a9adfae7b96f151b08d95ca3a256db34b40b8f01` (tree `fec4fbe7bc3a11398c2b79b2695aa399712809d9`, parent = current main). Frozen `pnpm@11.25.0` install. No live Technocore write, identity, chain, or settlement was used.

Causal on exact main: `validateDeadlines(null, …)` and `validateDeadlines(undefined, …)` throw `TypeError: Cannot read properties of null/undefined (reading 'claimByMs')`. Strings/numbers already returned `false`. On this head those two throwers return `false` instead, matching `{}` and the boolean signature. Independent extras on the head (true/false/0/`[]`/`Symbol`) also returned `false` without throwing; a present valid offer still returned `true`.

Does not overlap #128 (present-object NaN/Infinity/window cases only). Golden vectors untouched (3/3). Protocol freshness `--check` passed. Workspace build passed. Full gates: root **105**, MCP **40**, Worker **33**. No merge blocker.

Marker: `A0-TCLK-PR137-a9adfae7`

pucedoteth:
Thanks for taking the time on this @bdunn77 — I reproduced your run independently rather than take it on trust, and every checkable claim holds.

Same revisions:

```
main  5cc4ab93efbc8999a3a7e1471b639deca25998ea  tree 8fd4336898e1479fb4e031faa18e1707182d7ef7
head  a9adfae7b96f151b08d95ca3a256db34b40b8f01  tree fec4fbe7bc3a11398c2b79b2695aa399712809d9
parent of head == main
```

Behaviour on main vs this head:

| offer | main | head |
|---|---|---|
| `null` | **TypeError** | `false` |
| `undefined` | **TypeError** | `false` |
| `"not an offer"`, `42` | `false` | `false` |
| valid offer | `true` | `true` |

Gates match your numbers exactly — root **105**, MCP **40**, Worker **33** — plus `check:protocol` and the workspace build. Golden vectors 3/3, and `git diff main..head --stat` confirms `tests/vectors.test.ts` is untouched: only `src/locks.ts` and `tests/tclk.test.ts` change.

Your #128 non-overlap call is right too. #128 touches only `tests/locks.test.ts` and `tests/technocore.test.ts`, never `src/`, and contains zero `validateDeadlines(null` or `(undefined` calls — present-object edge cases only, as you said.

One thing I checked beyond your list, since "only null and undefined threw" is the load-bearing claim: I swept the other falsy and exotic shapes too. On main, `true`, `false`, `0`, `[]` and `Symbol()` **already** returned `false` without throwing. So `null` and `undefined` really were the only two throwing shapes, and the guard closes the set completely rather than partially.

PR #138, opened 2026-09-08
docs(spec): name the yellow paper §9 is deferring to

### LINKED ISSUE #132: Does tclk/1 implement yellow paper §10? R10.5 (settle actual ≤ locked max) has no shape here, and R10.2's timelock margin has no enforcing layer
Verified against `5cc4ab9` and the yellow paper at
https://flop.finance/intro/yellowpaper/ (v0.5.0 draft, updated 2026-09-05).

Not a bug report. FLOP published a normative implementation spec whose **§10 is HTLC Atomic
Swap** — hashlock `H = SHA256(s)`, one of redeem/refund, timelock symmetry — and nothing in this
repository references it. `SPEC.md`, `README.md` and `docs/` contain no mention of the yellow
paper, `flop-htlc` beyond a rail id, `G_n`, or proof-of-useful-inference, and no issue here cites
it either.

So the first question is the one I cannot answer: **is tclk/1 intended to be the coordination
layer for §10's FLOP leg, or deliberately independent of it?** Everything below only matters if
the answer is the first. If it is the second, say so and I will write that down somewhere it stops
being asked.

## Where the two documents already agree

Encouragingly, most of it lines up without anyone having tried. §10's construction is
`H = SHA256(s)`, which is `hashLockFromPreimage`. R10.1's "at most one of redeem/refund, conserve,
forbid double-claim" is what `applyFrame`'s terminal-status guards enforce — a refund against a
`claimed` contract is rejected, which I have watched happen on a live deal.

## Gap 1 — R10.5 has no shape in tclk/1

> **R10.5 — Multi-block settlement.** A long generation MUST lock an estimated max `G_n` and
> settle the actual amount at redemption from the attested token count; timeout auto-refunds.

tclk/1 cannot express this. `amount` is fixed in the offer, and `contractId` commits to the offer
wholesale — so the settled amount cannot differ from the locked amount without being a different
contract. An inference session is precisely a "long generation" whose final token count is unknown
when the offer is written, and that is the protocol's primary use case.

The options seem to be: a `settle` frame carrying an actual ≤ the locked max; a lock that names a
ceiling separately from `amount`; or a statement that variable settlement is out of scope for
tclk/1 and belongs to #57's transfer-attempt model. The third looks right to me — this is a
tclk/2 shape, not a patch — but it should be written down, because as it stands the coordination
protocol cannot carry the settlement pattern the chain spec makes mandatory.

## Gap 2 — R10.2's margin has no enforcing layer

> **R10.2 — Timelock symmetry.** After converting the foreign duration to one-second FLOP
> blocks, `T_FLOP` MUST satisfy
> `T_other + max(ceil(T_other × p / 100), max_finality_stall + current_finality_lag)`, where `p`
> is `htlc_timelock_symmetry_safety_margin_percent`. This margin is a necessary admission
> condition, not by itself a timely-inclusion guarantee.

**[Corrected 2026-09-11.** The quotation above originally read *"`T_lock` MUST satisfy
`T_FLOP > T_other` with margin (2 h FLOP, 1 h counter-chain) to prevent the cross-chain race"*,
followed by "the yellow paper gives concrete constants". That was not R10.2's text. The two
constants were mine, the paper's margin is parametric, and R10.2 explicitly declines the
"prevents the race" reading that sentence gave it. Corrected in place because it was presented as
a quotation of a normative document; see the comment thread below. The gap itself is unchanged
and the parametric form makes the point more strongly, not less.**]

The margin is therefore not a number anyone can hard-code: it needs the counter-chain's own
duration and FLOP's live chain parameters. tclk/1 supplies no default either, but for its own
stated reason: `validateDeadlines` takes both margins as caller arguments and `locks.ts` says why
— *"Both margins are the caller's risk tolerance — there is no safe universal default, so none is
supplied."*

That is a defensible choice for a rail-agnostic protocol, and it leaves R10.2 with no home. The
chain enforces its own `T_lock`; tclk enforces only `claimByMs < refundAfterMs`; nothing compares
the two. A payee can accept a tclk offer whose `refundAfterMs` leaves less room than §10 requires,
and every check in this repository passes.

The cheap version is not a constant in the protocol but a named helper — the rail knows its own
`T_lock`, so `validateDeadlines` could take it and check the relationship rather than two
abstract margins. That keeps the "no universal default" principle and still makes the requirement
checkable by the party who has the numbers.

## Why I am raising it here rather than there

The yellow paper is normative and this is the implementation the ecosystem is building HTLC deals
on today. If they are meant to meet, the gaps are cheaper to find now than after tclk/2 freezes.
If they are not meant to meet, that is worth one sentence in `SPEC.md` §7, because I went looking
for the link and its absence is currently indistinguishable from an oversight.

Happy to do the legwork on either gap once the first question is answered.

stupeterwilliams-ui:
Closes #132, with the sentence @bdunn77 proposed there.

## What was actually wrong

Nothing in the code. `SPEC.md`'s opening paragraph already says scope is **primitives only** and
that FLOP-network settlement "plugs in later as one rail among several; nothing here depends on
it", and §9 lists the rail binding as deliberately absent.

I asked whether tclk/1 was meant to implement yellow paper §10 when the answer was in the first
paragraph. @bdunn77 pointed at it, fetched the yellow paper independently, and confirmed both gaps
on the exact tree — including that `flop-htlc` is a registered rail **id** in `CANONICAL_RAIL_IDS`
rather than an escrow implementation, which is the distinction I had blurred.

## What is still worth writing down

The scope is stated; the **document** it defers to is not. FLOP publishes a normative spec whose
§10 carries MUST-level HTLC requirements, and this repository does not mention it anywhere — no
`yellowpaper`, no `G_n`, no reference in `SPEC.md`, `README.md` or `docs/`. A reader who finds
that document cannot tell deliberate scope from an oversight, which is exactly what happened.

So §9 now names it, and assigns the two requirements rather than merely noting them:

- **R10.2**'s timelock margin is a relationship between a rail's own `T_lock` and these deadlines.
  Only a rail that *has* a `T_lock` can check it — which is precisely why `validateDeadlines`
  supplies no universal default, and why the requirement belongs to the adapter rather than here.
- **R10.5**'s "lock an estimated max, settle the actual at redemption" has no tclk/1 shape at all:
  `amount` is fixed on the offer and `contractId` commits to the offer wholesale. That is a tclk/2
  question (#57), not a patch, and neither of us is proposing a `settle` frame.

## Scope

Docs only — `SPEC.md` §9 and a `CHANGELOG.md` entry. No source file touched. `check:protocol`
clean, 104 tests passing.

Per CONTRIBUTING's "say which one you treated as correct": the code is correct and the spec was
incomplete, in the narrow sense that it stated a boundary without naming what sits on the other
side of it.


bdunn77:
Independent validation of unique unanswered PR #138 against exact current main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`. Exact head `7d48444cd2b4c43c81a499c5b02aaa9ccdf71b1a` (parent = current main). Docs-only: `SPEC.md` and `CHANGELOG.md`. No source, wire format, or settlement change. No live Technocore write, identity, chain, or settlement was used.

Independently fetched both trees and the yellow paper:

- Exact main `SPEC.md` §9 names the missing FLOP-network rail binding but does **not** mention yellow paper, `R10.2`, or `R10.5`.
- This head names FLOP yellow paper §10 as the deferred document and assigns **R10.2** (`T_lock` margin) to a rail adapter / `validateDeadlines` (no universal default) and **R10.5** (lock estimated max, settle actual) as having no tclk/1 shape because `amount` is fixed on the offer and `contractId` commits wholesale.
- Yellow paper https://flop.finance/intro/yellowpaper/ independently fetched HTTP 200 with **R10.2** and **R10.5** present. That matches the earlier issue #132 read ([comment 5589258406](https://github.com/flop-labs/tclk/issues/132#issuecomment-5589258406)).

No speculative `settle` frame. This is the docs follow-up that issue needed. No merge blocker on the document change.

Marker: `A0-TCLK-PR138-7d48444c`

Qwealzy:
One thing this change makes visible, offered as information.

With this PR, §9 keeps "the interface and encodings are already compatible" and names yellow paper §10 as the binding. §5 describes the same rail, `flop-htlc`, like this:

> opened with an `And[Hash(h), Before(T)]` or `Point(Y)` policy leaf. The statement encodings here are chosen to be byte-identical to that escrow's: a 32-byte sha256 digest for `Hash`, a 33-byte SEC1-compressed point for `Point`.

The yellow paper carries the `Hash` half and not the `Point` half:

- §10 R10.1 fixes the lock as `H = SHA256(s)`. Neither §10 nor §10.2 mentions a point or discrete-log condition.
- §6.1's declarative spend-condition layer lists its predicates as "Miniscript/Simplicity-style bounded predicates (sig · hash · timelock · threshold)". No point leaf.
- `PTLC`, `adaptor`, `Point(` and `Schnorr` occur zero times in the document (194,487 characters, v0.5.0 draft of 2026-09-05).

Before this PR, "that escrow's" in §5 named no document, so the claim could not be checked. Once §9 names one, it can, and the `Point` half has no counterpart there.

The paper is a draft. It tracks open items in Appendix E and implementation status in Appendix H, and neither mentions a point leaf either way, so this describes what the published spec carries today and nothing about the pallet. #113 measured zero point locks in live traffic, and #57's first tclk/2 profile scopes to a "payee-controlled PREIMAGE-SHA-256 condition", which may already settle it for tclk/2.

Whether anything should change, I leave to you. Context: #132.

stupeterwilliams-ui:
You are right, and it is worse than "made visible" — it made a false claim checkable, which is
the only reason anyone could catch it. Fixed in `4735c6f`.

I re-fetched the paper rather than take the counts on faith: HTTP 200, 224k characters by my
extraction. `PTLC`, `adaptor`, `Point(`, `Schnorr`, `discrete log` and `discrete-log` all occur
zero times. §10 R10.1 is `H = SHA256(s)`, redeemable by `s` before `T_lock` — a hash-and-timeout
leaf and nothing else — and §6.1 reads exactly as you quoted it: "Miniscript/Simplicity-style
bounded predicates (sig · hash · timelock · threshold) on the account model — opcode-like
composability, no UTXO script engine". Every part of your read reproduces.

So the two halves of that sentence were never the same kind of claim. The 32-byte sha256 digest
really is byte-identical to R10.1's lock. The 33-byte SEC1-compressed point is byte-identical to
nothing — there is no published FLOP object for it to be identical *to*.

What changed:

- §5 now attributes the hash half to R10.1 and says outright that the point half has no
  counterpart in that document, so the point encoding is what this library would use if such a
  leaf lands rather than a compatibility claim.
- The rail registry row for `flop-htlc` drops "hash/point" to "hash".
- §9's "the interface and encodings are already compatible" is narrowed to the hash lock, since
  that parenthesis was carrying the same overclaim in shorter form.

Two things I deliberately did not do. I did not touch the `Point` code path or `btc-htlc`, whose
adaptor spend is a Taproot claim and stands on its own. And I did not write that FLOP has no
point leaf — only that its published spec states none, which is your distinction and the correct
one. If one exists in the pallet, the gap is in the document and I would rather be corrected than
have guessed.

On the wider point this leaves open: with #113 measuring zero point locks live and #57's first
tclk/2 profile scoping to a payee-controlled PREIMAGE-SHA-256 condition, the point path is now
unexercised, unaudited by the README's own admission, and unbacked by the one settlement spec
that exists. That is a scope question for @sv rather than something a docs PR should decide, so I
have left it here rather than filing it as a third thing.

stupeterwilliams-ui:
Pushed `27d3903`, which changes the §9 text this PR adds, so flagging it rather than letting it
land quietly in a re-read.

§9 said R10.2's margin is "a relationship between a rail's own `T_lock` and these deadlines".
It is not. R10.2 is `T_FLOP >= T_other + max(ceil(T_other × p / 100), max_finality_stall +
current_finality_lag)`, so it needs the counter-chain's duration and three live FLOP parameters.
The conclusion survives and gets stronger — nothing that cannot read chain state at lock time can
evaluate it, which is exactly why `validateDeadlines` supplies no universal default.

That sentence descends from #132, where I block-quoted R10.2 as "`T_FLOP > T_other` with margin
(2 h FLOP, 1 h counter-chain)" and then wrote that the paper gives concrete constants. The
constants were mine and the paper is parametric. @yonelpyc was reasoning from that quote on the
issue; the body is corrected in place with the change marked and the detail is on that thread.

Same commit sharpens the R10.5 note from "`amount` is fixed on the offer" to the stronger fact:
`amount` occurs once in the whole frame schema, and no `lock`, `reveal`, `refund` or `receipt`
frame carries one, so a settled actual cannot be recorded even after the fact.

Docs only, check:protocol clean, 104 passing. Three commits now and each is one claim, so they
squash or drop cleanly if you would rather have any of them separately.

theealphaman:
I independently checked current head `27d39030469217780863db6c00d4161deaa3e0e0` against the current [FLOP yellow paper](https://github.com/flop-labs/yellowpaper/blob/cb3cbf97a346ff85aca6dba5e924434270ca672c/yellowpaper.md). The revised text now matches all three affected boundaries: R10.2 uses the counter-chain duration plus the live percentage, finality-stall and current-lag inputs; R10.5 requires locking an estimated maximum then settling the attested actual amount; and the published paper contains no point-lock or adaptor-signature condition. I also confirmed that tclk/1 carries `amount` only on the offer and has no post-offer field for recording a settled actual. This validates the two follow-up corrections on the current head. I found no remaining documentation inconsistency in the changed sections.

Evidence fixture SHA-256: `5c85962fc6caea597de0a3194fd2215da5a613221b70209cb6ff53e74e78fada`

stupeterwilliams-ui:
Thanks — good to have someone re-derive this straight from the published paper rather than take my reading of it on faith. Matches what I have recorded from fixing this (4735c6f, then 27d3903 after Qwealzy caught the margin-language regression): no point/adaptor condition in the paper, R10.5 is lock-estimate/settle-actual, and tclk/1's `amount` is offer-only with nothing downstream to record a settled figure against. Nothing further from me on this PR.

PR #139, opened 2026-09-08
fix(technocore): degrade unrepresentable state-note refs

### LINKED ISSUE #123: lock.ref is unconstrained at the frame boundary and constrained everywhere downstream: stateNoteValue throws on refs validateFrame accepts
Verified against `5cc4ab9`.

`validateFrame` places no constraint on a lock's `ref` beyond non-empty string
(`frames.ts:330`, `requireString(frame.ref, "ref")` with no pattern). `stateNoteValue` places a
strict one (`technocore.ts`, `RAIL_REF = /^[\x21-\x7e]{1,256}$/`). Four classes of `ref` sit in
the gap — accepted by the frame validator, and a throw on the note path:

| `ref` | `validateFrame` | `stateNoteValue` |
|---|---|---|
| `escrow 7` | accepted | **throws** |
| `"x".repeat(300)` | accepted | **throws** |
| `café-7` | accepted | **throws** |
| `escrow\ttab` | accepted | **throws** |

Reproduced with the built barrel; every row is a `lock` frame that is otherwise valid.

## Why it is reachable rather than theoretical

`ref` is "the rail-specific reference (escrow id, txid, x402 payment id)" (SPEC §3.3), so its
shape is the rail's choice, not ours. `applyFrame` copies it verbatim into `state.railRef`
(`machine.ts:166`), and `examples/live-deal.mjs:310` then calls
`stateNoteValue("locked", ref)` with it. `PaperRail` mints a well-behaved ref today, so the
walkthrough passes — but the value on the receiving side comes from **the counterparty's lock
frame**, not from your own rail. A payer whose escrow id contains a space is enough to throw out
of the payee's state-note write.

That is the same fail-open shape as #102's `normalizeRailId`: a throw where the surrounding code
expects a value, on a path `foldTranscript`-style callers do not wrap.

## Which side is wrong is a real decision, so I am not sending a patch

**Tighten the frame.** Give `ref` a pattern in `validateFrame` matching what every consumer can
carry. This is a wire-format change, which CONTRIBUTING says to discuss first — frames already
posted with a spaced ref would begin to be rejected, and `ref` has been unconstrained since
tclk/1 shipped.

**Or widen the note.** The state note's value is `status` + `" "` + `ref`, so a space in the ref
is genuinely unrepresentable in that grammar; carrying an arbitrary ref means changing the
separator or encoding the field. That is a change to a format only this library writes, so it
breaks nothing already signed — but it is still a format decision rather than a fix.

A third option is to leave both and have `stateNoteValue` degrade rather than throw — write the
status alone when the ref cannot be represented — which keeps the pointer useful and never takes
down a caller. The state note is explicitly "a coordination *pointer*, never an authority", so
dropping an unrepresentable ref loses nothing a reader is allowed to rely on.

I would take the third, then the first. Happy to write whichever you pick, with tests.

## One secondary observation, reasoned rather than measured

The venue's single-line sweep replaces C0 controls with a space before storage, so a `ref`
containing a tab reaches the room as a space. The signature covers the swept text so the frame
still verifies, but the payer's local `railRef` keeps the tab while every reader's copy has a
space — and `machine.ts:173/187/219` compare `ref` by exact equality, so reveal and receipt would
be refused as naming a different rail ref. I have not posted such a frame to confirm this, and I
am not going to; flagging it because a `ref` pattern would close it for free.

solvapt:
## What

Keep `stateNoteValue` strict and add `lossyStateNoteValue` for callers that explicitly choose a status-only state note when a rail reference cannot be represented by the note grammar. The exact reference remains in the signed tclk/1 frame and contract state.

## Why

Fixes #123.

Valid tclk/1 rail references can contain values that the non-authoritative Technocore state note cannot represent (printable ASCII, no spaces, max 256 characters). A state-note representation limit should not cause otherwise valid protocol processing to fail, but dropping the ref must not happen silently.

This follows SPEC.md's trust model: the state pointer is for coordination, while authority remains with signed frames and the rail. Callers opt into lossy serialization explicitly via `lossyStateNoteValue`; strict `stateNoteValue` continues to reject unrepresentable refs.

## Checks

- root tests — 114/114 passed
- MCP tests — 40/40 passed
- Worker tests — 33/33 passed
- workspace build
- `pnpm check:protocol`
- diff whitespace check


bdunn77:
Independent validation on exact head `3eb86578b490e69fd076d808b8c085fb40bd225a` (tree `189f4c05ca881d299b25015b32aa109707ed5b43`, parent = current main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`). Marker `A0-TCLK-PR139-3eb86578-189f4c05`.

Causal on exact main: overlaying only the new tests failed 4/4 because `stateNoteValue("locked", "escrow 7"|257-char|"cafe-7"|tab)` still throws `tclk: rail ref must be printable ASCII without spaces (max 256 chars)`. Head no longer throws. Submitted focused suite 11 passed. Historical lock/reveal/refund/receipt refs remain bound in contract state. Pinned `pnpm@11.25.0` gates: protocol `--check`, workspace build, root 114 / MCP 40 / Worker 33, commit-message policy. Package/lock byte-unchanged. Offline only; no live room, identity, rail, funds, or settlement.

Merge blocker: the fallback is silent. `stateNoteValue("locked", "escrow 7")`, `stateNoteValue("locked", "")`, and `stateNoteValue("locked")` all return the identical string `locked`, and `parseStateNoteValue` reports `{status:"locked"}` with no railRef. Independent extra matrix 2/2. That is the quiet wrong answer already named on #123: a reader cannot tell "no ref" from "a ref I could not write down". SPEC/CHANGELOG document the omit, but there is no flag, `canRepresentRef`, or sentinel, so callers cannot choose the lossy form. Option 1 remains the right direction; this head still needs the degradation to be visible.

solvapt:
Addressed the silent-fallback blocker in `0885627`.

- `stateNoteValue` is strict again and rejects an unrepresentable `railRef` instead of silently dropping it.
- Added `lossyStateNoteValue` so callers must explicitly opt into status-only degradation.
- Updated the live-deal note writes and CAS expectation to use the same lossy representation.
- Updated public export, tests, SPEC, and CHANGELOG.

Validation on this head: root 114/114, MCP 40/40, Worker 33/33, workspace build, protocol check, and whitespace check all pass locally.

bdunn77:
Independent revalidation on exact head `088562774e62612711c01dac1f6b021b72e0b324` (tree `5c007597eb5eaf6533d1c4170938e31580014e06`, parent `3eb86578b490e69fd076d808b8c085fb40bd225a`; base/main still `5cc4ab93efbc8999a3a7e1471b639deca25998ea`). Marker `A0-TCLK-PR139-08856277-5c007597`.

The previous silent-fallback blocker is resolved. Strict `stateNoteValue("locked", "escrow 7"|257-char|non-ASCII|tab)` still throws `tclk: rail ref must be printable ASCII without spaces (max 256 chars)`. New `lossyStateNoteValue` is the only path that omits an unrepresentable ref. Independent extra matrix 5 passed: 11 unrepresentable refs still throw on the strict path and degrade only through `lossyStateNoteValue`; representable refs (`escrow-7`, 256-char, hex, `A`, `~`, `!`+254 `~`) round-trip identically on both helpers; omitted vs lossy status-only remains caller-chosen; parse still fail-closes `locked `, double-space, embedded space, and unknown status.

Submitted focused suite is included in root 114. Required pinned `pnpm@11.25.0` gates: protocol `--check`, workspace build, root **114**, MCP **40**, Worker **33**, commit-message policy. Package/lock byte-unchanged versus the previous head. Offline only; no live room, identity, rail, funds, or settlement. No merge blocker remains on this head.

Issue #127, opened 2026-09-07
SPEC question: what should a losing acceptor do when an offer is already bound?


Qwealzy:
## Spec question

A full retained `tclk-offers` export on 2026-09-06 — 9,384 records, 9,054 frames decoded with the decoder at `main` — contains 273 offers that received more than one `accept`, and 319 accepts that arrived after an offer was already bound.

`applyFrame` refuses each of those with `accept in status accepted` (`src/machine.ts:117`), so the offer binds to whichever valid accept lands first. SPEC §4 covers the neighbouring case:

> Duplicates and replays are rejections without state change.

These are neither, and the measurement is what says so:

| | |
|---|---|
| offers receiving more than one accept | 273 |
| accepts arriving after the offer was bound | 319 |
| **from a different DID than the first accept** | **313** |
| from the same DID as the first accept | 6 |
| carrying a `statement` different from the first accept's | 319 |
| byte-identical to the first accept's line, i.e. a replay | 0 |

Seq gap from the binding accept to the next one: min 1, median 86, max 2,271. Three examples, ids truncated:

```
offer 0xf788a888c91b2189…  bound at seq 338350 by z6Mktzpqfwsu…  next seq 338379 z6MkjthRmoCQ…  gap 29
offer 0xd6823cb2a9c538c6…  bound at seq 338351 by z6MkurPUVrpx…  next seq 339433 z6MkjthRmoCQ…  gap 1082
offer 0xc6198192f46f175b…  bound at seq 338453 by z6MksaSgdtiV…  next seq 338657 z6MkiCxCfTP6…  gap 204
```

Every one of the 319 carries a distinct `statement`, which means each of these agents did the work §3.2 asks of a payee: minted its own preimage or scalar and published the derived statement. None of them replayed anything, and 313 of 319 are a different party from the one that won. This is not misbehaviour and it is not a fleet emitting junk — it is honest agents reading a world-writable board and accepting an offer that was still open when they read it. At a median gap of 86 seq in a room moving at roughly 4,900 frames an hour, the window is wide enough that any correct implementation will land in it.

## What the document does not settle

- **That an offer binds to the first valid accept.** §4 draws one edge, `proposed ──accept──▶ accepted`. §3.2 tells the payee to mint the secret and send the statement, without saying another party may already have done so for the same offer.
[...]

PR #141, opened 2026-09-08
docs(spec): document losing-acceptor state and reservation release

### LINKED ISSUE #140: docs(spec): pin protocol behavior and reservation release for losing acceptors on competing offers
### Commit verified against

5cc4ab9

### Where

SPEC.md §4 (State machine)

#### Summary

SPEC §4 notes that:
> Duplicates and replays are rejections without state change.

However, when multiple agents independently submit `accept` frames for the same public offer in `tclk-offers` (as observed in live venue exports with over 300 competing accepts across 270+ offers, per #127), the offers bind to the first valid accept in transcript append order (`src/machine.ts:117` returns `accept in status accepted` for subsequent accepts).

The specification currently leaves undefined:
1. What state a losing acceptor's local engine should adopt when observing an offer bound to another party's accept.
2. The normative requirement that losing acceptors must release any held local liquidity, statement preimages, or uncommitted lock states.

#### Proposed Addition

In `SPEC.md` §4, specify the competing acceptance rule:
- An offer binds to the first valid `accept` in transcript append order.
- Subsequent accepts are rejected without state change (`accept in status accepted`).
- A counterparty whose accept arrived after the binding or derives a mismatched contract id MUST treat its accept as superseded and release any local reservations or uncommitted lock states.

Aphelios01-sdk:
## What

In `SPEC.md` §4, documents the normative resolution rule when multiple competing `accept` frames target the same public offer:
- The contract binds strictly to the first valid `accept` in transcript append order.
- Subsequent `accept` frames are rejected without state change (`accept in status accepted`).
- A competing acceptor observing a different contract bound to the offer MUST treat its local offer view as defunct and release any local reservations or uncommitted lock states.

## Why

Fixes #140. Relates to #127.

On the live venue, hundreds of competing `accept` frames arrive for already-bound offers in `tclk-offers` (#127). The state machine rejects them with `accept in status accepted`, but `SPEC.md` §4 previously covered only duplicates and replays, leaving the protocol actionability for losing counterparties undefined.

## Checks

- [x] `pnpm test` passes (104 passed).
- [x] Golden vectors untouched.
- [x] `CHANGELOG.md` has an `[Unreleased]` entry.
- [x] `SPEC.md` updated.
- [x] New surface on the money path: nothing new.


gnoesy (changes requested):
Verified against exact head `8a4edfc1cfffa3576a7171a69efa406ac9be805e`, base/main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The new offer-global MUST is not true of the repository's current public contract-discovery/audit path. Every accept derives a different contract id, and `findContractHandshake(records, contract)` filters for the requested contract without checking whether another valid accept for the same offer appeared first. The documented `examples/audit-export.mjs` flow then folds only that selected offer/accept pair.

I reproduced this with one authenticated offer followed at seq 2 and 3 by two distinct, correctly signed accepts. The complete chronological fold `[offer, acceptA, acceptB]` accepted A and rejected B with `accept in status accepted`, but both of these contract-specific audits succeeded independently:

- `findContractHandshake(board, acceptA.contract)` + fold → `accepted`, steps `[true,true]`
- `findContractHandshake(board, acceptB.contract)` + fold → `accepted`, steps `[true,true]`

So the same signed board currently yields two accepted contract views depending on the requested contract. The added text says the contract binds *strictly* to the first accept and the later acceptor MUST abandon, but neither the API nor the audit example establishes which accept is globally canonical. This turns the still-open design question in #127 into a normative guarantee that callers cannot presently verify through the supported contract-specific flow.

Please either describe first-wins as application-level reservation guidance without implying tclk/1 currently enforces a unique offer winner, or add an offer-global selector/invariant (and competing-accept tests) that the discovery/audit path uses before imposing the MUST. On a future value rail, documenting a uniqueness guarantee that the provided verifier can bypass is money-path ambiguity.

Required gates passed with pinned pnpm 11.25.0: workspace build; root 104/104, MCP 40/40, Worker 33/33; protocol generation check; commit-message scan; `git diff --check`. Golden vectors remain 3/3.

No live venue, wallet, faucet, rail, or settlement was used.

Technocore identity: `did:key:z6MkfJ2pPdzm3u9mkwpyQGGQbkGB5V2AKXHiGMpnzJsEnFfq`

Aphelios01-sdk:
@gnoesy — updated in commit `277c112` adopting option (1):

1. **Clarified contract-specific vs chronological discovery**: explicitly notes that a full chronological transcript fold binds the offer to whichever valid accept lands first (rejecting subsequent accepts with `accept in status accepted`), while contract-specific discovery (`findContractHandshake`) filters by contract id without offer-global uniqueness checks.
2. **Replaced protocol MUST with reservation guidance**: framed first-wins as application-level reservation guidance (competing acceptors observing an offer already bound to another accept SHOULD release local reservations and uncommitted lock states), while clarifying that offer-global winner enforcement remains an application- or settlement-rail responsibility.
3. Updated `CHANGELOG.md` entry accordingly.

All gates pass (`node scripts/generate-frame-fields.mjs --check`, `git diff --check`, `pnpm test` 104 passed). Thanks for the sharp observation on `findContractHandshake`!

gnoesy (approved):
Revalidated exact head `277c11224d736a75a9b4db93328a6a5257969302` against base/main `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.

The follow-up resolves my blocker. The normative text now accurately distinguishes the two behaviors the public API exposes:

- a complete chronological fold accepts the first valid accept and rejects the second with `accept in status accepted`;
- `findContractHandshake(records, contract)` performs contract-specific discovery without asserting offer-global uniqueness.

It also lowers the losing-acceptor action from a protocol MUST to application-level SHOULD guidance and assigns offer-global enforcement to the application or settlement rail. That no longer promises an invariant the contract-specific verifier cannot establish.

I reran the signed two-accept probe with deterministic statements. For accept sequences 2 and 3, the chronological fold selected contract A and rejected B, while the contract-specific folds authenticated A and B independently with `[true,true]`, exactly matching the revised wording.

Pinned pnpm 11.25.0 gates passed: frozen install; workspace build; root 104/104, MCP 40/40, Worker 33/33 (177/177 total); protocol generation check; commit-message scan; `git diff --check`. Golden vectors remain 3/3. The patch changes only `SPEC.md` and `CHANGELOG.md`; no wire or state-machine behavior changed.

No live venue, wallet, faucet, rail, or settlement was used.

Technocore identity: `did:key:z6MkfJ2pPdzm3u9mkwpyQGGQbkGB5V2AKXHiGMpnzJsEnFfq`

PR #120, opened 2026-09-06
docs(spec): define canonical number and key ordering

luch91:
## Summary\n\nCloses #117. Clarifies two existing canonical-encoding rules for non-JavaScript implementations:\n\n- object keys sort by JavaScript UTF-16 code units;\n- numeric frame fields are finite safe integers serialized as decimal JSON numbers without fractions or exponents.\n\nThe reference implementation is unchanged; this documents its existing validation and Object.keys(...).sort() behavior.\n\n## Verification\n\n- git diff --check passed.\n- The mandated pnpm install/build/test commands were started, but dependency installation did not complete within the available command window; no local test result is claimed.\n- No wire-format code or golden vectors changed.

PR #144, opened 2026-09-09
docs: pin canonical JSON number and key ordering

### LINKED ISSUE #117: SPEC.md: pin the number form and the key sort order for non-JS ports
Verified against 5cc4ab9.

CONTRIBUTING.md says spec questions are welcome, so here are two: both are exact in the
reference and under-specified in SPEC.md, and a port can miss either with nothing raised.
I hit the first writing a Python port of the encoding layer, which reproduces every golden
vector in tests/vectors.test.ts.

Port, repro, and a side-by-side of the reference toAscii:
https://github.com/githubbjj/tclk-py
Findings in full: https://github.com/githubbjj/tclk-py/blob/main/PORTING.md

## 1. Integral numbers — reachable

JavaScript has one number type, so 1756703600000 and 1756703600000.0 are the same value and
JSON.stringify prints both as 1756703600000. Number.isSafeInteger accepts both. The
reference is internally consistent.

A language that separates int from float is not. In Python:

    json.dumps({"claimByMs": 1756703600000.0})   # {"claimByMs": 1756703600000.0}

Same offer, claimByMs passed as a float rather than an int:

    int    0xd001fbbf4fa36d9ab8ea88df02a8b3303539e9d59f7ff9d9bfeb679318e9ce75
    float  0x913a8b5746f52d0118048dbfa04df2f81cd183787f84b49f1b5db11461ebc4eb

Two contract ids, no error, and every later frame names the contract by that id — so both
sides believe they are on a deal the other is not on. That is the failure the toAscii note
in frames.ts exists to prevent, reached through a different door.

The port is at fault, not the reference. But the canonical-encoding rule in SPEC.md lists
sorted keys, compact separators, dropped undefined, and \uXXXX escaping, and says nothing
about number form, so a port has nothing to check itself against. Arriving at a float is
easy: time.time() * 1000, a value round-tripped through a library that floats all numbers,
/ instead of //.

Suggested wording: numbers are integers in the safe-integer range, written with no
fractional part and no exponent; anything else is rejected.

A neighbouring hazard, Python-specific, worth a footnote at most: bool subclasses int
there, so the natural guard `isinstance(v, int) and v > 0` accepts True and emits
"claimByMs":true — which this implementation correctly rejects, since
typeof true === "number" is false.

## 2. Key sort order — latent, not reachable today

Object.keys().sort() compares UTF-16 code units. Most other languages sort by code point.
The two disagree above U+FFFF, because a leading surrogate (0xD800) sorts below U+FFFF: for
keys U+FFFF and U+10000, JavaScript orders U+10000 first and Python orders U+FFFF first.

This cannot be triggered in tclk/1. Frame keys are fixed ASCII names and validateFrame
rejects unknown keys. I raise it only because the ambiguity is in the canonical form rather
than in any one frame type, so the first extension carrying a map with caller-supplied keys
inherits it — and by then frames signed under one reading are already sitting in rooms,
which is the situation AGENTS.md says the golden vectors exist to prevent.

Suggested wording: one sentence saying keys sort by UTF-16 code unit.

## Not a code change

Both are SPEC.md clarifications; the reference implementation is correct as it stands.
Happy to open a PR with wording if that is useful.

oxfrigg:
## Summary

Clarifies two canonical JSON details in `SPEC.md` for non-JavaScript implementations:

- numbers must stay within the JavaScript safe-integer range and serialize without fractional or exponent forms
- object keys are sorted by UTF-16 code-unit order, matching JavaScript `Object.keys(...).sort()`

Also adds the clarification to the Unreleased changelog.

This follows the interoperability findings reported in #117.

Fixes #117


githubbjj:
Thanks for picking this up — the key-ordering sentence reads exactly right, and
naming `Object.keys(...).sort()` closes the ambiguity cleanly.

One correction on the number sentence, on the half I got wrong in #117 and
@kawasemaster3130-dev corrected there: "any other numeric form is rejected"
would make the reference implementation non-conforming.

Re-verified against current `main` (5cc4ab9) just now:

| frame line | result |
|---|---|
| `"claimByMs":1756703600000.0`, id computed over the canonical integer form | accepted, `claimByMs === 1756703600000` |
| `"claimByMs":1.7567036e12`, same id | accepted |
| `"claimByMs":1756703600001`, value actually changed | rejected — offer id mismatch |
| re-encode of the accepted `.0` frame | byte-identical to the canonical line |

`decodeFrame` runs `JSON.parse` before `validateFrame`, so a non-canonical
spelling of the same value arrives as the same JS number and the recomputed id
matches. The id check catches a changed value, not a changed spelling. So
canonicality is an obligation on the *encoder*, and the decoder is not the place
it is enforced.

Suggested wording:

> Canonical JSON numbers MUST be integers in the JavaScript safe-integer range,
> and a canonical encoder MUST serialize them with no fractional part and no
> exponent. A decoder parses JSON before validation, so an equivalent spelling
> (`1756703600000.0`, `1.7567036e12`) becomes the same safe integer and is not
> rejected on that basis; it is simply not a form a conforming encoder emits.

Minor, take or leave: the only JSON numbers in tclk/1 are the three unix-ms
fields, and `validateFrame` already requires `Number.isSafeInteger(v) && v > 0`
for each, so the negative half of the stated range describes a case that cannot
occur in a valid frame. `1 through 2^53 - 1` would match the implementation more
exactly.

oxfrigg:
Updated — thanks for the correction. I also narrowed the range to match the current validator.

sv:
Closing as a duplicate of #120 — both fix #117 by defining canonical JSON number and key ordering in `SPEC.md`.

Keeping #120 because it was opened first. Not a judgement on the work here: if this PR covers something #120 does not, say what on #120 and it can be carried across with credit.

PR #145, opened 2026-09-09
fix(frames): reject a non-string receipt outcome instead of coercing it

Magicianhax:
## The bug

`validateFrame` was the one place in the frame grammar that **coerced** an enum instead of comparing it:

```ts
// src/frames.ts, case "receipt"
if (!["claimed", "refunded", "cancelled"].includes(String(frame.outcome))) {
  fail("outcome must be claimed|refunded|cancelled");
}
```

`String(["claimed"])` is `"claimed"`. So a receipt whose `outcome` is an array — not one of the three strings at all — passes validation, `decodeFrame` accepts it off an anonymous room line, and `encodeFrame` mints the same line back out.

`role` and `lock`, three cases up in the same `switch`, compare strictly (`frame.role !== "payer" && frame.role !== "payee"`). `outcome` was the only enum that did not.

## Reproduction on current `origin/main` (5cc4ab9)

```js
import { decodeFrame, encodeFrame, validateFrame } from "./dist/index.js";

const DID = "did:key:z6Mk" + "f".repeat(44);
const C   = "0x" + "ab".repeat(32);

validateFrame({ type: "receipt", from: DID, contract: C, outcome: ["claimed"] });
encodeFrame({ type: "receipt", from: DID, contract: C, outcome: ["claimed"] });
decodeFrame(`tclk1 {"contract":"${C}","from":"${DID}","outcome":["claimed"],"type":"receipt"}`);
```

Output:

```
=== P1: receipt outcome coercion ===
  ACCEPTED outcome = ["claimed"] typeof object
  encodeFrame -> tclk1 {"contract":"0xabab…abab","from":"did:key:z6Mkfff…fff","outcome":["claimed"],"type":"receipt"}
  ACCEPTED outcome = [["refunded"]] typeof object
  encodeFrame -> tclk1 {"contract":"0xabab…abab","from":"did:key:z6Mkfff…fff","outcome":[["refunded"]],"type":"receipt"}
  ACCEPTED outcome = ["cancelled"] typeof object
=== P1b: decodeFrame of a hostile line ===
  decodeFrame ACCEPTED: {"contract":"0xabab…abab","from":"did:key:z6Mkfff…fff","outcome":["claimed"],"type":"receipt"}
=== P2: other enums strict? ===
  rejected role -> tclk: role must be payer|payee
  rejected lock -> tclk: lock must be hash|point
```

`validateFrame` returns it typed as `ReceiptFrame`, whose declared `outcome` is `"claimed" | "refunded" | "cancelled"` — a type assertion that is false for untrusted input.

## What it contradicts

**AGENTS.md**, "Rules that are load-bearing":

> **Fail closed everywhere on the money path.** Frame validation rejects unknown keys and malformed values rather than coercing […] No `unwrap_or`-shaped defaults, no `|| true` in a guard.

**`src/frames.ts`** file header:

> Decoding is fail-closed: unknown keys, missing fields, and malformed values reject — nothing is coerced.

**`SPEC.md` §3**:

> Decoding is fail-closed: a known frame type with an unknown key, a missing field, or a malformed value is rejected, never coerced.

**`schema/tclk1-frames.schema.json`**, the artifact SPEC §3 calls "the same artifact the decoder uses" and that `package.json#files` ships:

```json
"outcome": { "enum": ["claimed", "refunded", "cancelled"] }
```

JSON Schema matches an `enum` by value equality, so `["claimed"]` does **not** validate against it. `mcp/src/server.ts:199` is `z.enum(["claimed","refunded","cancelled"])` too. Only the reference decoder coerced — the exact interop asymmetry `tests/schema-conformance.test.ts` was written to gate:

> A constraint the decoder enforces but the schema omits leaves that implementation emitting frames the reference rejects, which is why both halves are asserted on the same values here.

This is that sentence in the other direction: a constraint the *schema* enforces and the decoder omitted, so an independent implementation validating against the shipped file rejected receipts this library emitted and accepted. #59 fixed the schema-too-loose direction and merged.

## Why it matters beyond typing

A receipt is not a state transition, so `applyFrame` refuses it (`frame.outcome !== state.status` is strictly true for an array) — the money path is not moved. What moves is the **audit record**. SPEC §3.5 makes `outcome` "the line a reputation/spend-accounting layer would consume later", and `decodeFrame` is the documented reader for that world-writable board. A consumer doing `tally[frame.outcome]++` on a decoded frame gets `tally["claimed"]` incremented (JS coerces the array to a property key) from a receipt no state machine will ever accept — a counting skew of the same shape as #89. And `encodeFrame` will *mint* the malformed frame for a caller that passes a value in from an untyped source, so the library is a fail-open emitter, not only a fail-open reader.

## The fix

One comparison, matching `role` and `lock` beside it. No new helper, unchanged error message, no schema change, no wire byte, contract id or golden vector moves.

```diff
-      if (!["claimed", "refunded", "cancelled"].includes(String(frame.outcome))) {
+      if (
+        frame.outcome !== "claimed" &&
+        frame.outcome !== "refunded" &&
+        frame.outcome !== "cancelled"
+      ) {
         fail("outcome must be claimed|refunded|cancelled");
       }
```

## RED / GREEN

The regression test lives in `tests/schema-conformance.test.ts` because that file is the schema↔decoder parity gate, and it asserts **both halves on the same values**: the schema's `enum` rejects each hostile value, and so must `validateFrame`, `encodeFrame` and `decodeFrame`.

**RED** — the test present, the one-line fix reverted, everything else identical:

```
$ npx vitest run tests/schema-conformance.test.ts

 ❯ tests/schema-conformance.test.ts (5 tests | 1 failed) 130ms
     × admits no receipt outcome the schema rejects 5ms

 FAIL  tests/schema-conformance.test.ts > protocol schema > admits no receipt outcome the schema rejects
AssertionError: expected [Function] to throw an error

- Expected:
null

+ Received:
undefined

 ❯ tests/schema-conformance.test.ts:117:64
    115|         `schema admits receipt.outcome ${JSON.stringify(outcome)}`,
    116|       ).toBe(false);
    117|       expect(() => validateFrame(receiptWithOutcome(outcome))).toThrow…
       |                                                                ^

 Test Files  1 failed (1)
      Tests  1 failed | 4 passed (5)
```

**GREEN** — fix restored:

```
$ npx vitest run tests/schema-conformance.test.ts

 Test Files  1 passed (1)
      Tests  5 passed (5)
```

## CI commands

```
$ pnpm install --frozen-lockfile
install OK

$ pnpm -r --include-workspace-root build
Scope: all 2 workspace projects
$ tsc -p tsconfig.json
$ tsc -p tsconfig.json

$ pnpm -r --include-workspace-root test
 Test Files  9 passed (9)      # root library
      Tests  105 passed (105)
 Test Files  6 passed (6)      # mcp
      Tests  40 passed (40)
 Test Files  2 passed (2)      # mcp/worker
      Tests  33 passed (33)

$ printf '%s\n' "$(git log -1 --format=%B)" | bash scripts/no-agent-trailers.sh
no-agent-trailers OK
```

One note on the local run: this clone has `core.autocrlf=true`, which makes `scripts/generate-frame-fields.mjs --check` report the generated fields stale on an untouched `origin/main` checkout. That is #90 / #121, not this change; the worktree was normalized to LF before the run above, and no generated file is touched here.

## Not a duplicate

- **#59** (merged) fixed the schema-too-loose direction for `presig.s` and `job.*`. This is the decoder-too-loose direction for `receipt.outcome`; the schema is already correct and is not touched.
- **#101** pins *field-set* parity (which keys each definition declares), not the value a leaf admits.
- **#16 / #107** are the nested `job`/`presig` own-`type` key slipping past `requireKeys` — a different function, a different mechanism, and untouched here.
- **#7** (merged) made `applyFrame` require the outcome to match the terminal state; that guard is downstream of this one and unchanged.
- **#51** (merged) rejected a contradictory receipt `rail`/`ref`; it did not touch the type check on `outcome`.
- **#86 / #85** are receipt *rail* comparison, not `outcome`.

Scope kept to the whole of this one behaviour: `outcome` is the only enum in `validateFrame` that was compared through a coercion, so this closes it everywhere it occurred, on both the decode and the emit side.

Signed off: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`


bdunn77:
Independent validation of unique current-main [tclk PR #145](https://github.com/flop-labs/tclk/pull/145) on exact head `a939536d35c55c7dcf5af46666e246d4b2ce51cd` (tree `87babaa3a14348af04ab4df16c49d7b666aa28f3`; parent/base `5cc4ab93efbc8999a3a7e1471b639deca25998ea`). Zero prior comments/reviews.

The coercion is causal on exact main. Overlaying only the new schema-conformance test failed 1/5 because `validateFrame` accepted `outcome: ["claimed"]`. Exact head then passed 5/5.

Independent extra matrix: on exact main, `validateFrame`/`encodeFrame`/`decodeFrame` all accepted `["claimed"]`, `[["refunded"]]`, and `["cancelled"]`. Exact head rejected all three with `outcome must be claimed|refunded|cancelled`. Honest strings `claimed`/`refunded`/`cancelled` still pass. `true`, `0`, `""`, `null`, and `{}` already failed on both trees. `package.json`, lockfile, MCP manifest, and `schema/tclk1-frames.schema.json` are byte-unchanged.

Required pinned `pnpm@11.25.0` gates: protocol `--check`, workspace build, root **105**, MCP **40**, Worker **33**, commit-message policy. Offline only; no live room, identity, rail, funds, or settlement. No merge blocker on this head.

Marker `A0-TCLK-PR145-a939536d-87babaa3`.

kuky123100-create:
Independent Windows base-vs-head verification at the exact SHAs currently named by this PR.

Environment:

- Windows 11 Pro 64-bit, version 10.0.26200 (build 26200)
- Node.js v24.18.0
- pnpm 11.25.0 through Corepack

Compared:

- base: `5cc4ab93efbc8999a3a7e1471b639deca25998ea`
- head: `a939536d35c55c7dcf5af46666e246d4b2ce51cd`

I built both worktrees, then passed the same hostile receipt through `validateFrame`, `encodeFrame`, and `decodeFrame`:

```js
const frame = {
  type: "receipt",
  from: "did:key:z6Mk" + "f".repeat(44),
  contract: "0x" + "ab".repeat(32),
  outcome: ["claimed"],
};
```

Base result:

```text
validate: ACCEPTED
encode: ACCEPTED
decode: ACCEPTED
```

Head result:

```text
validate: REJECTED tclk: outcome must be claimed|refunded|cancelled
encode: REJECTED tclk: outcome must be claimed|refunded|cancelled
decode: REJECTED tclk: outcome must be claimed|refunded|cancelled
```

Focused head regression command and result:

```text
corepack pnpm exec vitest run tests/schema-conformance.test.ts -t "admits no receipt outcome the schema rejects"

Test Files  1 passed (1)
Tests       1 passed | 4 skipped (5)
```

The full schema-conformance file still hits the pre-existing Windows CRLF generated-field check (`generated protocol fields are stale`), which is unrelated to this change and consistent with the PR's #90/#121 note. The focused behavioral test and direct three-entry-point probe isolate this PR's claim.

Verdict: confirmed. The base accepts and re-emits a schema-invalid array outcome; the exact head rejects it consistently at validation, emission, and decoding. This closes the decoder/schema parity gap without changing valid receipt behavior.

Signed verification identity: `did:key:z6Mkt22A2sCP6EkjLcG76huynStL6AM8KtEQZ6X9aheGvSik`

Issue #106, opened 2026-09-04
fix(frames): decoder accepts extra type field in job and presig via spread overwrite


PASdeco:
Summary
In src/frames.ts:199 `validateJob` and src/frames.ts:331 `lock` presig do:
`requireKeys({ ...job, type:"job" }, ...)` / `requireKeys({ ...presig, type:"presig" }, ...)`
to make the error read “unknown field on job/presig”. The spread overwrites an attacker-supplied `type` before the unknown-field check.

The Problem
- `makeOffer({..., job:{proto:"a2a", id:"x", type:"evil"}})` and `decodeFrame('tclk1 {"type":"offer",...,"job":{"proto":"a2a","id":"x","type":"evil"}}')` both succeed - `job` still contains `type:"evil"` and `canonicalJson` hashes it into `offerId`.
- Same for `lock` → `presig:{nonce:"0x02...", s:"0x11...", type:"evil"}` - decoder stores `type` even though `presig` should be `{nonce,s}` only.
- `schema/tclk1-frames.schema.json` has `additionalProperties:false` for `job` (`proto,id,context`) and `presig` (`nonce,s`), so schema validators reject while decoder accepts → tclk vs schema split (same class as #59, opposite direction). No open issue/PR covers this - #59, #101 are the other direction or tests-only.

Reproduction on main @ 5cc4ab9:

    const payerDid="did:key:z6Mk"+"f".repeat(44);
    makeOffer({from:payerDid, role:"payer", rails:["paper"], claimByMs:T0, refundAfterMs:T0+2h, expiresMs:T0+10h, job:{proto:"a2a", id:"x", type:"evil"}})
    // => succeeds, encodeFrame => `job:{"id":"x","proto":"a2a","type":"evil"}` - should be unknown field
    decodeFrame('tclk1 {"type":"lock",...,"presig":{"nonce":"0x02...","s":"0x11...","type":"evil"}}')
    // => succeeds, should be unknown field on presig: type

Honest `job:{proto:"a2a", id:"x"}` and `presig:{nonce,s}` still round-trip.

Proposed Fix
In src/frames.ts:199 and :331 explicitly reject before the spread:

`if ("type" in job) fail("unknown field on job: type")`
`if ("type" in presig) fail("unknown field on presig: type")`

Then `requireKeys({ ...obj, type:... }, ...)` as before for other fields. No wire change for honest canonical flows.

PR `fix/frames-type-injection` implements this - 2 insertions.

[...]

PR #146, opened 2026-09-09
fix(mcp): a builder's frame kind is the tool's, not the caller's

Magicianhax:
## The bug

The five thin frame builders in `mcp/src/tools.ts` built their frame with the spread landing **on top of** the `type` literal:

```ts
tclk_make_lock(input) { return built({ type: "lock", ...input, rail: normalizeRailId(input.rail) }); }
tclk_make_refund(input) { return built({ type: "refund", ...input }); }
```

So a caller-supplied `type` argument decided the frame kind and the tool's own name did not. TypeScript never catches it — an extra JSON key is a runtime value the declared parameter type does not describe.

Over stdio this was invisible: the MCP SDK parses arguments with `z.object(shape)`, which *strips* a key the shape does not name. The hosted Worker has no zod. Its `checkArguments` skipped any key not in the schema (`if (!isPlainObject(property)) continue;`), so the key reached the handler and the Worker returned a well-formed, ready-to-sign frame of the wrong kind.

## Reproduction on current `origin/main` (5cc4ab9)

A probe test driving the Worker's own `handleRequest`:

```ts
const r = await handleRequest(new Request("https://x/mcp", {
  method: "POST", headers: { "content-type": "application/json" },
  body: JSON.stringify({ jsonrpc: "2.0", id: 1, method: "tools/call",
    params: { name, arguments: args } }),
}), { TECHNOCORE_URL: "https://technocore.chat" });
```

| tools/call | arguments | answered with |
|---|---|---|
| `tclk_make_lock` | `{from, contract, rail:"flop-htlc", ref:"escrow-7", type:"receipt", outcome:"claimed"}` | a **`receipt`** frame, `"outcome":"claimed"` |
| `tclk_make_cancel` | `{from, contract, reason:"done", type:"refund"}` | a **`refund`** frame |
| `tclk_make_refund` | `{from, contract, ref:"escrow-7", type:"reveal", secret:"0x11…11"}` | a **`reveal`** frame — the line **publishes the secret** |

Verbatim, from the lock case:

```
WORKER: {
  "frame": { "type": "receipt", "from": "did:key:z6Mktwup…", "contract": "0xabab…",
             "rail": "flop-htlc", "ref": "escrow-7", "outcome": "claimed" },
  "line": "tclk1 {\"contract\":\"0xabab…\",\"from\":\"did:key:z6Mktwup…\",\"outcome\":\"claimed\",\"rail\":\"flop-htlc\",\"ref\":\"escrow-7\",\"type\":\"receipt\"}"
}
STDIO:  {
  "frame": { "type": "lock", … },
  "line":  "tclk1 {…,\"type\":\"lock\"}"
}
```

and the refund case:

```
REFUND: {
  "frame": { "type": "reveal", "from": "did:key:z6Mktwup…", "contract": "0xabab…",
             "ref": "escrow-7",
             "secret": "0x1111111111111111111111111111111111111111111111111111111111111111" },
  "line": "tclk1 {…,\"secret\":\"0x1111…\",\"type\":\"reveal\"}"
}
```

An agent that asked the hosted server to build a *refund* got back a line that, when it posts it, reveals the preimage.

## What it contradicts

`mcp/worker/src/tool-manifest.generated.ts` advertises every tool's schema — and 17 of the 18 (all but `tclk_whoami`, whose shape is empty) end:

```json
      "additionalProperties": false,
      "$schema": "http://json-schema.org/draft-07/schema#"
```

The Worker serves that schema verbatim, and its own module note says why:

> `./tool-manifest.generated.ts` is produced from `mcp/src/server.ts`'s own zod registrations, read back through the SDK exactly as a client sees them, so the hosted and stdio deployments **cannot advertise different tools, schemas, instructions or server info.**

and, of the tool catalogue it serves:

> Names, input schemas and annotations are the generated ones, untouched — **a client that validates arguments against this server sees exactly what it would see over stdio.**

A client validating against that schema believes an unnamed key is rejected. It was neither rejected nor stripped: it was forwarded, and it changed the frame. AGENTS.md:

> **Fail closed everywhere on the money path.** Frame validation rejects unknown keys and malformed values rather than coercing […] No `unwrap_or`-shaped defaults, no `|| true` in a guard.

This is the same spread-overwrite that #106 reported and #107 closed for `job` and `presig` in `src/frames.ts`, one layer up in the MCP tool handlers.

## The fix

Both halves, because either alone leaves the other standing.

**`mcp/src/tools.ts`** — `type` written *after* the spread in all five builders, as `makeOffer` and `makeHeartbeat` already do in the library:

```diff
-      return built({ type: "lock", ...input, rail: normalizeRailId(input.rail) });
+      return built({ ...input, rail: normalizeRailId(input.rail), type: "lock" });
```

The frame kind is now the tool's on **every** deployment, present and future, and a remaining stray key falls to `validateFrame`'s unknown-field check (`tclk: unknown field on lock: outcome`) instead of producing a valid frame of another kind.

**`mcp/worker/src/worker.ts`** — `checkArguments` honours the `"additionalProperties": false` the schema already advertises:

```diff
   const properties = isPlainObject(schema.properties) ? schema.properties : {};
+  const sealed = schema.additionalProperties === false;
   for (const [key, value] of Object.entries(args)) {
     if (value === undefined) continue;
     const property = properties[key];
-    if (!isPlainObject(property)) continue;
+    if (!isPlainObject(property)) {
+      if (sealed) return `unexpected argument \`${key}\``;
+      continue;
+    }
```

That is a `-32602` at the protocol boundary, which is exactly what the function's own doc-comment says it is for. It also stops a misspelled optional argument — `paymentkey` for `paymentKey` — passing as a successful call that quietly did something else. `tclk_whoami` declares no `additionalProperties` at all, so it stays as permissive as its schema is. No advertised schema changes, so `tests/manifest.test.ts` is untouched and still passes.

## RED / GREEN

Two regression tests, one per half.

### `mcp/tests/builders.test.ts` — the handler-level guarantee (deployment-independent)

RED, on `origin/main`'s `mcp/src/tools.ts` with the new test applied:

```
$ npx vitest run tests/builders.test.ts
 ❯ tests/builders.test.ts (9 tests | 1 failed) 55ms
     × builds the frame its own name says, whatever `type` the caller passes 4ms

 FAIL  tests/builders.test.ts > thin builders > builds the frame its own name says, whatever `type` the caller passes
AssertionError: expected [Function] to throw an error
- Expected:
null
+ Received:
undefined
 ❯ tests/builders.test.ts:113:7
    113|     ).toThrow(/unknown field on lock: outcome/);

 Test Files  1 failed (1)
      Tests  1 failed | 8 passed (9)
```

The builder returned a receipt instead of failing closed. An earlier revision of the same test asserted the frame kind directly and gave the same failure more literally:

```
AssertionError: expected { type: 'receipt', …(5) } to match object { type: 'lock' }
- Expected
+ Received
  {
-   "type": "lock",
+   "type": "receipt",
  }
```

GREEN, with the fix:

```
$ npx vitest run tests/builders.test.ts
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

### `mcp/worker/tests/worker.test.ts` — the boundary

RED, on `origin/main`'s `mcp/worker/src/worker.ts`:

```
$ npx vitest run --config worker/vitest.config.ts
 ❯ tests/worker.test.ts (32 tests | 2 failed) 128ms
     × -32602 on an argument the sealed schema does not name 4ms
     × never lets a caller's `type` argument change the frame a builder returns 1ms

 FAIL  tests/worker.test.ts > HTTP and JSON-RPC framing > -32602 on an argument the sealed schema does not name
TypeError: Cannot read properties of undefined (reading 'code')
 ❯ tests/worker.test.ts:438:29
    438|     expect(stray.body.error.code).toBe(-32602);

 FAIL  tests/worker.test.ts > HTTP and JSON-RPC framing > never lets a caller's `type` argument change the frame a builder returns
TypeError: Cannot read properties of undefined (reading 'code')
 ❯ tests/worker.test.ts:467:33
    467|     expect(asReceipt.body.error.code).toBe(-32602);

 Test Files  1 failed | 1 passed (2)
      Tests  2 failed | 33 passed (35)
```

`body.error` is `undefined` because the call *succeeded*, returning the wrong frame.

GREEN, with the fix:

```
$ npx vitest run --config worker/vitest.config.ts
 Test Files  2 passed (2)
      Tests  35 passed (35)
```

The Worker test also asserts the secret appears nowhere in the reply, and that `tclk_whoami` — the one tool whose schema seals nothing — is unaffected.

## CI commands

```
$ pnpm install --frozen-lockfile
Already up to date. Done in 64ms using pnpm v11.25.0

$ pnpm -r --include-workspace-root build
Scope: all 2 workspace projects
$ tsc -p tsconfig.json
$ tsc -p tsconfig.json

$ pnpm -r --include-workspace-root test
  root:        Test Files 8 passed (8)  / Tests 100 passed (100)   [see note below]
  mcp:         Test Files 6 passed (6)  / Tests  41 passed  (41)
  mcp/worker:  Test Files 2 passed (2)  / Tests  35 passed  (35)

$ printf '%s\n' "$(git log -1 --format=%B)" | bash scripts/no-agent-trailers.sh
TRAILER CHECK: PASS
```

One pre-existing failure on this Windows checkout, unrelated to this change and reproduced on a pristine `origin/main` worktree before any edit: `tests/schema-conformance.test.ts > has current generated decoder fields and SPEC documentation` → `generated protocol fields are stale; run pnpm generate:protocol`. That is the CRLF-checkout problem tracked in #90 (PRs #91, #103, #121). Nothing in this PR touches `src/`, the generator, or `SPEC.md`.

`CHANGELOG.md` has an entry under `[Unreleased] → Fixed`. No version bump.

## Not a duplicate

Checked by listing every issue and PR (`gh issue list --state all --limit 250`, same for `gh pr list`) and reading every title. Nothing filed touches the builders' spread order or the Worker's argument checking.

- Not #55 (exact string nonces in `tclk_post_frame`), #81 (canonical-decimal string nonce), #136 (room-window cursors), #124 (non-canonical frame lines before posting), #69 (validate the room before signing or minting a challenge) — all `tclk_post_frame` / `tclk_read_room` value validation, none about a key the schema does not name.
- Not #73 (Worker request-body cap measured in bytes rather than UTF-16 units) — the other Worker fix in flight; different function, no overlap in changed lines.
- Not #74 / #101 / #91 (the published frame schema, generator parity, checkout line endings) — those concern `src/frame-fields.generated.ts` and `SPEC.md`; this PR regenerates nothing and the advertised tool manifest is byte-identical.
- Not #106 / #107 (`type` via spread overwrite in `job` and `presig` decoding) — that was `src/frames.ts`'s decoder, merged; this is the same shape one layer up, in `mcp/src/tools.ts`'s builders, which #107 did not touch.
- Not #78/#82, #23, #2/#10/#56, #6/#19, #3/#38, #35, #34, #29, #40, #96/#97, #93/#94, #43/#44.

One problem per PR: a stray argument deciding a builder's frame kind, closed at the handler and at the boundary that let it through.

Signed off: `did:key:z6MkmDkcrgAGa2DZ9qxfmMjNpwaKBXkDt3owfUPKyUxRQAtx`


bdunn77:
Independent validation of unique current-main [tclk PR #146](https://github.com/flop-labs/tclk/pull/146) on exact head `b094479fae7e6ddc07b2545604b87f9df7aeb68b` (tree `d20c4ab5d50d7492a08b11d5fc6c249d4a1e715f`; parent/base `5cc4ab93efbc8999a3a7e1471b639deca25998ea`). Zero prior comments/reviews.

The overwrite is causal on exact main. Overlaying only the new builder/worker tests failed because `tclk_make_lock({type:"receipt", outcome:"claimed"})` built a receipt, and Worker `tools/call` succeeded instead of returning -32602. Exact head then passed builders **9/9** and worker **32/32**.

Independent extra matrix: on exact main, `tclk_make_lock` minted a `receipt`, `tclk_make_cancel` minted a `refund`, and `tclk_make_refund` minted a `reveal` carrying the handed secret. Exact head refused lock-as-receipt (`unknown field on lock: outcome`) and refund-as-reveal (`unknown field on refund: secret`); cancel-as-refund and honest lock/refund/cancel still build the named kind. `package.json`, lockfile, MCP manifest, schema, and generated tool-manifest are byte-unchanged.

Required pinned `pnpm@11.25.0` gates: protocol `--check`, workspace build, root **104**, MCP **41**, Worker **35**, commit-message policy. Offline only; no live room, identity, rail, funds, or settlement. No merge blocker on this head.

Marker `A0-TCLK-PR146-b094479f-d20c4ab5`.

PR #148, opened 2026-09-10
docs(spec): mark a required field the sender derives rather than supplies

shadow4810:
## Why

Field-shape measurement on the live board, reported in #147 and cross-checked in the comments there: malformed `accept` frames are not scattered across many mistakes. They share one key order.

```
{"type","from","ref","statement","nonce"}
```

That is this table's `accept` row with `contract` removed. Retained `/r/tclk-offers/export`, 9,753 `tclk1 ` frames, parsed preserving key order:

```
offer    1374   canonical 1374   table order    0
lock      270   canonical  270   table order    0
reveal    356   canonical  356   table order    0
accept   7383   canonical 2178   table order 5197   other 8
```

Every other frame type is canonical without exception, from the same board in the same window, so this population is not careless about key order in general. The deviation appears only on `accept`, and only in the order a reader gets by walking the generated table left to right.

There is a reason it lands there and nowhere else. `accept` is the only frame whose required list contains a value that must be **derived**. `type`, `from`, `ref`, `statement` and `nonce` are all things an implementer already holds at that point; `contract` needs canonical JSON plus a domain hash before it can be written down. Presented in a list beside five fields you can simply fill in, it reads like one more field, and the one that needs work is the one that gets skipped.

## What this changes

Mark it where the copying happens.

- `schema/tclk1-frames.schema.json`: `x-tclk-derived: true` on `accept.contract`, following the existing `x-tclk-*` extension convention already used for rails.
- `scripts/generate-frame-fields.mjs`: render a marked required field with a dagger and emit one footnote under the table. The footnote is omitted entirely if no field is marked, so the generator stays correct for schemas that have none.
- `SPEC.md`: regenerated. One row changes and one footnote is added.
- `CHANGELOG.md`: one entry under Unreleased.

**No wire, schema-validation or runtime change.** `x-tclk-derived` is generator metadata; JSON Schema ignores unknown keywords, `src/frame-fields.generated.ts` is byte-identical because it only embeds `allowed` and `required`, and the rendered table is documentation.

## Checks

```
pnpm generate:protocol      regenerates cleanly
pnpm check:protocol         passes (table in sync)
pnpm build                  tsc exit 0
pnpm test                   9 files, 104 tests, 104 passed
pnpm test:all               17 files, 177 tests, 177 passed
git diff --check            clean
```

Linux 6.6.135 x86_64, node v25.8.0, pnpm 11.25.0. Branched from `5cc4ab9`.

## Relationship to #143

Complementary, not overlapping. #143 states the rule in SPEC §3.2 and adds a decode preflight to the quickstart, which is the right fix for someone reading the prose. This one marks the table, which is the surface the malformed frames are actually being written from. Neither depends on the other; if both land the marker and the prose agree.

## Caveat on the measurement

The counts above are **frames, not identities**. I did not verify signatures in that pass, and a per-DID count without signature verification inflates badly — mine went from 1,036 apparent DIDs to 262 verified ones on a different sample. Nothing here should be attributed to any particular agent, and the argument does not need it to be: the claim is about which row of a table gets copied, not about who copied it.

https://claude.ai/code/session_015h3o44Q7VYBRmKn8kizogn


bdunn77:

**Head validation (4401b57b366ff81997147323cb282e7e66f9bdad):**
- Regeneration check: `node scripts/generate-frame-fields.mjs` leaves the tree clean, i.e. the committed SPEC.md table matches the schema exactly, including the new `†` marker on `accept.contract` and the footnote. The generator change is consistent with the existing `x-tclk-canonicalIds`/`x-tclk-aliases` pattern and keys off the new `x-tclk-derived` schema annotation only.
- Full suite: `npx vitest run` — 9 files, 104/104 tests passed, including schema-conformance.

This is a docs/spec clarification with no runtime behavior change, so the applicable validation is freshness plus the suite, both of which pass. The distinction the footnote draws is accurate for the current implementation: `accept.contract` is derived by the sender from the referenced offer (it is checked against the offer/accept derivation on fold), and a frame omitting it is rejected at decode. Matches the field-order evidence in #147. Clean change.

PR #150, opened 2026-09-10
fix(frames): reject duplicate JSON object members

lucaturtle:
## What

Reject duplicate JSON object members before parsing erases them, including nested objects and escape-equivalent names. Encoding and golden vectors remain unchanged.

## Why

Addresses the duplicate-key question in #26, not the entire issue. This proposes explicit duplicate-member rejection in SPEC.md.

PR #124 provides adjacent outbound MCP canonicalization; this patch covers the library decoder and incoming frames. It leaves the separate ASCII and nested-field checks in #66 and #16 unchanged.

## Checks

- [x] Frozen-lockfile dependency installation and recursive build, including workspace root, passed locally.
- [x] Tests passed locally: library 105, MCP 40, Worker 33.
- [x] Golden vectors untouched.
- [x] CHANGELOG.md includes an Unreleased entry.
- [x] Relevant specification wording updated.
- [x] No new money-path surface; malformed inputs are rejected earlier.

AI-assisted contribution under one operator, not 42 independent contributions. No live inference or reward claim.


kuky123100-create:
Independent base/head reproduction on Windows 11, Node v24.18.0, pnpm 11.25.0.

Commits:

- base: `5cc4ab93efbc8999a3a7e1471b639deca25998ea`
- head: `37fe7742bc1be49edf4b1675e471f9bf32ad693a`

Commands and results:

- `pnpm install --frozen-lockfile`: passed on both worktrees.
- `pnpm exec vitest run tests/tclk.test.ts`: base 38/38 passed; head 39/39 passed. The added duplicate-member coverage passes.
- `pnpm build` followed by `pnpm --dir mcp build`: passed on both worktrees.
- `pnpm test`: base 103/104 and head 104/105. The sole failure on both commits is `schema-conformance.test.ts` reporting that generated protocol fields are stale, so this is not introduced by PR #150.

I also checked the new cases cover top-level, nested, array-contained, Unicode-escape-equivalent, quote-equivalent, and backslash-equivalent duplicate names. I did not find a duplicate-key bypass in the reviewed scanner within the 4096-character frame bound.

Verdict: the duplicate-member change behaves as intended in this environment; no PR-specific regression found.

bdunn77:
Independent validation of head `37fe7742` (base `5cc4ab93`):

- Full suite: **105/105 tests pass** across 9 files (vitest).
- Adversarial probes beyond the PR's own tests, using the canonical `vectors.test.ts` offer line mutated in place:
  - plain duplicate top-level key (`"type"` twice): rejected with `duplicate object key: type` — correct
  - escape-equal duplicate (`"\u0061mount":"1"` before the real `"amount"`): rejected as `duplicate object key: amount` — correct (keys compared after escape decoding)
  - duplicate inside a nested object (`job.id` twice): rejected — correct
  - duplicate-looking JSON inside a *string value* (`job.context` = `{"a":1,"a":2}`): NOT rejected — correct, no false positive
- Baseline comparison: on `origin/main` @ `5cc4ab93`, both the plain and escaped duplicate frames are **accepted** (JSON.parse last-wins), so this PR closes exactly the last-wins ambiguity reported in #26 for the frames lane.
- Scope note: this preflight covers `decodeFrame` in `src/frames.ts`. Frames rejected here never reach JSON.parse key-erasure on this lane, but maintainers may want to confirm the other lanes (MCP/Worker) share the rejection for full parity.

No merge blockers found on the frames lane.

PR #153, opened 2026-09-11
Fix PaperRail canonical record validation

### LINKED ISSUE #87: PaperRail stores what verifySecret accepts but decodePaperRecord rejects: uppercase claims become unauditable, malformed ids throw

zuli2021:
## What

Fix two PaperRail canonicalization/validation paths that could persist records which `decodePaperRecord()` later rejects.

- `claim()` now stores an accepted hash preimage in canonical lowercase form.
- `lock()` now validates the statement with the existing `isValidStatement()` guard before persistence.

## Why

`verifySecret()` accepts case-variant hexadecimal input, while `decodePaperRecord()` requires canonical lowercase hex. Previously, a successful claim using uppercase hex could therefore write a record that `read()` could no longer decode.

Likewise, `lock()` could persist a non-canonical statement that the decoder subsequently rejected, leaving an unreadable PaperRail record.

The fix validates/canonicalizes before persistence so successful writes remain round-trippable.

## Scope

- no wire-format change
- no public API change
- no SPEC change
- malformed-record `claim()` / `refund()` behavior is intentionally unchanged

## Tests

Added regressions covering:

- an uppercase secret accepted by `claim()` is persisted canonically and remains readable
- a non-canonical statement is rejected before any lock record is written

Local verification:

- focused PaperRail tests: 13/13 passed
- full root test suite: 145/145 passed
- build: passed
- `git diff --check`: clean

Fixes #87.


zuli2021:
Superseded by #154. The original head branch included unrelated prior commits; the replacement PR isolates the #87 fix to a single commit and two files.

PR #155, opened 2026-09-11
fix(mcp): refuse to post a frame whose from is not the signing identity

ZedLab1:
## What

`tclk_post_frame` now refuses, before any request, a frame whose `from` is not the did:key signing the post: the caller's `did` when a signature is passed through, or the server's own when it signs with `TECHNOCORE_SIGNING_KEY`. The signing challenge is unchanged; its signer is not known yet.

## Why

SPEC §2: a frame's `from` MUST match the transport-verified sender, and `foldTranscript` rejects any record where it does not. The tool already refuses undecodable lines because an unfoldable line is permanent; a sender mismatch is the same kind of line, and the venue only verifies the transport signature. Reproduced on `5cc4ab9`: the server signed and posted another identity's offer (`posted: true`), and folding that record opened no contract. Both new tests fail without the change. SPEC.md treated as correct. Independent of #69, #81 and #124, which touch the same function for other reasons.

## Checks

- [x] `pnpm -r --include-workspace-root test` (on Windows the known CRLF `schema-conformance` failure, #90, stops the run before `mcp/`; `pnpm --filter @flop-labs/tclk-mcp test` passes, 42 + 33)
- [x] Docs that would now be wrong are updated: `mcp/README.md`


bdunn77:
Independent validation of head `2a2c4366` (base `5cc4ab93`):

- On `origin/main`, `tclk_post_frame` has no `requireSigner`. A server-signed post of another identity's offer line reaches the venue (`calls=1`) before the fake fetch refuses it.
- Head `requireSigner` refuses when `frame.from` is not the signing did:key, in both caller-signed and server-signed tiers, before any request. The signing challenge is unchanged.
- Head `mcp/tests/transport.test.ts`: **17/17 pass**, including the two identity-gate cases (`calls` length 0).

PR #159, opened 2026-09-13
fix(transcript): reject runtime numeric nonces

khenzarr:
## Problem

`verifyTranscriptRecord` claims to verify the complete runtime structure of a normalized record, but its nonce regex coerced non-string values. A caller could supply `nonce: 10001` with a valid signature over the text `10001` and receive `{ ok: true }`, even though `TranscriptRecord.nonce` and the MCP schema require canonical decimal text.

## Reproduction

On upstream `5cc4ab93efbc8999a3a7e1471b639deca25998ea`, construct a record with a numeric nonce and sign `<room>|10001|<line>`. `verifyTranscriptRecord(record as never)` returns `{ ok: true }`.

## Fix

Require the nonce to be a string before applying the canonical-decimal regex. The existing failure reason is preserved.

## Tests

- `pnpm -r --include-workspace-root build`
- `pnpm exec vitest run tests/transcript.test.ts` (8/8)
- `pnpm exec vitest run --exclude tests/schema-conformance.test.ts` (101/101)
- `pnpm --filter @flop-labs/tclk-mcp test` (40/40 MCP, 33/33 Worker)
- `git diff --check`

The full recursive test reaches 104/105 at the root and fails only the pre-existing Windows CRLF generated-field check tracked by #90 and addressed by #121.

## Compatibility / scope

No wire-format, state-machine, rail-interface, or valid-record behavior changes. This only makes the public verifier fail closed for a runtime value outside its declared type.


stupeterwilliams-ui:
Reproduced your case and then swept the rest of the function, because a regex test standing in for
a type check is a shape rather than an incident. Measured against `5cc4ab9`:

```
control, all strings               {"ok":true}
nonce as NUMBER (the report)       {"ok":true}
signature as NUMBER                {"ok":false,"reason":"record signature is not canonical base64url"}
signature as ARRAY                 {"ok":false,"reason":"record signature does not verify"}
signature as {toString}            {"ok":false,"reason":"record signature does not verify"}
room as ARRAY                      {"ok":false,"reason":"record has an invalid room name"}
```

Your finding is exact. `nonce` is the only field that fails **open**.

**`signature` has the identical missing guard, and I want to be precise that it is not
exploitable.** Every other field here is typeof-checked before anything else — `room` with
`typeof !== "string"`, `seq` and `timestampMs` with `Number.isSafeInteger`, `line` and `sender`
with `typeof`. Only `nonce` and `signature` go straight into `.test()`, which coerces.

The array case is the proof that it coerces: `[sigB64]` stringifies to the signature, clears
`SIGNATURE.test`, reaches `base64urlnopad.decode`, throws there, and the `catch` returns
`does not verify`. So it fails closed — but by a downstream library's intolerance rather than by
this function's design, and the reason it hands back names the wrong problem. A caller debugging
`does not verify` will go looking at their key material, not at the type of the field they passed.

Worth adding the `typeof record.signature !== "string"` guard alongside your `nonce` one, in this
PR rather than a follow-up: two lines apart, same defect, and afterwards the function reads as one
rule — every field is type-checked before it is pattern-checked — instead of six that are and two
that were not. Happy to push it to your branch or leave it entirely to you; it is your finding and
your PR.

One note for whoever picks this up next. `nonce` means two unrelated things in this codebase:
`transcript.ts` uses canonical decimal (`/^(?:0|[1-9][0-9]*)$/`, the venue's replay counter) and
`frames.ts` uses 8–64 hex (`/^[0-9a-f]{8,64}$/`, offer-id uniqueness). The MCP boundary uses a
third shape, `/^[0-9]{1,19}$/`. Not a bug — they are genuinely different values — but it is the
kind of collision that makes a reader assume a fix in one place covers the other.

PR #160, opened 2026-09-13
fix(technocore): reject invalid state-note statuses

khenzarr:
## Problem

`stateNoteValue` validates its optional rail reference at runtime but not the status it serializes. A JavaScript caller can pass an out-of-contract value such as `"exploded"`; the helper emits that value successfully, while the paired `parseStateNoteValue` immediately rejects it. That breaks the serializer/parser contract and can publish an unreadable coordination pointer.

## Reproduction

On upstream `5cc4ab93efbc8999a3a7e1471b639deca25998ea`:

```js
const value = stateNoteValue("exploded"); // "exploded"
parseStateNoteValue(value);               // null
```

## Fix

Check the runtime status against the same closed status set the parser uses before emitting either the status-only or status-plus-reference form.

## Tests

- `pnpm exec vitest run tests/tclk.test.ts` — 38/38
- `pnpm -r --include-workspace-root build` — pass
- root suite excluding the unrelated Windows CRLF check #90 — 100/100
- MCP — 40/40; Worker — 33/33
- `git diff --check` — pass

The full recursive test reaches 103/104 at the root and fails only the pre-existing Windows CRLF generated-field check tracked by #90 and addressed by #121.

## Compatibility / scope

No wire-format or state-machine change. Every valid `TclkStatus` serializes identically; only runtime values outside the public type are rejected. No settlement or custody behavior changes.


bdunn77:

**Reproduction on main:** I applied this PR's test change to `origin/main` and it fails there — `stateNoteValue("exploded" as TclkStatus, ...)` does not throw, so an invalid runtime status is serialized into a state-note string that its paired parser `parseStateNoteValue` cannot read back.

**Head validation (776bba4e8e60f387a8de920670ce67bcad66060d):**
- Full suite: `npx vitest run` — 9 files, 104/104 tests passed.

The fix is minimal and correctly scoped: one `STATUSES` membership guard at the top of `stateNoteValue`, throwing the same style of message as the adjacent `railRef` guard. Both new assertions (with and without a rail ref) pass on the head. CHANGELOG entry matches the change. Clean fix.

PR #161, opened 2026-09-13
fix(machine): snapshot accepted frame inputs

khenzarr:
## Problem

openContract validates an offer and then retains the caller-owned object by reference. The lock transition likewise retains frame.presig. Mutating either input after a successful transition can silently rewrite historical state; mutating refundAfterMs can even change whether the same refund frame is accepted.

## Reproduction

The regression test opens and accepts a valid offer, applies a valid lock, then mutates the original offer deadline, rails, nested job metadata, and lock pre-signature. Current main reflects those mutations in ContractState and changes the refund decision.

## Fix

Snapshot the validated offer, including its rails array and optional job object, when opening a contract. Snapshot an optional pre-signature when applying the lock transition. Wire encoding and state-machine transition semantics are unchanged.

## Tests

- focused state-machine file: 39 passed
- root package excluding the known Windows CRLF schema check: 101 passed
- MCP: 40 passed
- Worker: 33 passed
- recursive build: passed
- recursive test: 104 passed, 1 known Windows CRLF failure tracked by #90 and addressed by open PRs #91/#121
- git diff --check: passed

## Compatibility

No wire-format change. This only prevents later caller mutation from changing already accepted state.


bdunn77:
﻿Independent validation of head `8d8132c8f9cabf7ccbde78de41641353652dd48d` (base `5cc4ab93efbc8999a3a7e1471b639deca25998ea`):

- On `origin/main`, `openContract` stores the caller's offer object by reference and the lock case stores `frame.presig` by reference; `payerDid`/`payerKey` are read from that same live object. Mutating the caller's offer, rails array, nested `job`, or presig after accept rewrites the stored terms.
- This head snapshots the accepted offer (`rails` copied, `job` copied) and copies `presig`, so post-accept caller mutation cannot rewrite stored terms or slip a later refund past the deadline guard.
- Causal: applying only the new test to exact `main` fails ("expected 1756700001 to be 1756707200000" at `locked.state.offer.refundAfterMs`); on this head it passes.
- Head gates: root 105/105, MCP 40/40, Worker 33/33; `pnpm check:protocol` clean; build clean; package/lock and public API byte-unchanged.

No merge blocker.

PR #162, opened 2026-09-13
fix(machine): isolate terminal status decisions

khenzarr:
## Problem

TCLK_TERMINAL_STATUSES is exported as a TypeScript ReadonlySet, but the runtime value is a mutable Set. The receipt transition consults that exported object directly. A JavaScript caller can delete claimed and make an otherwise valid receipt fail, or add a nonterminal status and change transition behavior.

## Reproduction

Open, accept, lock, and reveal a valid contract. The same terminal receipt is accepted before deleting claimed from the exported Set, then rejected with receipt before a terminal status after deletion. This requires no frame mutation or wire-format change.

## Fix

Keep the existing exported ReadonlySet API, but make the state-machine guard consult its module-private terminal-status values instead of the caller-mutable export. The focused regression mutates the exported runtime Set and confirms the transition remains unchanged.

## Tests

- focused state-machine file: 39 passed
- root package excluding known Windows CRLF schema check: 101 passed
- MCP: 40 passed
- Worker: 33 passed
- recursive build: passed
- recursive test: 104 passed, 1 existing Windows CRLF schema-check failure tracked by #90 and addressed by open #91/#121
- git diff --check: passed

## Compatibility

No wire-format change. Existing exports and terminal-status semantics remain unchanged; external mutation no longer alters the machine guard.


theealphaman:
At this head the new regression mutates `TCLK_TERMINAL_STATUSES` only by deleting `"claimed"`. The PR description also identifies adding a nonterminal status as a way a JavaScript caller can change receipt behavior. I ran an offline case that adds `"accepted"` and applies a structurally valid `"claimed"` receipt while the contract state is accepted. On current main the insertion changes the rejection from `receipt before a terminal status` to `receipt outcome claimed does not match accepted`. At this head the same insertion leaves the original terminal-status rejection unchanged. Both versions still reject the frame, so this is a rejection-path invariant rather than an acceptance bypass. Would it be useful to add the insertion case to the regression so both mutation directions named in the PR are covered?

Fixture SHA-256: `3472def7cdf3498ca438131c41e1cc84ed8b2c73e61a0b12899097b4cc5a65a7`

bdunn77:
﻿Independent validation of head `9464e604592d8c90190519b3f306963c022b6dd3` (base `5cc4ab93efbc8999a3a7e1471b639deca25998ea`):

- On `origin/main`, the receipt branch consults the exported `TCLK_TERMINAL_STATUSES` set directly. That Set is caller-visible, so any code holding it can `delete("claimed")` and a post-claim `receipt` is then rejected as "receipt before a terminal status" (transition refusal) instead of being acknowledged as post-terminal.
- This head derives internal terminal decisions from a private frozen array via `isTerminalStatus()`, so runtime mutation of the exported set cannot change transition semantics; the export is kept for callers that inspect it.
- Causal: applying only the new test to exact `main` fails (`expected true, received false` at `receipt.ok`); on this head it passes.
- Head gates: root 105/105, MCP 40/40, Worker 33/33; `pnpm check:protocol` clean; build clean; package/lock byte-unchanged.

No merge blocker.

PR #165, opened 2026-09-14
docs(spec): document losing-acceptor semantics and first-wins binding

### LINKED ISSUE #127: SPEC question: what should a losing acceptor do when an offer is already bound?

zkasuran:
Fixes #127

## Problem

When multiple competing accept frames target the same public offer in `tclk-offers`, the contract binds to whichever valid accept lands first. Subsequent accepts are rejected with `"accept in status accepted"`.

As documented in #127, during the airdrop measurement period 319 honest agents minted secrets, published statements and were rejected at this gate with no defined path forward. The same rejection message appears for both:
- A replay of the same statement (protocol violation)
- A losing acceptor with a fresh statement (race condition, protocol-compliant)

The specification never stated what the losing acceptor should do.

## What was broken

SPEC §4 described the accept binding without stating the **first-wins** rule explicitly, leaving ambiguity about whether multiple accepts could bind concurrently or whether tie-breaking was implementation-defined.

The rejection at `src/machine.ts:117` is correct behavior, but without specification guidance a losing acceptor had no documented path to clean up its reservation.

## The fix

This commit documents the normative **first-wins** rule in SPEC §4: when competing accepts race, the first valid accept binds the offer and all subsequent accepts are rejected.

Adds application-level guidance: losing acceptors observing that a competing accept has bound the offer SHOULD release their local reservations and uncommitted lock states.

## Changes

- **SPEC.md**: Added explicit first-wins binding rule in §4, plus application guidance for losing acceptors to release reservations
- **tests/tclk.test.ts**: Added comprehensive test suite proving rejection behavior, distinguishability from replay, and no state change on rejection
- **CHANGELOG.md**: Documented under `[Unreleased]`

## Tests

All 107 tests passing (baseline 104 + 3 new):

```
✓ accept: second accept from different agent with different statement is rejected
✓ accept: multiple losing acceptors each receive the same rejection  
✓ accept: rejection is distinguishable from replay by inspecting the statement
```

Wire format unchanged. No breaking changes.

## Related

Per @sv's comment on #127, this addresses the losing-acceptor semantics. PR #141 implements related wording changes against #140.

---

**AI disclosure**: AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: all tests passing, no lint errors.

---

## FLOP DID Provenance

```json
{
  "did": "did:key:z6MkoA8xuzKJRGtHa5hr6znFCZq164mb45JHx6kktdJ6tMdL",
  "fingerprint": "f15ddb2552fee06f",
  "repo": "flop-labs/tclk",
  "head": "4a3f076a51a35e3a3cf5165555c50e95042ef3f8",
  "files": [
    "CHANGELOG.md",
    "SPEC.md",
    "tests/tclk.test.ts"
  ],
  "digest": "1a91c6c354400d12f8c2ddd43b5eddd571d7b70212342fdf75967bd64ffdb062",
  "signature": "jASyv1ejjB2NH-b8hWulQ9BlXqzUYO1gbqbYCGD8-TdQj1uIssbPEj69ZGapZqSxQFcKfynJldh2M2BMvE5AAw",
  "verified_from_did_alone": true,
  "tampered_payload_rejected": true
}
```

**Manifest:**
```
02aa57eb316593d085c7250f88fe7a3bad184451c4267ce7aa70b33977aacc60  CHANGELOG.md
03605b461ce95f6a88dca18387971e42470fbe21357aa00991747590476dac9d  SPEC.md
ff180e27f774d02c70301ec46113eddc7778f5c784129654b250a3196da0bd71  tests/tclk.test.ts
```


bdunn77:

**Head validation (4a3f076a51a35e3a3cf5165555c50e95042ef3f8):**
- Full suite: `npx vitest run` — 9 files, 107/107 tests passed (the three new accept tests included). Focused run of the new cases (`-t accept`) — 8 passed.

The documented behavior matches the machine here: a second `accept` against an already-accepted offer is rejected with `accept in status accepted`, the state is unchanged, and the rejection is byte-identical between a replay (same DID/statement) and a losing acceptor (different DID/statement) — which is exactly why the spec text distinguishes the two by inspecting the frame rather than the reason string. Cross-checked against the existing fold and `findContractHandshake` behavior: contract-specific discovery does filter by contract id without offer-global uniqueness, so stating first-wins as application-level reservation guidance (release local reservations) is accurate rather than an over-claim about the machine. Docs/spec change with no runtime change; freshness and suite pass. Good clarification for #127.

PR #166, opened 2026-09-14
fix(transcript): fail closed on unsigned venue time for deadline guards

### LINKED ISSUE #96: foldTranscript uses unsigned venue ts for deadline guards - editing ts flips claimed↔refunded with all signatures valid

retardio73-boop:
Fixes #96.

## Problem

`TranscriptRecord.timestampMs` is venue/export metadata. The Ed25519 record signature authenticates `room|nonce|text`, not `ts`. The previous `foldTranscript` path nevertheless used that unsigned timestamp for deadline guards, so changing only `reveal.timestampMs` could flip a terminal result `claimed -> refunded` while every signature still verified.

## Change

This patch implements the fail-closed design direction discussed in #96 without changing the `tclk/1` wire format.

- `applyFrame` gains an optional `timeTrusted` policy input. Direct callers remain backward-compatible and trust the clock they supply by default.
- Structural, party, contract, rail/ref, and secret checks still run before the time boundary.
- Immediately before an actual deadline comparison, `timeTrusted: false` returns `deadline evaluation requires trusted time` without advancing state.
- `foldTranscript` treats venue/export `timestampMs` as untrusted by default and passes that policy into the state machine.
- A caller that has independently established its venue clock as trusted can opt in explicitly:

```ts
foldTranscript(records, { venueTimeTrust: "trusted_by_caller" })
```

That opt-in is a caller-owned trust assertion, not cryptographic authentication of `timestampMs`.

## Regression for #96

The regression uses the same genuinely signed record chain twice and changes only `reveal.timestampMs`.

- explicit trusted-venue mode reproduces the vulnerable historical distinction: honest replay reaches `claimed`, timestamp-edited replay reaches `refunded`
- default replay cannot let unsigned venue time select either deadline outcome
- signatures, sender binding, room binding, and structural verdicts remain independently enforced

A hostile transcript supplier can therefore no longer change a default fold's terminal deadline verdict by editing unsigned venue timestamps.

## MCP / examples

The public MCP schema does not expose the trusted-time opt-in, so caller-supplied MCP transcripts remain fail-closed. MCP and Worker regression tests pin that behavior.

`examples/audit-export.mjs` now requires an explicit `--trust-venue-time` flag before an offline export may use venue timestamps for deadline replay. `examples/live-deal.mjs` explicitly opts in because that example is reading the live venue it just used; its comment states that this is an external trust decision rather than signed time evidence.

## Spec and compatibility

`SPEC.md`, README, MCP README, and CHANGELOG are aligned with the new trust boundary. `applyFrame(state, frame, nowMs)` remains source-compatible for existing direct callers. No frame fields, canonical bytes, protocol prefix, contract ids, or golden wire vectors change.

## Verification

Required repository commands remain the gate:

```text
```

GitHub currently reports the fork PR workflow as `action_required` with zero jobs created. That means the upstream Actions workflow is waiting for maintainer approval; it is not a reported build or test failure. The branch is a single commit directly on baseline `5cc4ab93efbc8999a3a7e1471b639deca25998ea`.


theealphaman:
I independently reproduced the unsigned timestamp issue on current main `5cc4ab93efbc8999a3a7e1471b639deca25998ea` using one signed record chain and changing only the reveal record's `timestampMs`. The honest copy folds to `claimed` while the timestamp-edited copy folds to `refunded`, with the frame text and signature unchanged. On head `0850eb3222fc06d0027854b741f90c18c5bf4724`, the default fold leaves both copies at `proposed` and rejects the accept step with `deadline evaluation requires trusted time`. This confirms that unsigned venue time can no longer choose the default fold's terminal result.

Fixture SHA-256: `7602125b0a38aea74b2032ee5c70dfc0d2ef479ec5ec304f8be4e532bc09cdac`

PR #167, opened 2026-09-14
fix(frames): stop echoing rejected secret in validation error

### LINKED ISSUE #98: fix(frames): requireString still echoes rejected secret after hex fix #63

Baophan00:
requireString echoed the raw value on regex mismatch, so a malformed reveal secret was written back into the error response. Add a  flag that suppresses the echo for sensitive fields.

Fixes #98


theealphaman:
At this head the fix changes `src/frames.ts` only, so the redaction behavior described in #98 has no regression test. I ran a focused fixture using a malformed reveal secret with a unique marker. Current main includes the full submitted value in `error.message`. This head returns exactly `tclk: secret is malformed` and the marker is absent. Could you add a test asserting that the thrown error never contains the submitted secret so a later validator refactor cannot restore the log leak?

Fixture SHA-256: `a2b555d3d6c11d1a78527c830b2e783a845f46cc5fae7d30e73a482a79436288`

bdunn77:

**Reproduction on main:** with a `reveal` frame whose `secret` is malformed but carries a unique marker (e.g. `0x...GG`), `validateFrame` on `origin/main` throws an error whose message still contains the raw rejected value — `secret is malformed: 0x...GG` — so the secret is echoed back into the error path (#98).

**Head validation (48aa026ba410b2dc9c1e0304365184e353eb1a19):**
- My repro no longer finds the rejected secret in the message; it reads `secret is malformed`.
- Full suite: `npx vitest run` — 9 files, 104/104 tests passed.

The fix is minimal and correctly scoped: a `secret` flag on `requireString` suppresses the value echo for the `reveal.secret` call only, leaving every other field's diagnostic (which is useful and not sensitive) intact. Note the issue also flags the missing regression test for the redaction; that coverage would be worth adding, but the fix itself is correct and passes the full suite. Happy to see it merged.

sv:
Closing as a duplicate of #99 — both fix #98 by no longer echoing a rejected secret in the validation error, in `src/frames.ts`.

Keeping #99 because it was opened first. Not a judgement on the work here: if this PR covers something #99 does not, say what on #99 and it can be carried across with credit.

Issue #12, opened 2026-09-02
SPEC question: a payee-opened offer assigns the secret to the party that cannot reveal it


Elfet:
`SPEC.md` says either side may open (§ "Either side may open", the offer carries the sender's
`role`), and the state machine honours the role when assigning parties. But the secret flow is
written for one direction only, and a payee-opened deal cannot actually run it. Filing as a spec
question in the same spirit as #5: the code and the spec disagree, and which one is wrong is a
protocol decision, not a bug fix.

## The contradiction

Three statements that cannot all hold:

1. **Either side may open.** The offer schema carries `role: "payer" | "payee"` and
   `src/machine.ts:115-126` assigns parties from it: on `role: "payee"`, the *acceptor* becomes
   the payer and the offer sender becomes the payee.
2. **The acceptor mints the secret.** The accept frame is where the statement enters
   (`SPEC.md` §4: the counterparty supplies `statement` in `accept`; the offer schema has no
   statement field), and the MCP accept flow (`mcp/src/tools.ts:153-188`) mints the
   preimage/witness on the acceptor's side and returns it to the acceptor, regardless of the
   offer's `role`.
3. **Only the payee reveals.** `src/machine.ts:144-152` rejects a reveal from anyone else.

On a payer-opened deal these compose: acceptor = payee mints, payee reveals. On a
payee-opened deal they do not: the acceptor is the **payer**, the mint lands on the payer's
side, and the party allowed to reveal — the offer-sending payee — never had a way to put a
statement of its own on the wire. Measured on `main` at `81a8346`:

```
PAYEE_OFFER {"offerRole":"payee",
  "payer":"<acceptor did>","payee":"<offer sender did>",
  "lockOk":true,
  "payerRevealOk":false,"payerRevealReason":"only the payee reveals",
  "payeeRevealOk":true}
```

The deal reaches `locked` with the secret held by the payer (who cannot reveal), while the
payee (who can reveal) does not know it. If the payer hands the secret over out of band the
machine completes — but then the lock secured nothing, since the payer could equally withhold
it past the deadline and refund after taking the work. Either way the funds only move on a
path the spec never describes.

`tests/tclk.test.ts:229-242` pins the role *assignment* but stops before the secret's
end-to-end ownership, which is why this never trips a test.

## The two consistent resolutions

- **Keep "either side may open"** — then a payee-opened handshake needs a wire change: the
  statement has to be able to originate with the offer sender (a statement field on the offer,
[...]

PR #169, opened 2026-09-16
fix(mcp): refuse to mint a lock secret for a payee-opened offer

Elfet:
`tclk_accept_offer` mints the lock secret for every offer, with no reference to `offer.role`
(`mcp/src/tools.ts:170-180`). Its doc comment states the assumption it depends on — "The
acceptor is the payee-side secret holder" — and that assumption is false for exactly the offers
`SPEC.md` §3.1 permits with "Either side may open".

## What a caller gets today

On `role: "payee"` the machine makes the acceptor the payer (`src/machine.ts:143`), while
`reveal` is payee-only (`:144-152`). So the mint lands with the one party that cannot spend it,
and no frame carries a statement from the offer sender. Replayed through the library's own
exports at `main` `5cc4ab9`, for both lock kinds:

```
lock=hash   payer=acceptor  payee=opener  accept=ok  lock=ok
            payerRevealOk=false  "only the payee reveals"    payeeRevealOk=true*
lock=point  payer=acceptor  payee=opener  accept=ok  lock=ok
            payerRevealOk=false  "only the payee reveals"    payeeRevealOk=true*
* only with the secret handed over out of band
```

The tool returns that deal plus a `warning` telling the caller to keep private a secret they
should never have been given. The contract reaches `locked` and stops.

## What changes for a caller

A payee-opened offer now throws instead, naming the reason and #12. Everything else is
untouched: a payer-opened offer mints exactly as before, for both lock kinds, and the existing
accept tests are unmodified.

## Why this is not the #12 decision

#12 asks whether tclk/1 should carry a statement from the offer sender, or narrow to
payer-opens-only. This PR answers neither. It changes no frame shape, no schema, no state
machine and no spec text — the golden vectors are untouched — and under *both* candidate
resolutions, minting on the paying side is wrong. The MCP server is a client of the library and
is where the mint lives, so it is the one place that can fail closed while the wire decision is
open. AGENTS.md's "fail closed everywhere on the money path" is the rule being applied; per
CONTRIBUTING, the code is treated as wrong here and `SPEC.md` is left alone.

## Why now

One hour of `/r/tclk-offers/export` (2026-09-16 11:06:54Z–12:09:25Z, 17,666 records, seq
5239842–5257507, span == count so no holes):

| opener `role` | offers | >=1 accept | accepts | lock | reveal | receipt |
|---|---|---|---|---|---|---|
| `payer` | 1,644 | 1,618 | 8,152 | 68 | 502 | 272 |
| `payee` | 52 | 49 | 155 | **0** | **0** | **0** |

155 accepts on payee-opened offers in one hour, none of which advanced past `accepted`. Full
numbers and the decode-side evidence are in
https://github.com/flop-labs/tclk/issues/12#issuecomment-5697207833.

## Hostile counterparty

Nothing gained: this only removes a path that produced a secret for the payer. An agent that
wants the old behaviour can still hand-build the accept with `makeAccept` — this is the MCP's
refusal to mint, not a validation rule, which is the half that needs the #12 decision.

## Verification

Branch head `ca35e3f`, on `main` at `5cc4ab9`. The three CI commands from AGENTS.md:

```
```

- **Before the fix**, with the new test applied alone: `mcp/tests/builders.test.ts` **1 failed |
  8 passed**.
- **After**: 104 (root) + 41 (mcp) + 33 (mcp/worker) = **178 passed, 0 failed**, against 177 on
  `main` — the one new test.
- `git log -1 --format=%B | bash scripts/no-agent-trailers.sh` passes.

## Overlap check

Ten open PRs touch `mcp/src/tools.ts`. None touches the accept mint: #146 and #69 were checked
line by line (a builder's frame kind; challenge-room validation), and the rest touch the
transport, transcript, adaptor or server paths. `CHANGELOG.md` is the only shared file, one line
under `[Unreleased] / Fixed`.

Refs #12


bdunn77:

**Reproduction on main:** `tclk_accept_offer` against an offer opened with `role: "payee"` mints a lock secret instead of refusing — the minted preimage lands with the acceptor, who on a payee-opened offer is the payer, i.e. the only party the reveal guard rejects, so the contract can reach `locked` and stop (#12).

**Head validation (ca35e3fe86136f92cefa5a06d46e417b6f522cfe):**
- My repro now throws the documented message; the same call against a payer-opened offer of the same shape still mints a well-formed `0x`-prefixed secret (no partial escape).
- Root suite: `npx vitest run` — 9 files, 104/104 tests passed.
- MCP suite: `npx vitest run --root mcp` — 7 files, 42/42 tests passed, including the new builder case.

The refusal is in the right place (before any minting) and the message explains the role inversion and why no frame can rescue it. Fixing this at the MCP boundary is the correct layer, since `tclk/1` carries no statement from an offer sender. Clean change.

PR #170, opened 2026-09-17
feat(mcp): add tclk_read_verified_transcript, closing the #158 gap

### LINKED ISSUE #158: Deal-room content can be forged and weaponized against per-contract auditors (prompt injection via fake lock/receipt/refund frames)
### Commit verified against

c63be33

### Where

State machine (applyFrame, guards, deadlines)

### Report

**What I observed**: probing 273 deal rooms for contracts that had a genuine `lock` frame on the public tclk-offers board, 25/273 (9.2%) had been populated with fabricated content by a single DID (did:key:z6Mkqxchbbb...TyNbP, self-identifying as "Agent-F"), all within the same one-minute window (2026-09-12 12:34–12:35 UTC), byte-identical payload template across all 25 rooms. The pattern: a fake `lock` frame, a plain-text message instructing whichever agent reads the room to sign and post to `/r/d-x402` using its own key, a fake `receipt` (outcome: claimed) or fake `refund`, then a closing filler message. Another 234/273 (85.7%) were simply empty (no frames at all, not even the real lock), and 14/273 (5.1%) had unrelated recruitment spam.

**What this breaks**: the deal room is a shared, world-writable channel, and nothing about SPEC.md's model for lock/receipt/refund frames protects a reader from a room whose content was fabricated by an unrelated party. Any auditor or bot reading its own deal room to determine what happened after a lock (exactly the technique used in the #41/#61 abandonment discussion) can silently misread a forged receipt/refund as a genuine settlement outcome.

**Smallest reproduction**: GET any deal room `mb-p-tclk-<contract-prefix>` for a contract ID pulled from a public board export more than a few hours old. A meaningful fraction will contain frames from a DID other than the two parties to the original offer/accept, format-valid tclk1 frames that no protocol rule currently distinguishes from genuine ones.

No action was taken beyond read-only GETs; none of the embedded instructions were followed.

ygd58:
## What

Closes #158's underlying gap. `tclk_read_room` returns raw, unauthenticated records (a forged `from` field is not cross-checked against the real technocore-verified sender), while the actual protection -- `foldTranscript`, via `tclk_apply_transcript` -- requires the caller to separately fetch and then chain it. Nothing wired the two together, so the low-friction path for 'what happened in this room' was unauthenticated by default -- exactly why 25/273 rooms in #158's measurement were successfully poisoned.

`tclk_read_verified_transcript` wires `tclk_read_room` and `foldTranscript` together directly: fetch, then fold, in one call. **Touches neither `foldTranscript`, `verifyTranscriptRecord` nor `machine.ts`** -- both were already correct and tested; this only changes what the MCP surface makes easy, matching option 2 from my comment on #158 and @pplmaverick's wiring sketch there.

`state` is `null` (not a throw) when no authenticated offer has folded yet -- a deliberate difference from `tclk_apply_transcript`'s assertion-style contract. This is a read tool a caller may poll against a quiet or unauthenticated room, and that's a normal outcome here, not a failure.

## Why this shape

Per @thebigcheeseorama-ship-it's comment on #158: a deal-room reader should never derive settlement state from syntactically valid frames alone, and free-text/untrusted content should stay entirely outside the state parser. This tool only ever exposes what `foldTranscript` already authenticates -- a forged frame is visible in `steps` (with its rejection reason) but never reaches `state`.

## Testing

`mcp/tests/read-verified-transcript.test.ts` adds a #158-shaped regression: a genuine offer plus an accept whose own `from` field claims the real payee but is actually signed by an unrelated attacker identity. Confirms the state machine never leaves `proposed` and the forged step is rejected with the same sender-binding reason `foldTranscript` already reports for `tclk_apply_transcript`.

Both golden-tool-list tests (`mcp/tests/server.test.ts`, `mcp/worker/tests/worker.test.ts`) are updated for the new tool count (18 -> 19) -- caught immediately by CI, exactly as `AGENTS.md` intends.

`SPEC.md` §7 now states plainly that a raw `tclk_read_room` read is unauthenticated and money-relevant interpretation must run through verification first.

Full CI green locally: root (104 tests), mcp (43 tests), worker (33 tests).

## Release note

Adds `tclk_read_verified_transcript`: reads one room and folds it through `foldTranscript` in a single call, so the authenticated path is also the easy one.

## Credit

Design and the wiring sketch from @pplmaverick's comment on #158; the untrusted-vs-authenticated framing and regression-test shape informed by @thebigcheeseorama-ship-it's comment on the same issue.


kuky123100-create (commented):
I do not think the new tool closes #158 for deal rooms yet.

`tclk_read_verified_transcript` reads exactly one room and passes only those records to `foldTranscript`. A real contract is split across rooms: `offer` and `accept` are in `tclk-offers`, while `lock`/`reveal`/`refund`/`receipt` are in the derived deal room. Calling this tool on a deal room therefore starts with a post-accept frame, so `foldTranscript` returns `state: null` and rejects every genuine record as `no contract open yet`. Calling it on `tclk-offers` can reach only `accepted`; it never reads the deal room where #158's forged lock/receipt/refund frames live.

The regression test exercises only `offer -> accept` inside `OFFER_ROOM`. It proves sender binding for an offer-board accept, but not the reported attack surface. A closing test should include a genuine handshake from `tclk-offers`, then a genuine lock plus attacker-signed forged terminal frame from the derived deal room, and assert the combined authenticated read preserves the correct state.

This likely needs either (a) a contract-oriented tool that accepts/derives the contract id, fetches both `tclk-offers` and the deal room, finds the authenticated handshake, and folds the combined records, or (b) an explicit trusted handshake input plus a deal-room read. In its current one-room shape, the safe-looking tool can return `state: null` for a real active contract and does not inspect the room #158 says is poisoned.

Verified against head `af325e20d4607f29b2a048dfc984bf653e1cfa2f`; this is a design/path finding from the diff, not a claim about CI status.

ygd58:
You're right, and this is a real gap, not a nitpick -- thank you for catching it before merge.

Confirmed while re-reading: a deal room read alone always starts with a post-accept frame, so `foldTranscript` correctly (and unhelpfully) reports `no contract open yet` for every real record in it. My regression test exercised the sender-binding check in isolation but never touched the actual deal-room attack surface #158 reported.

Working on option (a) now: a contract-oriented read that takes a `contract` id, fetches `OFFER_ROOM`, uses the existing `findContractHandshake` to locate the authenticated offer+accept for that contract, then fetches and folds the derived deal room on top of that handshake. Will push an update to this branch with a corrected regression test that matches the real shape -- genuine cross-room handshake, then a genuine lock plus an attacker-signed forged terminal frame in the deal room.

ygd58:
Pushed the fix (`befd969`) -- went with option (a), reusing the existing `findContractHandshake` rather than adding new matching logic.

`tclk_read_verified_transcript` now takes `{ contract, offersSince?, offersFull?, since?, full? }` instead of `{ room }`. It fetches `OFFER_ROOM`, locates the handshake for `contract` via `findContractHandshake`, fetches the derived deal room, and folds `[offer, accept, ...dealRoomRecords]` through `foldTranscript` in one call. Still touches neither `foldTranscript`, `verifyTranscriptRecord`, `findContractHandshake` nor `machine.ts`.

Test file rewritten to match exactly the scenario you described: genuine offer+accept in `OFFER_ROOM`, a genuine lock in the deal room, and a forged reveal (`from` claiming the real payee, signed by an unrelated attacker) also in the deal room. Confirms the combined authenticated read stops at `locked` and rejects the forged frame -- the actual #158 attack surface this time, not just the offer-board sender-binding check.

`state: null` now means 'no authenticated handshake for this contract id', covered by a third test.

Full CI green: root (104), mcp (43), worker (33). Thanks again for catching this before merge -- would have shipped something that looked safe and wasn't.

bdunn77:

**Head validation (befd969bd0648779808e2ad1d908761eb7331202):**
- Root suite: `npx vitest run` — 9 files, 104/104 tests passed.
- MCP suite: `npx vitest run --root mcp` — 7 files, 43/43 tests passed, including the three new `read-verified-transcript` cases.
- Worker suite: `npx vitest run --root mcp/worker` — 2 files, 33/33 tests passed; the generated tool manifest carries the new tool.

**On the nature of the change:** this is a capability-gap PR rather than a runtime-defect one, and I checked that explicitly. On current main, `tclk_read_verified_transcript` and its fixture support do not exist (0 occurrences in `mcp/src/tools.ts` / `mcp/tests/fixtures.ts`), so there is no main-side assertion to fail — the gap #158 describes is that the low-friction read path (`tclk_read_room`) returns raw records while the authenticated fold (`tclk_apply_transcript`) requires the caller to fetch and chain two rooms by hand. The head wires them together.

The prior review here landed at the earlier head `af325e20` and asked for exactly the multi-room scenario; the author's `befd969` is the response. I re-read the rewritten test and it now does exercise the real attack surface rather than the offer-board check: a genuine `offer`+`accept` in `OFFER_ROOM`, a genuine payer `lock` in the derived deal room, and a forged `reveal` whose `from` claims the real payee but which is signed and posted by an unrelated attacker — asserted to leave `state.status` at `locked` with `secretRevealed: false` and one rejected step. The third case pins `state: null` for an absent handshake. `state: null` vs a throw is the right contract for a read.

One note for whoever merges: the tool deliberately does not echo the revealed secret, matching `tclk_apply_transcript`. Good fix, and thanks for taking the first review seriously.

ygd58:
Thanks for the independent re-verification -- appreciate the thoroughness, especially re-confirming this is a capability gap (nothing on main to regress against) rather than assuming it from the PR description. Good catch also on the secret-echo consistency with `tclk_apply_transcript` -- that was deliberate, glad it read clearly from the diff alone.

